PR #3193, opened 2024-09-10
Implement REST password provider support

raphaelbadawi:
For now, compat login in MAS only supports local (DB) password authentication. But Synapse allows to configure a password_rest_provider in homeserver.yaml, to be able to authentify against services implementing the [Identity Service API](https://spec.matrix.org/v1.11/identity-service-api/), such as [MA1SD](https://github.com/ma1uta/ma1sd).

For this PR, I made it possible to configure a rest_auth_provider in the config.yaml, in the passwords block:
```
rest_auth_provider:
  url: "XXX" # the service endpoint implementing the Identity Service API
  version: "vX" # the implemented API version
```

If this configuration is filled, whenever a compat login occurs:
- password is checked against the provider, not against the database
- if the provider authenticates a user (so the user actually exists in the identity provider) but the user doesn't exist yet in MAS/Synapse, the user is created
- whenever a device ID is present in the request body, it is reused, as per [the matrix API spec](https://spec.matrix.org/v1.11/client-server-api/#post_matrixclientv3login)

If this configuration is not filled, we fallback to the previous behaviour.

The same flow is applied when authenticating from the UI and when authenticating from the login API.

Signed-off-by: Raphaël Badawi <badawiraphael@posteo.net>

### DIFF
--- a/crates/config/src/sections/passwords.rs
+++ b/crates/config/src/sections/passwords.rs
@@ -31,6 +31,22 @@ fn default_minimum_complexity() -> u8 {
     3
 }
 
+/// Configuration for REST password provider
+#[derive(Debug, Clone, Serialize, Deserialize, JsonSchema)]
+pub struct RestAuthProviderConfig {
+    /// The base URL where the identity service is implemented
+    pub url: String,
+    /// Implemented version of the identity service ("v1", "v2")
+    pub version: String,
+}
+
+impl RestAuthProviderConfig {
+    /// Constructor for RestAuthProviderConfig
+    pub fn new(url: String, version: String) -> Self {
+        Self { url, version }
+    }
+}
+
 /// User password hashing config
 #[derive(Debug, Clone, Serialize, Deserialize, JsonSchema)]
 pub struct PasswordsConfig {
@@ -52,6 +68,10 @@ pub struct PasswordsConfig {
     /// - 4: any more than that
     #[serde(default = "default_minimum_complexity")]
     minimum_complexity: u8,
+
+    /// The REST authentication provider URL
+    #[serde(skip_serializing_if = "Option::is_none")]
+    rest_auth_provider: Option<RestAuthProviderConfig>,
 }
 
 impl Default for PasswordsConfig {
@@ -60,6 +80,7 @@ impl Default for PasswordsConfig {
             enabled: default_enabled(),
             schemes: default_schemes(),
             minimum_complexity: default_minimum_complexity(),
+            rest_auth_provider: None,
         }
     }
 }
@@ -151,6 +172,12 @@ impl PasswordsConfig {
         }
 
         Ok(mapped_result)
+    }
+
+    /// Get the REST authentication config, if set.
+    #[must_use]
+    pub fn rest_auth_provider(&self) -> Option<&RestAuthProviderConfig> {
+        self.rest_auth_provider.as_ref()
     }
 }
 
--- a/crates/handlers/Cargo.toml
+++ b/crates/handlers/Cargo.toml
@@ -34,6 +34,7 @@ tower-http.workspace = true
 axum.workspace = true
 axum-macros = "0.4.1"
 axum-extra.workspace = true
+reqwest = { version = "0.12", features = ["json"] }
 rustls.workspace = true
 
 aide.workspace = true
@@ -103,4 +104,5 @@ zxcvbn = "3.1.0"
 insta.workspace = true
 tracing-subscriber.workspace = true
 cookie_store = "0.21.0"
+mockito = "1.5"
 sqlx.workspace = true
--- a/crates/handlers/src/compat/login.rs
+++ b/crates/handlers/src/compat/login.rs
@@ -9,22 +9,26 @@ use axum_extra::typed_header::TypedHeader;
 use chrono::Duration;
 use hyper::StatusCode;
 use mas_axum_utils::sentry::SentryEventID;
+use mas_config::RestAuthProviderConfig;
 use mas_data_model::{
-    CompatSession, CompatSsoLoginState, Device, SiteConfig, TokenType, User, UserAgent,
+    CompatSession, CompatSsoLoginState, Device, InvalidDeviceID, SiteConfig, TokenType, User,
+    UserAgent,
 };
-use mas_matrix::BoxHomeserverConnection;
+use mas_matrix::{BoxHomeserverConnection, ProvisionRequest};
 use mas_storage::{
     compat::{
-        CompatAccessTokenRepository, CompatRefreshTokenRepository, CompatSessionRepository,
-        CompatSsoLoginRepository,
+        CompatAccessTokenRepository, CompatRefreshTokenRepository, CompatSessionFilter,
+        CompatSessionRepository, CompatSsoLoginRepository,
     },
     user::{UserPasswordRepository, UserRepository},
-    BoxClock, BoxRepository, BoxRng, Clock, RepositoryAccess,
+    BoxClock, BoxRepository, BoxRng, Clock, Pagination, RepositoryAccess,
 };
 use rand::{CryptoRng, RngCore};
+use reqwest::Client;
 use serde::{Deserialize, Serialize};
 use serde_with::{serde_as, skip_serializing_none, DurationMilliSeconds};
 use thiserror::Error;
+use tracing::{error, info};
 use zeroize::Zeroizing;
 
 use super::MatrixError;
@@ -64,6 +68,27 @@ struct LoginTypes {
     flows: Vec<LoginType>,
 }
 
+#[derive(Serialize)]
+struct AuthRequest {
+    user: AuthUser,
+}
+
+#[derive(Serialize)]
+struct AuthUser {
+    id: String,
+    password: String,
+}
+
+#[derive(Deserialize)]
+struct AuthResponse {
+    auth: AuthResult,
+}
+
+#[derive(Deserialize)]
+struct AuthResult {
+    success: bool,
+}
+
 #[tracing::instrument(name = "handlers.compat.login.get", skip_all)]
 pub(crate) async fn get(State(password_manager): State<PasswordManager>) -> impl IntoResponse {
     let flows = if password_manager.is_enabled() {
@@ -97,6 +122,9 @@ pub struct RequestBody {
 
     #[serde(default)]
     refresh_token: bool,
+
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    device_id: Option<String>,
 }
 
 #[derive(Debug, Serialize, Deserialize)]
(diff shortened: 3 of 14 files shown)


crates/handlers/src/views/login.rs
```diff
@@ -207,76 +212,116 @@ pub(crate) async fn post(
...
+                .hash(&mut rng, password.to_owned().into_bytes().into())
+                .await;
+            let (version, hashed_password) = match result {
+                Ok((version, hashed_password)) => (version, hashed_password),
+                Err(_err) => return Err(FormError::Internal),
+            };
+            repo.user_password()
+                .upsert(&mut rng, clock, &user, version, hashed_password)
+                .await
+                .map_err(|_err| FormError::Internal)?;
```
sandhose, line 279: I'm a little confused on the usefulness of getting that persisted in the database still? Why would we want that, if we're talking to an external service to get verify the passwords?

crates/handlers/src/views/login.rs
```diff
@@ -207,76 +212,116 @@ pub(crate) async fn post(
...
         .add(&mut rng, clock, &user, user_agent)
         .await
         .map_err(|_| FormError::Internal)?;
 
+    let user_password = repo
+        .user_password()
+        .active(&user)
+        .await
+        .map_err(|_| FormError::InvalidCredentials)?
+        .ok_or(FormError::InvalidCredentials)?;
```
sandhose, line 323: Oh, is that why? To get an authentication session going on?
I feel like this should get recorded another way, maybe a new table recording authentications from an external password verification service?

crates/handlers/src/passwords.rs
```diff
@@ -35,6 +36,9 @@ struct InnerPasswordManager {
 
     /// A map of "old" hashers used only for verification
     other_hashers: HashMap<SchemeVersion, Hasher>,
+
+    /// The REST authentication provider URL, if any
+    rest_auth_provider: Option<RestAuthProviderConfig>,
 }
 
 impl PasswordManager {
```
sandhose, line 44: It feels like the password verification logic should be somewhere here instead of in the `compat` module

crates/handlers/Cargo.toml
```diff
@@ -34,6 +34,7 @@ tower-http.workspace = true
 axum.workspace = true
 axum-macros = "0.4.1"
 axum-extra.workspace = true
+reqwest = { version = "0.12", default-features = false, features = ["json", "rustls-tls"] }
```
sandhose, line 37: So pulling `reqwest` might be something we want to do in the whole project. Right now we have a bit of a manual HTTP client based on `hyper` with a lot of metrics/tracing capabilities, and that use the right CA certificates & co.

I'll try to do a PR to move to reqwest overall, because it feels a lot simpler, but I would like to avoid having a mix of both reqwest and the current http client

crates/config/src/sections/passwords.rs
```diff
@@ -31,6 +31,23 @@ fn default_minimum_complexity() -> u8 {
     3
 }
 
+/// Configuration for REST password provider
+#[derive(Debug, Clone, Serialize, Deserialize, JsonSchema)]
+pub struct RestAuthProviderConfig {
+    /// The base URL where the identity service is implemented
+    pub url: String,
```
sandhose, line 38: How does that work in terms of authentication, etc? It would be useful to include links here to specifications for the API of such services, if there is any

crates/handlers/src/compat/login.rs
sandhose, line 1: This feels like the wrong place to have all this logic

sandhose (commented):
Thanks a lot for this! I've quickly skimmed through it, so this is only a partial review for now.

I like the idea of pluggable password providers, but I would like to understand the use case, compared to using a lightweight IDP like [Dex](https://dexidp.io)? This is what we use for LDAP deployments for instance.

I would also like to point out that we're considering disabling the legacy `m.login.password` API, at least by default; what is your use case exactly for keeping it?

raphaelbadawi:
> Thanks a lot for this! I've quickly skimmed through it, so this is only a partial review for now.
> 
> I like the idea of pluggable password providers, but I would like to understand the use case, compared to using a lightweight IDP like [Dex](https://dexidp.io)? This is what we use for LDAP deployments for instance.
> 
> I would also like to point out that we're considering disabling the legacy `m.login.password` API, at least by default; what is your use case exactly for keeping it?

Thanks for this first review which promises some discussions and clarifications to come. I will answer to every thread by the end of next week since I'm on vacation right now, but just to clarify the main purpose of all this, [it is to have a replacement for the REST password provider for Synapse](https://github.com/ma1uta/matrix-synapse-rest-password-provider). We use a REST password provider to authenticate against a [MA1SD](https://github.com/ma1uta/ma1sd) instance (which implements the identity service API) which itself checks credentials against our own API.

Our current workflow is that a user can login to Element using our web platform credentials, with a simple Matrix login API call, then if MA1SD says the credentials are valid Synapse would actually create the user if it doesn't exist yet in its database and authenticate the user so we have a valid authenticated session. When login we pass a device_id unique to each user so we have a unique session per user coming from our web platform (otherwise each login to the platform would stack as a new session in Element and ask to be checked again which would be a mess).

I will check and answer every threads above at the end of next week, thanks a lot :-)

crates/handlers/src/views/login.rs
raphaelbadawi: Yes exactly I wanted to have an actual session going on after the call to the login API is made, so subsequent whoami calls and things like that actually work. If there is a risk to mess things up with having every compat sessions on the same table, we can store them in another table, but if it's not that risky it may be a bit overkill. In Synapse if you have REST password enabled all login requests go to the REST password provider, and it's the same with this implementation, if REST password provider is configured every compat login goes it. This is why it seemed logical to just use the same table, since when this config is enabled a compat session is unambiguously always a REST password provider compat session.

crates/handlers/src/passwords.rs
raphaelbadawi: You mean the whole call to the password provider (`authenticate_via_rest_api`)? This could be moved in the handlers create, of course, but then we should also move `user_login_with_password` there, no?

crates/handlers/Cargo.toml
raphaelbadawi: Yes it's true such change should actually be in its own PR and be global. So for now I should use the current `hyper` implementation (I must admit, I didn't manage to make it work in my first attempts, but I will try again ;-) ).

crates/config/src/sections/passwords.rs
raphaelbadawi: REST password providers are a kind of [Synapse modules](https://element-hq.github.io/synapse/latest/modules/password_auth_provider_callbacks.html). What they do is calling an identity provider on login (e. g. [MA1SD](https://github.com/ma1uta/ma1sd)), which itself implements a Matrix [Identity Service API](https://spec.matrix.org/v1.12/identity-service-api/).

It takes over all API login call (e. g. all authentications using the login API are authenticated against the identity provider), and creates accounts in Synapse database when an authentication is successful for a user which doesn't exist yet in the database.

crates/handlers/src/compat/login.rs
```diff
@@ -294,6 +472,10 @@ pub(crate) async fn post(
     };
 
     repo.save().await?;
+    info!(
+        "Session and tokens saved successfully for user: {}",
+        user.username
+    );
```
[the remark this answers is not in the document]
raphaelbadawi: When login we pass a device_id unique to each user so we have a unique session per user coming from our web platform (otherwise each login to the platform would stack as a new session in Element and ask to be checked again which would be a mess). Since it's actually in [the Matrix specification](https://spec.matrix.org/v1.12/client-server-api/#relationship-between-access-tokens-and-devices), I felt like it was actually a missing piece from the compat create.

crates/handlers/src/compat/login.rs
```diff
@@ -97,6 +122,9 @@ pub struct RequestBody {
 
     #[serde(default)]
     refresh_token: bool,
+
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    device_id: Option<String>,
```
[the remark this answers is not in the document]
raphaelbadawi: When login we pass a device_id unique to each user so we have a unique session per user coming from our web platform (otherwise each login to the platform would stack as a new session in Element and ask to be checked again which would be a mess).

Since it's actually in [the Matrix specification](https://spec.matrix.org/v1.12/client-server-api/#relationship-between-access-tokens-and-devices), I felt like it was actually a missing piece from the compat create. But yes, maybe it should sit in its own PR.

raphaelbadawi:
Bump (waiting for some answers to be sure of the direction to take for the next iteration on this ;-) )

sandhose:
Very sorry for not getting back to you earlier!

First of all, I'm in favour of adding support for giving out a `device_id` on the compat login APIs, as this is currently a divergence with the spec. It would be very helpful to isolate this change in a separate PR

Second, I'm not convinced pluggable password backends really is the way to go for now. This can introduce unreliable delays during logins due to the network call, during which the database transaction/connection is held, reducing the effective concurrency of the service.

I'm saying this especially because writing a simple OAuth provider which does user/password auth like you need is relatively straightforward. The only thing this wouldn't cover is the 'legacy' `m.login.password` auth method, which ideally I really would like to remove.
I understand your use case with `ma1sd`, but it also looks like this project is abandoned?

adminceovision:
Those are great news for reqwest and the relevance of passing the device_id.

> I understand your use case with ma1sd, but it also looks like this project is abandoned?

Yeah, this is a bit sad since **password providers are still a thing in Synapse**,  but most identity providers are not maintained anymore.

Using password providers is the only session we found to be able to have some login occurring entirely in the background with API calls. We can't do it with some OpenID logic since this would trigger some redirection to a 3rd party login page at some point. So we're a bit stuck with this solution, in the absence of other solutions for background login :/

sandhose:
> Using password providers is the only solution we found to be able to have some login occurring entirely in the background with API calls.

So I think part of the issue here is that we're focusing on classic end-user interactions/logins for now, not necessarily in integrating with other systems (like if you have a Matrix chat embedded in another app).
There are multiple routes to do this in Synapse right now:

 - through the admin API
 - using the users credentials with `m.login.password`
 - by crafting JWTs and using the `org.matrix.login.jwt` login method
 - using appservice logins

What I want to avoid is having to support all of this in a backward-compatible way, as those are relatively niche uses, but rather design APIs that make sense for integrations like that.

I assume you have some piece of backend that could interact with MAS somewhere, using your users tokens? What if that backend had an API to create a new session with MAS, probably though some kind of admin API MAS would provide?
That admin API to create sessions doesn't exist yet, but the skeleton for this is already there: https://element-hq.github.io/matrix-authentication-service/topics/admin-api.html

raphaelbadawi:
Hello, thanks for the things.

MAS admin API doesn't seem to be able to create sessions (and doesn't seem to be intended to do that), authentication appservices tend to have an UI-step at some point, our user credentials are stored in our service and we would like to avoid duplicating them in multiple services, for security/sync/many-other-things reasons.

But we do store session tokens, so maybe we could manage something similar with JWT auth, we'll definitely make some trials with this option.

And if this works for us, this would reduce the scope of this PR to the device_id thing, right ?

raphaelbadawi:
Another PR will be done [for the device_id Matrix spec](https://spec.matrix.org/v1.12/client-server-api/#relationship-between-access-tokens-and-devices), this one may be closed since there is no intention to set up backward compatibility with legacy rest auth providers.

PR #3245, opened 2024-09-25
UI for linking/unlinking the Upstream Providers

### LINKED ISSUE #3244: Users can link/unlink multiple IdPs to their account

MTRNord:
Fixes #3244

image

(Email missing for privacy)

### Todos:

- [x] Allow linking the Provider
- [ ] Allow unlinking the Provider
- [ ] Improve the UI


crates/handlers/src/graphql/model/upstream_oauth.rs
```diff
@@ -45,6 +45,33 @@ impl UpstreamOAuth2Provider {
...
+            .user()
+            .ok_or_else(|| async_graphql::Error::new("User ID not found in the request context"))?
+            .id;
+
+        let mut repo = state.repository().await?;
+        let links = repo
+            .upstream_oauth_link()
+            .find_by_user_id(&self.provider, user_id)
+            .await?;
+        repo.cancel().await?;
```
sandhose, line 71: You technically can have multiple links for the same user and the same provider, which means this should be a list. You could be using `UpstreamOAuthLinkRepository::list` with the right filter? It's a little annoying because this API requires pagination, but on the other hand… we could just return the 100 first entries, and warn if (in a very unlikely scenario) there are more than this?

MTRNord: Oh right. I missed that the list function allows filtering. I will look into that.

On the pagination yeah that was my thought as well here. Like how likely is it to have multiple links on the same provider. Arguably there should be only a single link per single provider. But not sure if its enforced to be like this.

crates/storage-pg/.sqlx/query-100c6cfa46133f3b196b00ccf8f33ba0bcb4bb1afc1985092dd8ec51b40f631a.json
MTRNord, line 1: Actually not sure why it deleted them... I just ran `cargo sqlx prepare --database-url postgresql://postgres:postgres@localhost/postgres` in the storage-pg crate folder

MTRNord:
Just to keep you updated on whats missing from my POV:

The unlink button should have a way to display the error. I dont think it makes sense to actually show the error but it should show that something went wrong. I will have a look how that works in other places :)

One thing I have to check is actually the pagination. I currently think that the email and the upstream list will clash as they both use the same hook which (I think?) uses query parameters. I am not sure how to solve this issue though. Maybe thats beyond scope of this PR? Maybe not?

sandhose:
> One thing I have to check is actually the pagination. I currently think that the email and the upstream list will clash as they both use the same hook which (I think?) uses query parameters. I am not sure how to solve this issue though. Maybe thats beyond scope of this PR? Maybe not?

It's a local useState so no need to worry about them

https://github.com/element-hq/matrix-authentication-service/blob/87f34522c8a8a2f1d187f14540ee1ac01211837d/frontend/src/pagination.ts#L51-L76

I try to keep state in the URL as much as possible, but sometimes in 'embedded' lists that rarely paginate, it's fine to have local state like that

sandhose:
> Just to keep you updated on whats missing from my POV:

And just to keep a note before I forget: it would be nice to be able to set in the config whether it's possible for user to link or unlink accounts themselves, as the administrator might not want that.

You would do that by adding it to the account config, then propagate it in the SiteConfig and in the GraphQL API

crates/storage-pg/.sqlx/query-100c6cfa46133f3b196b00ccf8f33ba0bcb4bb1afc1985092dd8ec51b40f631a.json
sandhose: I did clean that up in `main`, if you want to merge back/rebase on it

crates/handlers/src/graphql/model/upstream_oauth.rs
```diff
@@ -45,6 +49,41 @@ impl UpstreamOAuth2Provider {
...
+    /// The human-readable name of the provider.
+    pub async fn human_name(&self) -> Option<&str> {
+        self.provider.human_name.as_deref()
+    }
+
+    /// UpstreamOAuth2Links associated with this provider for the current user.
+    pub async fn upstream_oauth2_links_for_user(
+        &self,
+        ctx: &Context<'_>,
+    ) -> Result<Vec<UpstreamOAuth2Link>, async_graphql::Error> {
```
sandhose, line 62: Not convinced it is good to have it at this level. I would rather have a list on the `User` to fetch links, and have the frontend do the stitching together of both lists?

crates/handlers/src/graphql/mutations/upstream_oauth.rs
```diff
@@ -0,0 +1,152 @@
+// Copyright 2024 New Vector Ltd.
+// Copyright 2023, 2024 The Matrix.org Foundation C.I.C.
```
sandhose, line 2:
```suggestion
```

crates/handlers/src/graphql/mutations/upstream_oauth.rs
```diff
@@ -0,0 +1,152 @@
...
+            RemoveUpstreamLinkPayload::Removed(link) => Some(UpstreamOAuth2Link::new(link.clone())),
+            RemoveUpstreamLinkPayload::NotFound => None,
+        }
+    }
+
+    /// The provider to which the upstream link belonged
+    async fn provider(
+        &self,
+        ctx: &Context<'_>,
+    ) -> Result<Option<UpstreamOAuth2Provider>, async_graphql::Error> {
```
sandhose, line 67: Aren't those properties already on the link?

crates/handlers/src/graphql/mutations/upstream_oauth.rs
```diff
@@ -0,0 +1,152 @@
...
+            .upstream_oauth_provider()
+            .lookup(provider_id)
+            .await?
+            .context("Upstream OAuth 2.0 provider not found")?;
+
+        Ok(Some(UpstreamOAuth2Provider::new(provider)))
+    }
+
+    /// The user to whom the upstream link belonged
+    async fn user(&self, ctx: &Context<'_>) -> Result<Option<User>, async_graphql::Error> {
```
sandhose, line 85: Same

crates/storage-pg/src/upstream_oauth2/link.rs
```diff
@@ -355,4 +355,50 @@ impl<'c> UpstreamOAuthLinkRepository for PgUpstreamOAuthLinkRepository<'c> {
...
+    async fn remove(&mut self, upstream_oauth_link: UpstreamOAuthLink) -> Result<(), Self::Error> {
+        let mut tx = self.conn.begin().await?;
+
+        // Unset the authorization sessions first, as they have a foreign key
+        // constraint on the links.
+        sqlx::query!(
+            r#"
+                    UPDATE upstream_oauth_authorization_sessions SET upstream_oauth_link_id = NULL
+                    WHERE upstream_oauth_link_id = $1
+                "#,
```
sandhose, line 379:
```suggestion
            "#,
```

crates/storage-pg/src/upstream_oauth2/link.rs
```diff
@@ -355,4 +355,50 @@ impl<'c> UpstreamOAuthLinkRepository for PgUpstreamOAuthLinkRepository<'c> {
...
+        err,
+    )]
+    async fn remove(&mut self, upstream_oauth_link: UpstreamOAuthLink) -> Result<(), Self::Error> {
+        let mut tx = self.conn.begin().await?;
+
+        // Unset the authorization sessions first, as they have a foreign key
+        // constraint on the links.
+        sqlx::query!(
+            r#"
+                    UPDATE upstream_oauth_authorization_sessions SET upstream_oauth_link_id = NULL
```
sandhose, line 377: It looks like I modelled the link on the rust side to not be possible to be in this state:

https://github.com/element-hq/matrix-authentication-service/blob/44c31ed142cbbc82fba55bfda74357b9baa17c80/crates/data-model/src/upstream_oauth2/session.rs#L23-L28

I think we should:

 - add an 'unlinked_at' timestamp
 - add a variant to this enum like
   ```rust
   Unlinked {
     competed_at: DateTime<Utc>,
     consumed_at: Option<DateTime<Utc>>,
     unlinked_at: DateTime<Utc>,
     id_token: Option<String>
   }
   ```
 - set this timestamp when we null this. This also means passing the clock to this `remove` function

Could you split out the 'link removal' API in a separate PR? That would make it easier to iterate on it, add tests, etc.

frontend/src/components/UpstreamProvider/LinkUpstreamProvider.tsx
```diff
@@ -0,0 +1,65 @@
+// Copyright 2024 New Vector Ltd.
+// Copyright 2023, 2024 The Matrix.org Foundation C.I.C.
```
sandhose, line 2:
```suggestion
```

frontend/src/components/UpstreamProvider/LinkUpstreamProvider.tsx
```diff
@@ -0,0 +1,65 @@
...
+
+const FRAGMENT = graphql(/* GraphQL */ `
+  fragment LinkUpstreamProvider_provider on UpstreamOAuth2Provider {
+    id
+    humanName
+    upstreamOauth2LinksForUser {
+      id
+      provider {
+        id
+      }
```
sandhose, line 23: Why are we selecting this?

frontend/src/components/UpstreamProvider/UnlinkUpstreamProvider.tsx
```diff
@@ -0,0 +1,111 @@
+// Copyright 2024 New Vector Ltd.
+// Copyright 2023, 2024 The Matrix.org Foundation C.I.C.
```
sandhose, line 2:
```suggestion
```

sandhose (changes requested):
I've skimmed through it, thanks a lot for working on this.

Sorry if the first round of review is a bit rough, but I think I would like to see this change split in 4:

 - changes on the storage layer for the unlink functionality
 - add a config flag to allow linking/unlinking of new accounts. This should gate existing linking flow, probably default to false
 - all the new GraphQL API you need
 - then the actual frontend

This would make it easier to review and get all the bricks together

Also I've asked our design team to come up with actual designs for these screens, I'll share that once I have them

frontend/src/components/UpstreamProvider/LinkUpstreamProvider.tsx
```diff
@@ -0,0 +1,65 @@
...
+  const [inProgress, setInProgress] = useState(false);
+  const { t } = useTranslation();
+
+  const onConfirm = async (
+    e: React.MouseEvent<HTMLButtonElement>,
+  ): Promise<void> => {
+    e.preventDefault();
+
+    setInProgress(true);
+    const upstreamURL = `/upstream/authorize/${data.id.replace("upstream_oauth2_provider:", "")}`;
```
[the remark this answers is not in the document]
MTRNord: Oops did bit read far enough. Graphql it is. :D

frontend/src/components/UpstreamProvider/LinkUpstreamProvider.tsx
```diff
@@ -0,0 +1,65 @@
...
+        kind="primary"
+        onClick={onConfirm}
+        disabled={disabled ?? inProgress}
+        Icon={disabled ?? inProgress ? undefined : IconUserAdd}
+      >
+        {inProgress && <LoadingSpinner inline />}
+        {t("frontend.link_upstream_button.text", {
+          provider: data?.humanName ?? "Unknown",
+        })}
+      </Button>
```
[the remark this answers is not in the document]
MTRNord: It probably could. I just went for the Design here in the end. A link is probably more semantically correct though. 🤔

frontend/src/components/UpstreamProvider/LinkUpstreamProvider.tsx
MTRNord: Good question. I have no idea what I wanted to do here anymore. Lets get rid of it :D

MTRNord:
Just to avoid some confusion: I will iterate on it here, then do my git magic to split this up into the 4 chunks and then see if I can split those into different PRs for review. Makes it a little less bad in terms of actual development initially :)

frontend/src/components/UserProfile/UpstreamProviderList.tsx
```diff
@@ -0,0 +1,127 @@
+// Copyright 2024 New Vector Ltd.
+//
+// SPDX-License-Identifier: AGPL-3.0-only
+// Please see LICENSE in the repository root for full details.
+
+import { H5 } from "@vector-im/compound-web";
+import { useCallback, useTransition } from "react";
+import { useTranslation } from "react-i18next";
+import { useQuery } from "urql";
```
sandhose, line 9: fyi, I'm in the process of moving off urql to @tanstack/react-query for querying data. This shouldn't make a massive difference, you'll have basically three things to change:

 - useQuery calls have a slightly different pattern (see https://github.com/element-hq/matrix-authentication-service/pull/3504/files#diff-824d80f55dbfb304574729bfc83deb51ec11af2ed3e3467cbd132c10ffaafffd )
 - useMutation calls are also slightly different (see https://github.com/element-hq/matrix-authentication-service/pull/3504/files#diff-79ea95d633c486de54ab8a248faa3ba49b217bd739d5a22644b982743dd0544e )
 - there are now 'render' snapshot tests, so you'll need to add the additional fragments to the mocked GraphQL response here: https://github.com/element-hq/matrix-authentication-service/pull/3504/files#diff-74973935dbe4236af3173d6b7989acbb33a1318c44a5175ac9d573b825b9afa8R99

MTRNord:
So sorry but I think I just dont have the time to finish this :/ 

CLA is signed so feel free to reuse the code within if it is any use in this state. It turns just out that this exceeds my available time currently and I cant estimate when I get to this :/

PR #3256, opened 2024-09-30
Support backchannel logout metadata

avdb13:


sandhose (commented):
Thanks for this! I'm unclear though why would MAS need this? This is not used anywhere, and MAS will happily ignore unknown properties when it does OP discovery

sandhose:
I'm closing this, as this pull request has been waiting for the author for a few months now

PR #3257, opened 2024-10-01
Upgrade sea-query-binder to 0.7.0-rc.2

sandhose:
This removes the dependency override we had because of an unreleased patch


zeromage:
Will this also come to 0.11 ?
We have an issue to build 0.11.0:
error: failed to load source for dependency `sea-query`

sandhose:
Sorry that was an unintended consequence of me deleting my fork of sea-query. I've restored the fork now, can you try building again?

PR #3299, opened 2024-10-04
Upgrade sqlx: 0.8.1 -> 0.8.2 to bring in compile-time fix

reivilibre:
sqlx-postgres 0.8.1 does not compile on Rust 1.79 or earlier (against MSRV policy).
This was fixed in 0.8.2.

Upstream: https://github.com/launchbadge/sqlx/issues/3460

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5878,9 +5878,9 @@ dependencies = [
 
 [[package]]
 name = "sqlx"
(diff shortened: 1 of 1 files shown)


sandhose:
Could you do that bump in the Cargo.toml, at the workspace level? I prefer to update them there usually

Issue #2785, opened 2024-09-10
Rewrite the async job system


matrixbot:
*This issue was originally created by [**@sandhose**](https://github.com/sandhose) at <https://github.com/matrix-org/matrix-authentication-service/issues/2785>.*

We're currently using apalis for async jobs, which… is probably not mature enough.
A few issues with it:

 - the crates are a bit messy
 - they are doing breaking changes all the time. See #2784 
 - the database structure is… weird
 - the jobs don't retry correctly
 - the "workers" crash when it looses the connection and don't restart
 - sharing state in the job is not type safe

Requirements:

 - have jobs in the database
 - use triggers to NOTIFY workers for new jobs (optional?)
 - lock rows with PG's `FOR UPDATE SKIP LOCKED`
 - pass the connection which locked the task row to the handler
 - have a way to do cron-like schedules for maintenance tasks

Related: 
 - https://github.com/matrix-org/matrix-authentication-service/issues/2625
 - https://github.com/matrix-org/matrix-authentication-service/issues/1490

```[tasklist]
### Pull requests
- [ ] #3307
- [ ] https://github.com/element-hq/matrix-authentication-service/pull/3367
- [ ] https://github.com/element-hq/matrix-authentication-service/pull/3455
- [ ] https://github.com/element-hq/matrix-authentication-service/pull/3558
- [ ] https://github.com/element-hq/matrix-authentication-service/pull/3678
```

```[tasklist]
### Tasks
- [x] Perform leader election for maintenance tasks
- [x] Shutdown 'lost' workers
- [x] Insert jobs in the database
- [x] Consume jobs from the database
- [x] Retry failed jobs
- [x] Gracefully shutdown workers
- [x] Use a single connection to LISTEN and interact with jobs
- [x] Schedule jobs in the future
- [ ] Re-schedule jobs from lost workers
- [x] Cron-like schedules
- [x] Metrics
- [x] Tracing with span links
```

sandhose:
Split the last task in https://github.com/element-hq/matrix-authentication-service/issues/4491

PR #3307, opened 2024-10-07
New job queue: worker registration and leader election

sandhose:
This adds the base for the new job queue system, with a simple worker registration system, as well as a leader election system.

The worker registration is meant to be used to detect lost workers and reschedule dead tasks they locked.
The leader election system is meant to have one leader performing all the maintenance work, like rescheduling tasks.

Part of #2785


crates/storage-pg/migrations/20241004075132_queue_worker.sql
```diff
@@ -0,0 +1,37 @@
...
+
+  -- When the worker was shut down
+  shutdown_at TIMESTAMP WITH TIME ZONE
+);
+
+-- This single-row table stores the leader of the queue
+-- The leader is responsible for running maintenance tasks
+CREATE UNLOGGED TABLE queue_leader (
+  -- This makes the row unique
+  active BOOLEAN NOT NULL DEFAULT TRUE UNIQUE,
```
reivilibre, line 24: I know it sounds silly, but I'd make this a `PRIMARY KEY` — maybe this sounds dogmatic? But there a handful of tools are not happy with tables that don't have a primary key, e.g. logical replication in Postgres by default, I'd say it's worth always using it instead of UNIQUE etc.

crates/storage-pg/src/queue/worker.rs
```diff
@@ -0,0 +1,237 @@
...
+            now,
+        )
+        .traced()
+        .execute(&mut *self.conn)
+        .await?;
+
+        // If no row was updated, the worker was shutdown so we return an error
+        DatabaseError::ensure_affected_rows(&res, 1)?;
+
+        Ok(worker)
```
reivilibre, line 104: the docstring says this returns the modified worker, but I don't see us modifying it.

I would expect the worker to track its own validity timestamps, but I guess the critical thing here is just that we 'take away' the Worker if we can't renew it?

crates/storage-pg/src/queue/worker.rs
```diff
@@ -0,0 +1,237 @@
...
+            db.query.text,
+        ),
+        err,
+    )]
+    async fn shutdown_dead_workers(
+        &mut self,
+        clock: &dyn Clock,
+        threshold: Duration,
+    ) -> Result<(), Self::Error> {
+        let now = clock.now();
```
reivilibre, line 149: is it reasonable to rely on the system clock (which could drift between servers)?

I suppose we could use the Postgres database's clock alternatively. But I don't know which one is best, mostly just interested in considering it carefully

reivilibre (commented):
This seems reasonable, but it also seems quite intricate and would benefit from a careful review, including aspects like whether it's robust against clock drift, node failures, etc — that sort of thing.

I would also want to carefully review what happens whether there are any problems if the current leader loses connection but still believes it is the leader.

crates/storage-pg/src/queue/worker.rs
sandhose: That's a fair point, although I'd imagine multiple servers in the same datacenter usually have the same time source/NTP server?

The clock is abstracted through a trait though, so maybe at some point we can take into account time drift, and regularly sync the local system clock with the database or something, but I wouldn't worry too much about it for now

reivilibre: It'd be good to document this at least maybe?

For a variety of reasons:

- assuming everyone will deploy workers only in the same datacentre is optimistic
  - besides, multi-datacentre or multi-zone is the hot stuff for high availability in case your datacentre does an OVH
- NTP could be misconfigured on some servers, or the network could be misconfigured preventing NTP from working
- one of the NTP servers may deviate from the others and may provide a fluke result to one server in isolation (probably a bit far-fetched and maybe NTP clients handle these problems themselves)

But overall I don't like the feeling of relying too heavily on the system clock for correctness.

Postgres' clock is at least probably consistent, assuming no multi-master Postgres fun and games (I don't know if that's feasible or not anyway).

If we can't get by just with locks and need to rely on some sort of clock, I have to say I'd be tempted to rely only on Postgres `NOW()`. If you need to pull any times into MAS for some reason, maybe do it all as relative times? (Basically avoiding using the MAS host's wall clock at all..).

Maybe this is too paranoid, perhaps? But it feels like something that  could give someone a bad time and it's not as though misconfigured clocks is a rare unheard of problem :-)...

sandhose: I've used `NOW()` only for the `expires_at` field, as this is really the time-sensitive one now. The 'dead worker shutdown' logic has a 2min threshold, which I feel like is fine to assume with have less than a minute of clock skew, especially since we're already relying on local clocks for token expirations, etc., so other things will break with large clock skews anyway

crates/storage-pg/migrations/20241004075132_queue_worker.sql
sandhose: Logical replication doesn't work anyway with UNLOGGED TABLE, but I'm happy to make that a PK, I don't think it really would change anything anyway

reivilibre: ah true. But other tools may exist. I personally just treat it as good practice, so I'd still lightly vote in favour of PK instead of UNIQUE

crates/storage-pg/src/queue/worker.rs
reivilibre: 2 mins is still a pretty tight bound/assumption on clock skew, IMO — it wouldn't surprise me to get machines that are hours out of sync in some circumstances.
You say it breaks access tokens anyway and that's true. But I think it's worth considering what the failure mode of the failure is. For access tokens, this is just going to be a wrong decision on whether a token is expired or not.

For the lock system intended to uniquely choose a worker, the failure mode is probably a lot worse than that indeed, so good to use `NOW()` there.

For this `last_seen_at`..... if this means we start assuming workers are dead when they aren't, won't this mean we potentially 'kill off' workers because we have the wrong clock? That still sounds like a hazard to me, but I haven't yet read enough to know what the effect is.

If a non-leader can kill off the leader, that certainly sounds like a hazard

reivilibre (approved):
yep I think this looks OK. The clock issues seem a lot safer with the mitigation for `expires_at` now.

crates/storage-pg/src/queue/worker.rs
sandhose: It's the leader doing the shutting down of dead workers, so the leader should in theory never be remote shutdown like that

We can probably make that a little less tight, let's say 10 minutes? In most cases anyway, workers should really gracefully shutdown, and more than a few minutes of clock skew has implications in other places, like the `id_token` generation sets the 'issued at' to now, and most OIDC libraries will validate that with max a 5min clock skew.

PR #3343, opened 2024-10-07
Implement login_hint

tonkku107:
Implements the most basic case of `login_hint` handling for MSC4198 where the username field is prepopulated based on the hint value


crates/data-model/Cargo.toml
```diff
@@ -22,7 +22,9 @@ rand.workspace = true
 rand_chacha = "0.3.1"
 regex = "1.11.0"
 woothee = "0.13.0"
+ruma-common = "0.13.0"
 
 mas-iana.workspace = true
 mas-jose.workspace = true
+mas-matrix.workspace = true
```
sandhose, line 29: I would like to avoid this mas-data-model -> mas-matrix dependency. Can we instead pass the homeserver name as a string instead of the homeserver connection?

sandhose (changes requested):
Thanks a lot, that looks good! Sorry for the delay on reviewing.

Just one small nit, if you could remove the mas-data-model -> mas-matrix dependency, that would be perfect!

sandhose (approved):
Thanks, LGTM!

PR #3352, opened 2024-10-09
Use CancellationToken and a TaskTracker to handle graceful shutdowns

sandhose:
This revamps how we gracefully shutdown, so that it's easier to spawn multiple tasks and have them gracefully shutdown.

At its core, it uses `CancellationToken` to propagate cancellation to all tasks, and a `TaskTracker` to keep track of tasks we want to wait for completion on shutdown.


reivilibre (approved):
nice

PR #3363, opened 2024-10-15
Add fetch_userinfo to upstream SSO provider

MatMaul:
Untested for now so I am keeping it as a draft.


crates/handlers/src/upstream_oauth2/callback.rs
```diff
@@ -212,32 +214,54 @@ pub(crate) async fn get(
...
+        .await?;
+        minijinja::Value::from_serialize(&user_info_resp)
+    } else {
+        minijinja::Value::from_serialize(&id_token)
+    };
 
     let env = {
         let mut env = environment();
-        env.add_global("user", minijinja::Value::from_serialize(&id_token));
+        env.add_global("user", userinfo);
```
sandhose, line 264: Just thinking out loud at this point, but it would be nice to have both the userinfo and the idtoken claims in the environment.

I don't know what the right config is to tell whether to fetch userinfo or not for example, *but* I would set `userinfo_claims` in the template environment if we fetched it, `id_token_claims` if we have an `id_token` and `user` being somehow a merge of the two?

One stretch goal of this would be for this to work with non-OIDC OAuth 2.0 providers, like GitHub, so we'd need to remove the id_token requirement at some point as well

crates/storage-pg/migrations/20241014145741_upstream_oauth_userinfo.sql
```diff
@@ -0,0 +1,7 @@
+-- Add migration script here
+ALTER TABLE "upstream_oauth_providers"
+  ADD COLUMN "user_profile_method" TEXT NOT NULL DEFAULT 'auto',
+  ADD COLUMN "userinfo_endpoint_override" TEXT;
+
+ALTER TABLE "upstream_oauth_authorization_sessions"
+  ADD COLUMN "userinfo" TEXT;
```
sandhose, line 7:
```suggestion
  ADD COLUMN "userinfo" JSONB;
```

frontend/src/gql/graphql.ts
sandhose, line 1: I think you need to update your local dependencies in the `frontend/` folder

```sh
cd frontend/
npm ci
cd ..
sh misc/update.sh
```

crates/storage-pg/migrations/20241014145741_upstream_oauth_userinfo.sql
MatMaul: This is going to be annoying : the return of `sqlx::query_as!` then wants a `serde_json::Value`.

But for example `data-model` crate doesn't depend on `serde_json`.

It may be fine to keep it a string ? There is currently not a lot of use of `JSONB` here, and it's not coherent with `id_token`.

sandhose: I'm fine to depend on `serde_json` in `mas-data-model`. I'll be doing something similar in #3521 to store extra query parameters passed on the callback

crates/config/src/sections/upstream_oauth2.rs
```diff
@@ -401,6 +421,14 @@ pub struct Provider {
     #[serde(skip_serializing_if = "Option::is_none")]
     pub token_endpoint_auth_signing_alg: Option<JsonWebSignatureAlg>,
 
+    /// Whether to fetch the user profile from the userinfo endpoint,
+    /// or to rely on the data returned in the id_token from the token_endpoint.
+    ///
+    /// Defaults to `auto`, which uses the userinfo endpoint if `openid` is not
+    /// included in `scopes`, and the ID token otherwise.
+    #[serde(default, skip_serializing_if = "UserProfileMethod::is_default")]
+    pub user_profile_method: UserProfileMethod,
```
sandhose, line 430: I'm not convinced we should replicate the Synapse behaviour here, to avoid implicit behaviour as much as possible.

What I would like to have instead, is just a boolean whether we do fetch the userinfo or not; this way, we always get/verify the ID token if we get one, we fetch the userinfo if it is configured to, and we make both available in the template, with the nice merge you did for the `user` property

crates/data-model/src/upstream_oauth2/session.rs
```diff
@@ -19,12 +19,14 @@ pub enum UpstreamOAuthAuthorizationSessionState {
         completed_at: DateTime<Utc>,
         link_id: Ulid,
         id_token: Option<String>,
+        userinfo: Option<String>,
```
sandhose, line 22: I'm fine with adding a dependency to `serde_json` here, make that a `serde_json::Value`

crates/handlers/src/upstream_oauth2/callback.rs
```diff
@@ -209,35 +213,52 @@ pub(crate) async fn get(
...
+            response.access_token.as_str(),
+            Some(verification_data),
+            &id_token_map.ok_or(RouteError::MissingIDToken)?,
+        )
+        .await?
+    } else {
+        HashMap::new()
+    };
+
+    let env = environment(id_token, userinfo.clone());
```
sandhose, line 260: I think I used the `environment` here the wrong way. In #3521, I made the environment 'static', and rather passed the dynamic context as template context. I'll let you know once that PR is merged so you can rebase your work on it

crates/handlers/src/upstream_oauth2/callback.rs
```diff
@@ -266,9 +287,15 @@ pub(crate) async fn get(
             .await?
     };
 
+    let userinfo_str = if userinfo.is_empty() {
+        None
+    } else {
+        Some(json!(userinfo).to_string())
+    };
```
sandhose, line 294: This is better implemented with `userinfo.as_ref().map(|v| json!(v))`

crates/handlers/src/upstream_oauth2/callback.rs
```diff
@@ -209,35 +213,52 @@ pub(crate) async fn get(
...
+        UpstreamOAuthProviderUserProfileMethod::UserinfoEndpoint => true,
     };
 
+    let userinfo = if fetch_userinfo_endpoint {
+        mas_oidc_client::requests::userinfo::fetch_userinfo(
+            &client,
+            lazy_metadata.userinfo_endpoint().await?,
+            response.access_token.as_str(),
+            Some(verification_data),
+            &id_token_map.ok_or(RouteError::MissingIDToken)?,
```
sandhose, line 253: Interesting, I did not know we were checking that the userinfo returns the same sub & co as the id_token… we probably want to remove that, as we'd like to support non-OIDC providers as well…

I can do that in a later PR though

sandhose (changes requested):
Skimming through this, it looks alright, thanks a lot! I think we need to wait for #3521 to land, so that you can rebase on top of it

crates/handlers/src/upstream_oauth2/callback.rs
sandhose: #3521 has been merged, so instead what I suggest is adding the `userinfo_claims` in the template context here: https://github.com/element-hq/matrix-authentication-service/blob/7296364cd3ea9abe06ae4bb67bffeb7fc28a5978/crates/handlers/src/upstream_oauth2/template.rs#L24

and then make sure it can be used in the template here: https://github.com/element-hq/matrix-authentication-service/blob/7296364cd3ea9abe06ae4bb67bffeb7fc28a5978/crates/handlers/src/upstream_oauth2/template.rs#L54-L79

sandhose (approved):
LGTM! It just occurred to me that it's missing documentation about the new options here: https://element-hq.github.io/matrix-authentication-service/reference/configuration.html#upstream_oauth2

But I'm fine with doing this in another PR

pierreozoux:
Well done @MatMaul  :)

PR #3455, opened 2024-10-31
Consume jobs from the new job queue

sandhose:
On top of #3367 

This does the bulk of processing jobs in the queue.

With this PR there are four main things missing from the job queue:

 - schedule jobs later (which we need to do exponential backoffs when retrying jobs)
 - schedule recurring jobs
 - reschedule jobs from dead workers
 - metrics


reivilibre (approved):
generally LGTM

PR #3521, opened 2024-11-15
Support Sign in with Apple

### LINKED ISSUE #2119: Support what "sign-in with Apple" is doing for client auth
*This issue was originally created by [**@sandhose**](https://github.com/sandhose) at <https://github.com/matrix-org/matrix-authentication-service/issues/2119>.*

SIWA is doing a non-standard auth mechanism, as described here: https://developer.apple.com/documentation/accountorganizationaldatasharing/generate-and-validate-tokens

sandhose:
Fixes #2119

This can be reviewed commit-by-commit.

This does three things:

 1. support the weird auth mode SiWA uses
 2. support the `form_post` response mode
 3. record extra parameters we got from the callback and expose them in the mapping templates, as SiWA will give the user display name through that


crates/cli/src/sync.rs
```diff
@@ -205,6 +211,27 @@ pub async fn config_sync(
                 }
             };
 
+            let token_endpoint_auth_method = match provider.token_endpoint_auth_method {
```
reivilibre, line 214: maybe this would be better suited as an `.into()` implementation if that's not impossible due to the crates being different

crates/axum-utils/src/cookies.rs
```diff
@@ -101,7 +101,15 @@ impl CookieOption {
...
-        cookie.set_same_site(SameSite::Lax);
+
+        // The `form_post` callback requires that, as it means a 3rd party origin will
+        // POST to MAS. This is presumably fine, as our forms are protected with a CSRF
+        // token
+        cookie.set_same_site(if self.secure() {
+            SameSite::None
+        } else {
+            SameSite::Lax
+        });
```
reivilibre, line 112: confused about why this is gated on `secure()`.

Other than that, this feels a bit undesirable — `SameSite=Lax` is part of the reason why CSRF tokens are safe against some types of attacks.
 
As a not-very-thoroughly-thought-through suggestion: could we instead have the 3rd-party origin POST to some URL `/A`, which sanitises the POST and then performs a redirect (maybe this needs to be in JavaScript? not sure) to the intended destination `/B` to 'clean' it as a same-site request / bypass the `SameSite` restriction for only intended requests? (As I think the request to `/B` would then get the cookies, even though `/A` didn't...?? But I haven't tried this)

reivilibre (changes requested):
this all seems reasonable overall, albeit a lot of code for supporting one vendor (:/ :apple:...)!

Only thing giving me pause is losing the `SameSite=Lax` constraint on the cookies. When I was doing research into modern-day CSRF about a year ago, `SameSite=Lax` is probably the most important modern development for CSRF protection that prevents some types of attacks (called Session Fixation attacks or similar).

Losing that protection would be a shame, and I'd prefer to ask the security team before we did.
But I suspect we can find a workaround, which may or may not be palatable.

I haven't tried this myself, but it sounds like a client-side redirect (NOT server-side AFAICT) can let you bypass a SameSite restriction, so if we did that to intentionally bypass SameSite only for the types of requests we care about, I ***think*** that would be better for security?

crates/cli/src/sync.rs
sandhose: The crates where both structures are defined don't depend on each other, which is why we can't implement it

sandhose:
> this all seems reasonable overall, albeit a lot of code for supporting one vendor (:/ :apple:...)!

For the record, most of it is generic and reusable/useful for other providers! The `form_post` thing is not widely used but standard. Supporting different `response_mode` is standard as well. Being able to cache extra parameters on the way back could also be useful with other providers. The only real part that is very non-standard, is how they want you to craft a JWT for the client credentials

crates/handlers/src/upstream_oauth2/callback.rs
```diff
@@ -152,23 +155,38 @@ pub(crate) async fn handler(
...
+
     // Read the parameters from the query or the form, depending on what
     // response_mode the provider uses
     let params = match (provider.response_mode, query_params, form_params) {
-        (UpstreamOAuthProviderResponseMode::Query, Some(query_params), None) => query_params.0,
-        (UpstreamOAuthProviderResponseMode::FormPost, None, Some(form_params)) => form_params.0,
+        (UpstreamOAuthProviderResponseMode::Query, Some(Query(query_params)), None) => query_params,
+        (UpstreamOAuthProviderResponseMode::FormPost, None, Some(Form(form_params))) => {
+            // We got there from a cross-site form POST, so we need to render a form with
+            // the same values, which posts back to the same URL
```
reivilibre, line 180: add a comment mentioning this is due to `SameSite: Lax`, just in case anyone looks at this and wonders 'why?'

crates/handlers/src/upstream_oauth2/callback.rs
```diff
@@ -152,23 +155,38 @@ pub(crate) async fn handler(
...
+            if sessions_cookie.is_empty() {
```
reivilibre, line 181: might be also worth adding a URL parameter or something to indicate the redirect has already been done, so we don't cause a redirect loop if the cookie is genuinely empty for some reason

reivilibre (approved):

crates/handlers/src/upstream_oauth2/callback.rs
sandhose: Have done that

crates/handlers/src/upstream_oauth2/callback.rs
sandhose: Have done that

crates/axum-utils/src/cookies.rs
```diff
@@ -101,15 +101,7 @@ impl CookieOption {
...
-
-        // The `form_post` callback requires that, as it means a 3rd party origin will
-        // POST to MAS. This is presumably fine, as our forms are protected with a CSRF
-        // token
-        cookie.set_same_site(if self.secure() {
-            SameSite::None
-        } else {
-            SameSite::Lax
-        });
+        cookie.set_same_site(SameSite::Lax);
```
FullStackSimon, line 104: Hi. With this change it doesn't seem to allow auth when element-web (and mas login) are iframed from a different origin.

For example, while developing locally, I have a Vue app at http://localhost:3000. I have element-web running locally via docker at https://foo.bar.

When I access the chat component in my Vue app (where element-web is iframed) I can see the login screen etc jst fine. However, when I POST to the login endpoint, I get a 500 response as the cookie is not honoured with this change in place.

`curl 'https://auth.foo.bar/login?kind=continue_authorization_grant&id=01JDSK1J6KSAZER4X4PJG6KKKM' \
  -H 'Accept: text/html,application/xhtml+xml,application/xml;q=0.9,image/avif,image/webp,image/apng,*/*;q=0.8,application/signed-exchange;v=b3;q=0.7' \
  -H 'Accept-Language: en-GB,en-US;q=0.9,en;q=0.8' \
  -H 'Cache-Control: no-cache' \
  -H 'Connection: keep-alive' \
  -H 'Content-Type: application/x-www-form-urlencoded' \
  -H 'Origin: https://auth.foo.bar' \
  -H 'Pragma: no-cache' \
  -H 'Referer: https://auth.foo.bar/login?kind=continue_authorization_grant&id=01JDSK1J6KSAZER4X4PJG6KKKM' \
  -H 'Sec-Fetch-Dest: iframe' \
  -H 'Sec-Fetch-Mode: navigate' \
  -H 'Sec-Fetch-Site: same-origin' \
  -H 'Sec-Fetch-User: ?1' \
  -H 'Upgrade-Insecure-Requests: 1' \
  -H 'User-Agent: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/131.0.0.0 Safari/537.36' \
  -H 'sec-ch-ua: "Google Chrome";v="131", "Chromium";v="131", "Not_A Brand";v="24"' \
  -H 'sec-ch-ua-mobile: ?0' \
  -H 'sec-ch-ua-platform: "macOS"' \
  --data-raw 'csrf=vrgo4jxAFbaTvA-OKXjX2pMNCFt_rCtC2SCf3G04Yow&username=simon&password=xxxxxxxx'`

FullStackSimon: Would you happen to know of any workaround?

sandhose: This was a temporary change in this PR, which ended up getting reverted. Is the flow working as intended on a known-good client? You should try with:

 - https://element-oidc.element.dev/
 - https://areweoidcyet.com/client-implementation-guide/

I'd suggest opening a discussion or joining #matrix-auth:matrix.org for further help on troubleshooting

PR #3558, opened 2024-11-22
Allow scheduling jobs in the future

sandhose:
On top of #3455

This adds a way to schedule a job to run at a specific time.
It also tweaks the retry logic to use this new feature and retry with an exponential backoff.


crates/tasks/src/new_queue.rs
```diff
@@ -516,6 +524,17 @@ impl QueueWorker {
...
 
+        // Mark all the scheduled jobs as available
+        let scheduled = repo
+            .queue_job()
+            .schedule_available_jobs(&self.clock)
+            .await?;
+        match scheduled {
+            0 => {}
+            1 => tracing::warn!("One scheduled job marked as available"),
+            n => tracing::warn!("{n} scheduled jobs marked as available"),
```
reivilibre, line 535: surprised by `warn`.

PR #3574, opened 2024-11-25
Update configuration.md to include a missing parameter

Thanhphan1147:
Hello, 

The `translations_path` parameter was missing from the MAS configuration reference under `templates`, this PR simply adds it together with default values for different binary distributions.

Please correct me if I'm wrong about the default values.

Thanks!


sandhose (approved):
Thank you!

PR #3636, opened 2024-12-06
Foundation of new syn2mas tool

reivilibre:
This is the start of a new version of the syn2mas tool.

Given it's missing a lot of stuff still, I'm intending to merge this to an intermediary branch where I can then build upon this with the missing features.

## Notable inclusions

- parallel insert across multiple writer transactions
- syn2mas temporarily drops indices and constraints during migration, for much faster insertion performance. (rough numbers: 2 min → 15 s)
- safety feature: MAS tables are renamed during migration, to make sure a MAS instance doesn't accidentally run and access the database in its vulnerable state
- safety feature: syn2mas holds an advisory lock preventing other concurrent instances of syn2mas from running
- syn2mas can restart even if it died or was interrupted half-way through

## Notable missing features (for later PRs)

- migrating any data other than user or user password records
- migrating records for guest users (it's uncertain whether we want to do this or not)
- running sanity checks against the Synapse DB before you start migration
- configuring the server name
- configuring the source Synapse database location
- tests
- documentation page

## Performance

This currently runs a synthetic matrix.org-size migration under 15 seconds. (Bear in mind this is only one part of the migration, but it gives a rough idea of where we're going with this.)
The synthetic database was set up with the same number of **real** users as matrix.org, but guest and appservice users were not populated in the database so in reality this may run slower because the Synapse Postgres instance will need to read and skip rows corresponding to guests and appservice users.


crates/syn2mas/src/synapse_reader.rs
```diff
@@ -0,0 +1,281 @@
...
+}
+
+#[derive(Clone, Debug, sqlx::Decode)]
+pub struct FullUserId(pub String);
+
+impl Type<Postgres> for FullUserId {
+    fn type_info() -> <sqlx::Postgres as sqlx::Database>::TypeInfo {
+        <String as Type<Postgres>>::type_info()
+    }
+}
```
sandhose, line 30: We *generally* go for using the database type during lookup, and then convert (usually through a `TryFrom` impl), but I think in this case it makes sense, as Synapse has quite a few recurring patterns like timestamps as ints, booleans as ints, etc.

crates/syn2mas/src/synapse_reader.rs
```diff
@@ -0,0 +1,281 @@
...
+
+        Ok(localpart)
+    }
+}
+
+/// A Synapse boolean.
+/// Synapse stores booleans as 0 or 1, due to compatibility with old SQLite versions
+/// that did not have native boolean support.
+#[derive(Clone, Debug)]
+pub struct SynapseBool(pub bool);
```
sandhose, line 80: I think I would have gone for a private field + `impl From<SynapseBool> for bool`, mainly because I don't like having to access tuples through `.0`, but that's fiiiine

crates/syn2mas/src/synapse_reader.rs
```diff
@@ -0,0 +1,281 @@
...
+
+        let lock_type = if dry_run {
+            // We expect dry runs to be done alongside Synapse running, so we don't want to
+            // interfere with Synapse's database access in that case.
+            "ACCESS SHARE"
+        } else {
+            "EXCLUSIVE"
+        };
+        for table in TABLES_TO_LOCK {
+            query(&format!("LOCK TABLE {table} IN {lock_type} MODE NOWAIT;"))
```
sandhose, line 189: What exactly happens if we lock the table when Synapse is running? The `NOWAIT` means this will fail, or will Synapse just start hanging because it can't query those tables?

crates/syn2mas/src/synapse_reader.rs
```diff
@@ -0,0 +1,281 @@
...
+
+/// Number of migratable rows in various Synapse tables.
+/// Used to estimate progress.
+#[derive(Clone, Debug)]
+pub struct SynapseRowCounts {
+    pub users: i64,
+}
+
+pub struct SynapseReader<'c> {
+    conn: &'c mut PgConnection,
```
sandhose, line 160: If we used a `Transaction<'c, Postgres>` here, we'd get the `Drop` behaviour of rolling back the transaction

crates/syn2mas/src/synapse_reader.rs
```diff
@@ -0,0 +1,281 @@
...
+
+    /// Counts the rows in the Synapse database to get an estimate of how large the migration is going to be.
+    ///
+    /// # Errors
+    ///
+    /// Errors are returned under the following circumstances:
+    ///
+    /// - An underlying database error
+    pub async fn count_rows(&mut self) -> Result<SynapseRowCounts, Error> {
+        // TODO no need for query builder here
```
sandhose, line 225: I assume that if we wanted to use sqlx here, we'd need to split synapse reading logic to its own crate, and somehow setup a synapse-like schema for introspection?

crates/syn2mas/src/synapse_reader.rs
```diff
@@ -0,0 +1,281 @@
...
+            .and_where(Expr::col(ExtraSynapseUserIden::AppserviceId).is_null())
+            // TODO support migrating at least skeleton records for guests
+            .and_where(Expr::col(ExtraSynapseUserIden::IsGuest).eq(0))
+            .from(SynapseUserIden::Table)
+            .build_sqlx(PostgresQueryBuilder);
+
+        let conn = &mut *self.conn;
+
+        // The async stream macro works around an issue where the QueryAs output stream borrows the SQL.
+        // See: https://github.com/launchbadge/sqlx/issues/1594#issuecomment-1493146479
```
sandhose, line 269: That would be fine if the query was `'static`, right?

crates/syn2mas/src/mas_writer.rs
sandhose, line 1: We generally use the `[name]/mod.rs` pattern when we have child modules instead of `[name].rs`

crates/syn2mas/src/mas_writer.rs
```diff
@@ -0,0 +1,656 @@
...
+///
+/// - If any database error occurs whilst querying the database.
+/// - If some, but not all, syn2mas restoration tables are present.
+///   (This shouldn't be possible without syn2mas having been sabotaged!)
+pub async fn is_syn2mas_in_progress(conn: &mut PgConnection) -> Result<bool, Error> {
+    // Check if there is a resumption table...
+    let num_resumption_tables = query!(
+        r#"
+        SELECT 1 AS _dummy FROM pg_tables WHERE schemaname = current_schema
+        AND tablename IN ('syn2mas_restore_constraints', 'syn2mas_restore_indices')
```
sandhose, line 225: Maybe define an array of tables, pass it as query parameter so that you can use `len` after?

crates/syn2mas/src/mas_writer.rs
```diff
@@ -0,0 +1,656 @@
...
+    ///
+    /// Errors are returned in the following conditions:
+    ///
+    /// - If the database connection experiences an error.
+    #[allow(clippy::missing_panics_doc)] // not real
+    pub async fn new(
+        mut conn: LockedMasDatabase<'conn>,
+        mut writer_connections: Vec<PgConnection>,
+    ) -> Result<Self, Error> {
+        // Acquire an advisory lock on the database.
```
sandhose, line 255: Technically the advisory lock was already acquired at this point, or am I missing something?

crates/cli/src/commands/syn2mas.rs
```diff
@@ -0,0 +1,83 @@
...
+enum Subcommand {
+    Check,
+    Migrate,
+}
+
+/// The number of parallel writing transactions active against the MAS database.
+const NUM_WRITER_CONNECTIONS: usize = 8;
+
+impl Options {
+    pub async fn run(self, figment: &Figment) -> anyhow::Result<ExitCode> {
```
sandhose, line 37: Generally, I would appreciate tracing spans on most operations, which would help a lot diagnose what takes time, but that can be introduced later

sandhose (commented):
This looks very promising! I've left a bunch of feedback after a first read; I still want to look at the constraint save/restoration in more details later

crates/syn2mas/src/mas_writer/constraint_pausing.rs
```diff
@@ -0,0 +1,148 @@
...
+///
+/// The constraint must exist prior to this call.
+pub async fn drop_constraint(
+    conn: &mut PgConnection,
+    constraint: &ConstraintDescription,
+) -> Result<(), Error> {
+    let name = &constraint.name;
+    let table_name = &constraint.table_name;
+    debug!("dropping constraint {name} on table {table_name}");
+    sqlx::query(&format!("ALTER TABLE {table_name} DROP CONSTRAINT {name};"))
```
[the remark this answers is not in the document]
reivilibre: well, the table name isn't a data value, those can't be parameterised in any statement AFAIK.

Kind of makes sense if you imagine/remember that the table name is necessary before performing semantic analysis (or maybe this is better called statement preparation?), but it's semantic analysis that performs the type inference to understand what type the placeholder should be.

crates/syn2mas/src/synapse_reader.rs
reivilibre: *Probably* more effort than it's worth, it gets quite fiddly to maintain the different schemas :(

crates/syn2mas/src/synapse_reader.rs
reivilibre: afaik yes — do you think it's better to hardcode the query (even though not using the `query!` macro) and so it would be `&'static` then?

sandhose: we hardcode queries in Synapse all the time, right? 😬 
If you'd like to keep the query builder, you could always stick the built query in a static `LazyLock`? But maybe that's a little verbose… As you want, your call

crates/syn2mas/src/migration.rs
sandhose, line 1: I've just realised most files are missing the copyright headers

crates/cli/src/commands/syn2mas.rs
sandhose: we can always add more infos or better names on them later, but this is fine 👍

crates/syn2mas/src/mas_writer.rs
```diff
@@ -0,0 +1,656 @@
...
+    /// How many connections are in circulation
+    num_connections: usize,
+
+    /// A receiver handle to get a writer connection
+    /// The writer connection will be mid-transaction!
+    connection_rx: Receiver<Result<PgConnection, Error>>,
+
+    /// A sender handle to return a writer connection to the pool
+    /// The connection should still be mid-transaction!
+    connection_tx: Sender<Result<PgConnection, Error>>,
```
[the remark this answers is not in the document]
sandhose: You're right. The transaction either olds a ref to a connection or a connection from a pool. Weird that there is no good way to have an owned connection in there

crates/syn2mas/src/mas_writer/constraint_pausing.rs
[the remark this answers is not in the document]
sandhose: This looks less scary after looking at it again

crates/syn2mas/src/synapse_reader.rs
reivilibre: on balance, happy enough to hardcode the query, since there will only be a few Synapse queries here and they won't really be getting changed once syn2mas is done, so the risk is a bit lower than usual.

crates/syn2mas/src/synapse_reader.rs
sandhose: This point is still valid as you're already borrowing the connection here, but I'm fine leaving it like that

PR #3650, opened 2024-12-10
Make sure refreshing OAuth token is idempotent

### LINKED ISSUE #2795: Refresh token request should be idempotent
*This issue was originally created by [**@sandhose**](https://github.com/sandhose) at <https://github.com/matrix-org/matrix-authentication-service/issues/2795>.*

It can happen that a refresh token use works on the server side, but then the response never comes back to the client.

This usually means that the client gets logged out, as their old refresh token isn't valid anymore, and they don't have the new one.

One way clients could work around this would be to redo a authz grant with the same device ID, but this impractical in many situations.

To avoid this problem, we should make the refresh token grant idempotent. This means we should track whether access tokens are used or not, and allow re-using an access token if neither the new access token nor the new refresh token were used.

sandhose:
Fixes #2795

This can be reviewed commit by commit

It is basically:

 - keep track of the 'chain' of refresh tokens, so record the 'next' token when consuming one
 - recording when access tokens are first used (either through introspection or the userinfo endpoint)
 - make the cleanup job remove 'revoked' tokens, not the expired ones anymore. This is to make sure we keep the info whether the access token was used or not
 - then implement the "idempotence" logic on the refresh token grant


crates/handlers/src/oauth2/token.rs
```diff
@@ -534,6 +548,77 @@ async fn refresh_token_grant(
...
+                refresh_token: next_refresh_token_id,
+            })?;
+
+        if next_access_token.is_used() {
+            // XXX: This is a replay, we *may* want to invalidate the session
+            return Err(RouteError::RefreshTokenInvalid(next_refresh_token.id));
+        }
+
+        // Looks like it's a double-refresh, client lost their refresh token on
+        // the way back. Let's revoke the two new access and refresh tokens, and
```
reivilibre, line 604:
```suggestion
        // the way back. Let's revoke the unused access and refresh tokens, and
```

crates/handlers/src/oauth2/token.rs
```diff
@@ -534,6 +548,77 @@ async fn refresh_token_grant(
...
+        // issue new ones
+        info!(
+            oauth2_session.id = %session.id,
+            oauth2_client.id = %client.id,
+            %refresh_token.id,
+            "A refresh token was used twice, but the new refresh token was lost. Revoking the old ones and issuing new ones."
```
reivilibre, line 610: hard to get the phrasing just right but I thought I'd try..

```suggestion
            "Refresh token already used, but issued refresh and access tokens are unused. Assuming those were lost; revoking those and reissuing new ones."
```

or maybe there's no need to go to such detail in the logs

crates/handlers/src/oauth2/token.rs
```diff
@@ -1123,6 +1211,175 @@ mod tests {
...
+            .await
+            .unwrap();
+
+        let (AccessToken { access_token, .. }, RefreshToken { refresh_token, .. }) =
+            generate_token_pair(
+                &mut state.rng(),
+                &state.clock,
+                &mut repo,
+                &session,
+                Duration::microseconds(5 * 60 * 1000 * 1000),
```
reivilibre, line 1277: is there really no way to do `Duration::seconds(5 * 60)` lol? :D

reivilibre (approved):

reivilibre:
('idempotent' seems like a misleading word since the old tokens get revoked, rather than returned again, but not sure if there's a more accurate word that matches it better)

crates/handlers/src/oauth2/token.rs
sandhose: I think it's the only one that can't panic, and clippy will complain that we don't document that potential panic I think? 😬

PR #3664, opened 2024-12-12
Add id_token_signed_response_alg and userinfo_signed_response_alg

MatMaul:
Add options to upstream OAuth config to specify the expected signing algorithm for the endpoint JWT responses.


crates/config/src/sections/upstream_oauth2.rs
```diff
@@ -472,6 +482,17 @@ pub struct Provider {
...
+    /// Expected signature for the JWT payload returned by the token
+    /// authentication endpoint.
+    ///
+    /// If null, the response is expected to be an unsigned JSON payload.
+    /// Defaults to `RS256`.
+    #[serde(
+        default = "signed_response_alg_default",
+        skip_serializing_if = "is_signed_response_alg_default"
+    )]
+    pub id_token_signed_response_alg: Option<JsonWebSignatureAlg>,
```
sandhose, line 494: `None` is probably invalid for the `id_token`? I don't think this should be optional

crates/config/src/sections/upstream_oauth2.rs
```diff
@@ -497,6 +518,17 @@ pub struct Provider {
...
+    /// Expected signature for the JWT payload returned by the userinfo
+    /// endpoint.
+    ///
+    /// If null, the response is expected to be an unsigned JSON payload.
+    /// Defaults to `RS256`.
+    #[serde(
+        default = "signed_response_alg_default",
+        skip_serializing_if = "is_signed_response_alg_default"
+    )]
+    pub userinfo_signed_response_alg: Option<JsonWebSignatureAlg>,
```
sandhose, line 530: In contrast, by default most userinfo endpoints use the plain JSON response, so here the default should be `None`
```suggestion
    /// Expected signature for the JWT payload returned by the userinfo
    /// endpoint.
    ///
    /// If null, the response is expected to be an unsigned JSON payload.
    /// Defaults to `null`.
    #[serde(
        default,
        skip_serializing_if = "Option::is_none"
    )]
    pub userinfo_signed_response_alg: Option<JsonWebSignatureAlg>,
```

crates/data-model/src/upstream_oauth2/provider.rs
```diff
@@ -230,10 +230,12 @@ pub struct UpstreamOAuthProvider {
     pub token_endpoint_override: Option<Url>,
     pub userinfo_endpoint_override: Option<Url>,
     pub fetch_userinfo: bool,
+    pub userinfo_signed_response_alg: Option<JsonWebSignatureAlg>,
     pub client_id: String,
     pub encrypted_client_secret: Option<String>,
     pub token_endpoint_signing_alg: Option<JsonWebSignatureAlg>,
     pub token_endpoint_auth_method: TokenAuthMethod,
+    pub id_token_signed_response_alg: Option<JsonWebSignatureAlg>,
```
sandhose, line 238:
```suggestion
    pub id_token_signed_response_alg: JsonWebSignatureAlg,
```

crates/handlers/src/upstream_oauth2/callback.rs
```diff
@@ -274,72 +274,109 @@ pub(crate) async fn handler(
...
+            // Nonce must match.
+            mas_jose::claims::NONCE
+                .extract_required_with_options(&mut claims, session.nonce.as_str())
+                .map_err(mas_oidc_client::error::IdTokenError::from)?;
+
+            context = context.with_id_token_claims(claims);
+        } else {
+            let claims = serde_json::from_str(id_token)
+                .map_err(mas_oidc_client::error::IdTokenError::from)?;
+            context = context.with_id_token_claims(claims);
```
sandhose, line 335: I don't think this is a thing, unless you can point me to a spec which encodes the claims directly in the `id_token` field as a JSON string

crates/storage-pg/migrations/20241202123523_upstream_oauth_responses_alg.sql
```diff
@@ -0,0 +1,10 @@
+-- Copyright 2024 New Vector Ltd.
+--
+-- SPDX-License-Identifier: AGPL-3.0-only
+-- Please see LICENSE in the repository root for full details.
+
+-- Add columns to upstream_oauth_providers to specify the
+-- expected signing algorithm for the endpoint JWT responses.
+ALTER TABLE "upstream_oauth_providers"
+  ADD COLUMN "id_token_signed_response_alg" TEXT DEFAULT 'RS256',
+  ADD COLUMN "userinfo_signed_response_alg" TEXT DEFAULT 'RS256';
```
sandhose, line 10:
```suggestion
  ADD COLUMN "id_token_signed_response_alg" TEXT NOT NULL DEFAULT 'RS256',
  ADD COLUMN "userinfo_signed_response_alg" TEXT;
```

sandhose (changes requested):
Thanks for this!

A few things to fix:

 - `id_token_signed_response_alg` shouldn't be optional, I haven't seen that specced anywhere
 - `userinfo_signed_response_alg` should default to `None`, as it's the most common case

crates/storage/src/upstream_oauth2/provider.rs
```diff
@@ -42,11 +42,24 @@ pub struct UpstreamOAuthProviderParams {
...
     /// Whether to fetch the user profile from the userinfo endpoint,
     /// or to rely on the data returned in the `id_token` from the
     /// `token_endpoint`.
     pub fetch_userinfo: bool,
 
+    /// Expected signature for the JWT payload returned by the userinfo
+    /// endpoint.
+    ///
+    /// If `None`, the response is expected to be an unsigned JSON payload.
+    /// Defaults to `RS256`.
```
sandhose, line 60:
```suggestion
    /// Defaults to `None`.
```

sandhose (changes requested):
Almost there, just still would like to have the `userinfo_signed_response_alg` default to `Option::None`

sandhose (approved):
LGTM, thanks a lot!

PR #3672, opened 2024-12-13
Shutdown the server if any of the tasks crashes

### LINKED ISSUE #2625: Losing the connection to Postgres stops the background jobs
*This issue was originally created by [**@sandhose**](https://github.com/sandhose) at <https://github.com/matrix-org/matrix-authentication-service/issues/2625>.*

When the connection to the database is lost, the background workers stop getting new jobs, even when the connection is back. There is no good indication when this happens, and the only way to fix it is to restart the service.

We should investigate to make the background jobs more reliable

sandhose:
This is to avoid to be in a state where the activity tracker, the task worker or the listener crashed, but the app is still running.

This works by having each task have a drop guard on a cancellation token, meaning that if the task is being dropped, the cancellation token is cancelled, which will cause other tasks to stop as well.
If the service is stopeed that way, it will also use a failure exit code to indicate that the service has been stopped unexpectedly.

Fixes #2625

### DIFF
--- a/crates/cli/src/shutdown.rs
+++ b/crates/cli/src/shutdown.rs
@@ -3,7 +3,7 @@
 // SPDX-License-Identifier: AGPL-3.0-only
 // Please see LICENSE in the repository root for full details.
 
-use std::time::Duration;
+use std::{process::ExitCode, time::Duration};
 
 use tokio::signal::unix::{Signal, SignalKind};
 use tokio_util::{sync::CancellationToken, task::TaskTracker};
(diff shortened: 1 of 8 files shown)


crates/cli/src/shutdown.rs
```diff
@@ -112,5 +120,11 @@ impl ShutdownManager {
         self.task_tracker().wait().await;
 
         tracing::info!("All tasks are done, exitting");
+
+        if likely_crashed {
+            ExitCode::FAILURE
+        } else {
+            ExitCode::SUCCESS
```
reivilibre, line 127: I guess that's orthogonal to this PR, but is exiting with SUCCESS really the right thing to do if exiting due to a SIGTERM or so?

It's possible the SIGTERM was sent by another process due to an out of memory situation or something like that, but then a successful exit code doesn't necessarily trigger a restart in e.g. systemd by default. (I've had this failure mode with other services in the past.)

I mostly just wonder what the convention is for this.

sandhose: Hmmm SIGTERM is usually sent when asking for a graceful shutdown. Docker/Kubernetes does that, systemd probably as well

PR #3688, opened 2024-12-16
Automatic release notes out of PR labels

sandhose:
This is a start to make the release process easier. The current categories are a bit abritrary, will see how it goes.

To see how it would look like, draft a new release, set the tag to `v0.13.0`, target the `quenting/automatic-release-notes` branch and click 'generate release notes'

### DIFF
--- a/.github/release.yml
+++ b/.github/release.yml
@@ -1,0 +1,40 @@
+changelog:
+  categories:
+    - title: Bug Fixes
+      labels:
+        - T-Defect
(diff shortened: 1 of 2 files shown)


reivilibre (approved):
SGTM as a first cut. Time will tell if we want any improvements. But I'm not against having the PR titles (presumably) being == the changelog entries (this is what I do in Synapse with a script anyway)

PR #3700, opened 2024-12-17
Allow response_mode to be null and if so do not add the query param

MatMaul:


crates/cli/src/sync.rs
```diff
@@ -235,13 +235,17 @@ pub async fn config_sync(
                 }
             };
 
-            let response_mode = match provider.response_mode {
```
sandhose, line 238: Using `provider.response_mode.map(…)` would help simplify this

crates/handlers/src/oauth2/authorization/mod.rs
```diff
@@ -106,18 +106,25 @@ fn resolve_response_mode(
...
+        // If the response type includes either "token" or "id_token", the default
+        // response mode is "fragment" and the response mode "query" must not be
+        // used
+        if let Some(suggested_response_mode) = suggested_response_mode {
+            match suggested_response_mode {
+                M::Query => Err(RouteError::InvalidResponseMode),
+                mode => Ok(mode),
+            }
+        } else {
+            Ok(M::Fragment)
```
sandhose, line 119: It's unclear to me why this has to change? This code is for the client -> MAS login part, and it is spec-compliant, nothing to do with the upstream provider stuff

crates/handlers/src/upstream_oauth2/callback.rs
```diff
@@ -184,25 +187,46 @@ pub(crate) async fn handler(
...
+                }
+                Some(UpstreamOAuthProviderResponseMode::Query) | None => {
+                    return Err(RouteError::InvalidResponseMode {
+                        expected: UpstreamOAuthProviderResponseMode::FormPost,
+                    })
+                }
+            };
+        }
+        method => {
+            return Err(RouteError::InvalidReqMethod { method });
```
sandhose, line 228: This should never happen, the route is installed as POST and GET handler anyway

crates/handlers/src/upstream_oauth2/callback.rs
```diff
@@ -184,25 +187,46 @@ pub(crate) async fn handler(
...
-            if sessions_cookie.is_empty() && !params.did_mas_repost_to_itself {
-                let params = Params {
-                    did_mas_repost_to_itself: true,
-                    ..params
-                };
-                let context = FormPostContext::new_for_current_url(params).with_language(&locale);
-                let html = templates.render_form_post(&context)?;
-                return Ok(Html(html).into_response());
-            }
+    match method {
```
sandhose, line 190: I'm finding the old single match on the tuple easier to read. Could you go back to using that?

crates/storage-pg/migrations/20241212154426_oauth2_response_mode_null.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE "upstream_oauth_providers" ALTER COLUMN "response_mode" DROP NOT NULL;
```
sandhose, line 1: Can you add the copyright header + a short description of what this migration does?

sandhose (changes requested):
Just a few style points, otherwise looks good!

sandhose (approved):

PR #3719, opened 2024-12-19
Allow longer & shorter usernames, complying with the MXID length spec

### LINKED ISSUE #2454: Why does upstream OAuth registration limit username to 14 characters?
*This issue was originally created by [**@eforest**](https://github.com/eforest) at <https://github.com/matrix-org/matrix-authentication-service/issues/2454>.*

Why does upstream OAuth registration limit username to 14 characters?
Is there any reason regarding integration with matrix-home-server?

see https://github.com/matrix-org/matrix-authentication-service/blob/main/policies/register.rego [21-23]

Best Regards

### LINKED ISSUE #2620: Upstream OAuth registration requires usernames of length 3 or more, Matrix Spec allows 1 or more
*This issue was originally created by [**@thegcat**](https://github.com/thegcat) at <https://github.com/matrix-org/matrix-authentication-service/issues/2620>.*

We unfortunately have a user with a 2-character username that is currently unable to register with our Matrix server. This seems to be cause by the registration policy https://github.com/matrix-org/matrix-authentication-service/blob/main/policies/register.rego#L17-L19 requiring usernames to be of a length between 3 and 64. The [MXID spec](https://spec.matrix.org/v1.10/appendices/#user-identifiers) however allows the local part of the ID (i.e. the username) to be as short as 1 character, and only requires that the length of the whole MXID not be over 255 characters.

A stop-gap solution for us would be to remove the lower bound and just check that a username is present at all. The better solution would probably be to rework this part of the policy to adhere to the Matrix spec.

Thank you for your help.

sandhose:
This adds the `server_name` to the policy data, which lets us compute the MXID, which means we can accurately make sure we allow MXIDs up to 255 characters long.
It also makes the minimal username length 1 instead of 3.

I also made it so that we get 'error codes' from the policy violations, which lets us translate more specific error messages

Fixes #2454
Fixes #2620

### DIFF
--- a/crates/cli/src/commands/debug.rs
+++ b/crates/cli/src/commands/debug.rs
@@ -8,7 +8,7 @@ use std::process::ExitCode;
 
 use clap::Parser;
 use figment::Figment;
-use mas_config::{ConfigurationSectionExt, PolicyConfig};
+use mas_config::{ConfigurationSection, ConfigurationSectionExt, MatrixConfig, PolicyConfig};
 use tracing::{info, info_span};
 
 use crate::util::policy_factory_from_config;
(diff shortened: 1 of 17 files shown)


reivilibre (approved):
are these policies overridable by the admin? It doesn't seem sensible to let an admin override the maximum length (or alphabet check, particularly if it might allow unicode user IDs).
I would say to be careful that `count` counts bytes and not 'characters', but if we restrict to the restricted set of ASCII then that's fine.

sandhose:
> are these policies overridable by the admin? It doesn't seem sensible to let an admin override the maximum length (or alphabet check, particularly if it might allow unicode user IDs).

It's anyway calling Synapse's 'username available' endpoint, which checks that as well. This is mainly making the error better for the end user

PR #3752, opened 2025-01-06
Use taiki-e/install-action for installing some tools in CI

sandhose:
This unifies a few installs we're doing, and will make sure we get up-to-date tools. I'm tempted to pin the action, but that would probably induce a lot of dependabot update for it, as any tool update means they release a patch version (even for tools we don't use)


reivilibre (approved):
works for me.

I thought I recognised the name somewhere: the author is also the maintainer of pin-project!

Issue #3032, opened 2024-09-10
Include an invisible username field in password change forms


matrixbot:
*This issue was originally created by [**@sandhose**](https://github.com/sandhose) at <https://github.com/matrix-org/matrix-authentication-service/issues/3032>.*

We had that originally in the server-side rendered password recovery form: if you include a hidden `<input type="text" value="alice" autocomplete="username" />` so that password manager can detect which account the password was just changed.

This means:

 - we need an anonymous GraphQL API to fetch infos about the recovery ticket
 - we need to implement that on the password recovery form
 - we need to implement that on the password change form (easier because we already know the user here)

PR #3780, opened 2025-01-07
Polish the password recovery flow and other small design tweaks

sandhose:
Fixes https://github.com/element-hq/matrix-authentication-service/issues/3032

This can be reviewed commit by commit:
- **Render an earlier loading screen fallback** shows a spinner earlier
- **Fix the cross signing reset cancel and error screens** as there were some layout issues
- **Typo in the consent screen**
- **Make the rate limiter available to the GraphQL API handlers** so that we can rate limit the email recovery resend
- **Polish the password recovery page**, which includes:
  - showing an error message if the recovery link is expired, with a button to resend the email
  - showing an error message if the recovery link has already been used
  - including an invisible username field in the form, so that password managers can save the new password
- **Better collapsible sections**, in sync with the [latest Figma designs](https://www.figma.com/design/66PalsLxjI7v0UAWF1bGE0?node-id=132-12863#1083011931)
- **Adjust layout spacings**, to make them more in line with the designs

### DIFF
--- a/crates/handlers/src/graphql/mutations/user.rs
+++ b/crates/handlers/src/graphql/mutations/user.rs
@@ -1,4 +1,4 @@
-// Copyright 2024 New Vector Ltd.
+// Copyright 2024, 2025 New Vector Ltd.
 // Copyright 2023, 2024 The Matrix.org Foundation C.I.C.
 //
 // SPDX-License-Identifier: AGPL-3.0-only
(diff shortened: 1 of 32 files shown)


crates/handlers/src/graphql/mutations/user.rs
```diff
@@ -323,6 +328,61 @@ impl SetPasswordPayload {
     }
 }
 
+/// The input for the `resendRecoveryEmail` mutation.
+#[derive(InputObject)]
+pub struct ResendRecoveryEmailInput {
+    /// The recovery ticket to use.
+    ticket: String,
+}
```
reivilibre, line 336: I don't understand the intended flow for this.

I assumed the recovery ticket is the code that arrives in the e-mail (embedded in a URL, probably) and that lets you recover the account.

So why would you use a recovery ticket to resend the recovery e-mail?

If the ticket is more public than that, I don't think I am happy with the idea that you can use a ticket to request the username of the user...

edit: from the client code I think I gleam that it's for expired tickets. If so, can you state that more clearly in the documentation for the resend operation?

reivilibre (changes requested):
just one (probably small) thing

crates/handlers/src/graphql/mutations/user.rs
sandhose: That's right, I've added a comment in 8a69c9aa93b54894b42777be57d215ef03eb1c51

PR #3790, opened 2025-01-13
Add 'introspection_endpoint' to homeserver config example

Stogas:
This value was previously undocumented, but helpful in our case to access MAS within the same K8s cluster


sandhose (approved):
I've reworded a little bit so that we can merge this, thanks again for this!

PR #3797, opened 2025-01-14
Recognise macaroons as access tokens from Synapse

### LINKED ISSUE #2150: Migration advisor doesn't detect macraoon-like tokens
*This issue was originally created by [**@sandhose**](https://github.com/sandhose) at <https://github.com/matrix-org/matrix-authentication-service/issues/2150>.*

Very old macaroon access tokens will not work with MAS, the migration advisor needs to warn about them

reivilibre:
A short and sweet one!

Closes #2150 

p.s. I have verified on matrix.org that all tokens start with either `MDAx` or `syt_`.

### DIFF
--- a/crates/data-model/src/tokens.rs
+++ b/crates/data-model/src/tokens.rs
@@ -4,6 +4,7 @@
 // SPDX-License-Identifier: AGPL-3.0-only
 // Please see LICENSE in the repository root for full details.
 
+use base64ct::{Base64Url, Encoding};
 use chrono::{DateTime, Utc};
 use crc::{Crc, CRC_32_ISO_HDLC};
 use mas_iana::oauth::OAuthTokenTypeHint;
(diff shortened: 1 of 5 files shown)


crates/data-model/src/tokens.rs
```diff
@@ -294,7 +294,8 @@ impl TokenType {
     pub fn check(token: &str) -> Result<TokenType, TokenFormatError> {
         // these are legacy tokens imported from Synapse
         // we don't do any validation on them and continue as is
-        if token.starts_with("syt_") {
+        // Tokens starting `MDAx` are macaroons from even older versions of Synapse.
```
sandhose, line 297: If I'm being pedantic for a second, macaroons are a list of hex length plus (key space value), then HMAC'd and base64'd.

The first K/V is `location`/servername, which means:

 - it will start with `000flocation ` for HS with a single character
 - it be `001Nlocation ` (with N varying) for HS with 2 to 17 characters
 - etc.

So the prefix `MDAx` works because it's the base64 serialisation of `001`. The `MDA` prefix would work for server names up to 145 characters.

If we want to be extra sure, we would:

 - base64url decode the string
 - check on the result that byte 5 to 13 are exactly `6c 6f 63 61 74 69 6f 6e` (`location`)

sandhose (changes requested):

crates/data-model/src/tokens.rs
reivilibre: fair for digging in to the format! I wasn't able to find (and still haven't found) the spec for Macaroons.
I probably should have looked at some code at least

sandhose (approved):

PR #3800, opened 2025-01-15
Add some tests for the syn2mas `MasWriter`

reivilibre:


crates/syn2mas/src/mas_writer/mod.rs
```diff
@@ -667,5 +667,185 @@ impl<'writer, 'conn> MasUserWriteBuffer<'writer, 'conn> {
...
+        ($db: expr) => {
+            let db_snapshot = snapshot_database($db).await;
+            ::insta::assert_yaml_snapshot!(db_snapshot);
+        };
+    }
+
+    /// Runs some code with a `MasWriter`.
+    ///
+    /// The callback is responsible for `finish`ing the `MasWriter`.
+    async fn with_mas_writer<F, R>(pool: &PgPool, callback: F) -> R
```
sandhose, line 766: Couldn't that be simplified with a function which initialises and returns a `MasWriter`? or is it because it borrows the `main_conn`?
In which case you could solve this by passing the main connection in arguments?

reivilibre: good point, this started off because I needed to make a few conns but then afterwards I needed a connection to do the snapshot anyway, so in the new commit I just use that same connection as the MasWriter connection. That is quite a bit nicer (though 90% of that is because async closures are way more painful than they should be, for understandable reasons but still..)

sandhose (approved):

PR #3801, opened 2025-01-15
Support compatibility sessions that do not have devices

### LINKED ISSUE #2149: Migration advisor does not warn about device-less tokens
*This issue was originally created by [**@sandhose**](https://github.com/sandhose) at <https://github.com/matrix-org/matrix-authentication-service/issues/2149>.*

The migrator does, but not the advisor.
The error message is also not very actionable

reivilibre:
This is needed for full support of existing Synapse access tokens

Closes #2149 by supporting deviceless tokens instead

Not sure if all the design choices here are right:

- we don't sync deviceless sessions with Synapse
- we don't add a device scope token for deviceless sessions

### DIFF
--- a/crates/storage-pg/src/compat/session.rs
+++ b/crates/storage-pg/src/compat/session.rs
@@ -100,7 +100,7 @@ impl TryFrom<CompatSessionLookup> for CompatSession {
 #[enum_def]
 struct CompatSessionAndSsoLoginLookup {
     compat_session_id: Uuid,
-    device_id: String,
+    device_id: Option<String>,
     user_id: Uuid,
     user_session_id: Option<Uuid>,
     created_at: DateTime<Utc>,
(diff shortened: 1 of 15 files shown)


crates/storage-pg/src/compat/session.rs
```diff
@@ -118,12 +122,12 @@ impl TryFrom<CompatSessionAndSsoLoginLookup> for (CompatSession, Option<CompatSs
 
     fn try_from(value: CompatSessionAndSsoLoginLookup) -> Result<Self, Self::Error> {
         let id = value.compat_session_id.into();
-        let device = Device::try_from(value.device_id).map_err(|e| {
+        let device = Some(Device::try_from(value.device_id).map_err(|e| {
```
sandhose, line 125: That doesn't feel right, shouldn't `value.device_id` be nullable?

crates/tasks/src/matrix.rs
```diff
@@ -208,7 +208,11 @@ impl RunnableJob for SyncDevicesJob {
                 .map_err(JobError::retry)?;
 
             for (compat_session, _) in page.edges {
-                devices.insert(compat_session.device.as_str().to_owned());
+                let Some(ref device) = compat_session.device else {
+                    // Do not sync compat sessions without devices
+                    continue;
+                };
+                devices.insert(device.as_str().to_owned());
```
sandhose, line 215: We should still update the cursor, else if we end up with a page with device-less sessions, the cursor won't be updated and it will loop indefinitely 
```suggestion
                if let Some(ref device) = compat_session.device {
                    devices.insert(device.as_str().to_owned());
                };
```

sandhose (changes requested):

sandhose (approved):

PR #3805, opened 2025-01-17
Add pre-migration checks to syn2mas

reivilibre:
As far as I know, this matches or exceeds `advisor.mts` from the old tool.


crates/syn2mas/src/synapse_reader/checks.rs
```diff
@@ -0,0 +1,290 @@
...
+            };
+
+            // unwrap: we already check the issuer is present before accepting the config dict
+            let issuer = matching_syn.issuer.clone().unwrap();
+
+            // TODO extract this logic and reuse it for the migration
+            let matching_mas = mas_oauth2
+                .providers
+                .iter()
+                .find(|mas_provider| mas_provider.issuer == issuer);
```
sandhose, line 277: we've talked about the heuristic for matching the providers on both sides; I'll assume that when you implement that in the migration, you'll implement what we talked about (having a `synapse_idp_id` in the MAS config to match) and use it here as well

crates/syn2mas/src/synapse_reader/checks.rs
```diff
@@ -0,0 +1,290 @@
...
+        query_scalar("SELECT COUNT(1) FROM user_threepids WHERE medium <> 'email'")
+            .fetch_one(&mut *synapse_connection)
+            .await?;
+    if num_non_email_3pids > 0 {
+        errors.push(CheckError::NonEmailThreepidsInDatabase {
+            num_non_email_3pids,
+        });
+    }
+
+    let oauth_provider_user_counts = query_as::<_, UpstreamOAuthProvider>("SELECT auth_provider, COUNT(*) AS num_users FROM user_external_ids GROUP BY auth_provider ORDER BY auth_provider").fetch_all(&mut *synapse_connection).await?;
```
sandhose, line 250: could you make the query a multi-line string, it's pretty long

crates/syn2mas/src/synapse_reader/checks.rs
```diff
@@ -0,0 +1,290 @@
...
+}
+
+/// A potential hazard found whilst checking the Synapse database, that should be presented
+/// to the operator to check they are aware of a caveat before proceeding with the migration.
+#[derive(Debug, Error)]
+pub enum CheckWarning {
+    #[error("Synapse config contains OIDC auth configuration (issuer: {issuer:?}) which will need to be manually mapped to an upstream OpenID Connect Provider during migration.")]
+    UpstreamOidcProvider { issuer: String },
+
+    #[error("Synapse config contains {0} auth configuration which will need to be manually mapped as an upstream OpenID Connect Provider during migration.")]
```
sandhose, line 84:
```suggestion
    #[error("Synapse config contains {0} auth configuration which will need to be manually mapped as an upstream OAuth 2.0 provider during migration.")]
```

crates/syn2mas/src/synapse_reader/checks.rs
```diff
@@ -0,0 +1,290 @@
...
+
+    if synapse_config.enable_registration {
+        warnings.push(CheckWarning::DisableRegistrationAfterMigration);
+    }
+    if synapse_config.enable_registration_captcha {
+        warnings.push(CheckWarning::ShouldPortRegistrationCaptcha);
+    }
+    if synapse_config.user_consent {
+        warnings.push(CheckWarning::DisableUserConsentAfterMigration);
+    }
```
sandhose, line 111: I wonder if for those two we should look at the MAS config? Checking whether CAPTCHA is enabled is straightforward. User consent, the only equivalent we have in MAS is the `terms_url`

crates/syn2mas/src/synapse_reader/checks.rs
```diff
@@ -0,0 +1,290 @@
...
+        .load()
+        .await
+        .map_err(Error::MasPasswordConfig)?;
+
+    let mas_matrix = MatrixConfig::extract(mas)?;
+
+    // Look for the MAS password hashing scheme that will be used for imported Synapse passwords,
+    // then check the configuration matches so that Synapse passwords will be compatible with MAS.
+    let mut found_password_scheme = false;
+    for (scheme_version, scheme_algorithm, _scheme_rounds, scheme_secret) in &mas_password_schemes {
```
sandhose, line 175: I would do something like

```rust
if let Some((_, alg, _, secret)) = mas_password_schemes.find(|(version, _, _, _)| version == MIGRATED_PASSWORD_VERSION) {
  …
} else {
  errors.push(CheckError::MissingPasswordScheme);
}
```

crates/syn2mas/src/synapse_reader/checks.rs
```diff
@@ -0,0 +1,290 @@
...
+    #[derive(FromRow)]
+    struct UpstreamOAuthProvider {
+        auth_provider: String,
+        num_users: i64,
+    }
+
+    let mut errors = Vec::new();
+    let warnings = Vec::new();
+
+    let num_guests: i64 = query_scalar("SELECT COUNT(1) FROM users WHERE is_guest <> 0")
```
sandhose, line 233: I wonder whether we should disable inactive guests here? Like the server may have historical guests but they don't matter? That would be the case on m.org for example

crates/syn2mas/src/synapse_reader/checks.rs
```diff
@@ -0,0 +1,290 @@
...
+            let Some(matching_syn) = matching_syn else {
+                errors.push(CheckError::SynapseMissingOAuthProvider {
+                    provider: row.auth_provider,
+                    num_users: row.num_users,
+                });
+                continue;
+            };
+
+            // unwrap: we already check the issuer is present before accepting the config dict
+            let issuer = matching_syn.issuer.clone().unwrap();
```
sandhose, line 271: use `.expect("caller to make sure the issuer is set")` or something like that

sandhose (changes requested):

crates/syn2mas/src/synapse_reader/config.rs
```diff
@@ -0,0 +1,166 @@
...
+    /// Load a Synapse configuration from the given list of configuration files.
+    ///
+    /// # Errors
+    ///
+    /// - If there is a problem reading any of the files.
+    /// - If the configuration is not valid.
+    pub fn load(files: &[Utf8PathBuf]) -> Result<Config, figment::Error> {
+        let mut figment = figment::Figment::new();
+        for file in files {
+            figment = figment.merge(Yaml::file(file));
```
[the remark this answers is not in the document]
reivilibre: argh, this doesn't match; Synapse only operates on top-level keys (later = preferred)

(c.f. https://github.com/element-hq/synapse/blob/develop/synapse/config/_base.py#L870)

crates/cli/src/commands/syn2mas.rs
```diff
@@ -1,16 +1,23 @@
...
 use rand::thread_rng;
 use sqlx::{Connection, Either, PgConnection};
-use syn2mas::{LockedMasDatabase, MasWriter, SynapseReader};
+use syn2mas::{synapse_config, LockedMasDatabase, MasWriter, SynapseReader};
 use tracing::{error, warn};
 
 use crate::util::database_connection_from_config;
 
+/// The exit code used by `syn2mas check` and `syn2mas migrate` when there are errors preventing migration.
+const EXIT_CODE_CHECK_ERRORS: u8 = 10;
```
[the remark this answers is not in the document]
reivilibre: indeed

crates/syn2mas/src/synapse_reader/checks.rs
reivilibre: fb8c3370ce0b169a948a07599bc9e97338686ea7

crates/syn2mas/src/synapse_reader/checks.rs
reivilibre: 1c6914616bbce812f9c61c5379840376c27e7bb5

For user consent we still emit a warning when enabled since I guess we don't want Synapse to lock out the user until it's been agreed on the Synapse side.

crates/syn2mas/src/synapse_reader/checks.rs
reivilibre: 9f8fbad79d5b60b9a4471904f5b9c282a1c72712

crates/syn2mas/src/synapse_reader/checks.rs
reivilibre: This so far is a copy of the existing tool for this aspect. My intention here is to disable this error once we import the bare minimum guest metadata that lets us not clash with guest IDs in the future?
Whichever way you look at it, we need to remove this error before we do matrix.org :-)

sandhose: Alright, let's tackled that in another PR :)

crates/syn2mas/src/synapse_reader/config.rs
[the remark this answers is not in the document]
sandhose: It's probably fiiiine for now, can you just stick a TODO somewhere to make sure we don't forget that we don't exactly match synapse's behaviour?

sandhose (approved):

PR #3809, opened 2025-01-20
Configure nextest to run database tests serially

sandhose:
This should resolve the flakes we're regularly seeing in CI by running all the database tests serially.
To avoid getting too much of a speed penalty, I've also added another shard to the test runs in CI

This is necessary until https://github.com/launchbadge/sqlx/pull/3334 gets merged


reivilibre (approved):
SGTM, seems like nextest has some nice flexibility

PR #3817, opened 2025-01-20
Create release branches as part of the release workflow

sandhose:
The intended workflow is as follows:

 - `release-branch.yaml`:
    - creates a new RC tag for the next pre-release
    - pushed to the main branch
    - and pushed to a `release/vX.Y` branch
    - creates a `vX.Y` branch in Localazy. Upload and download actions were adapted to use that branch
 - `release-bump.yaml` bumps a version on a release branch
   - if it has `RC` set it will do a new RC
   - if not it will do a stable release (next patch version if we were not on an RC yet)
   - if it has `Merge back` set, it will open a PR `release/vX.Y` -> `main`
 - the git tag creates the usual docker image + GitHub draft release through the `build.yaml` workflow

One flaw of this workflow, is that the docker `latest` tag is updated every time we publish a 'stable' release, so if we decide to bump on an older release branch, it will set that as the latest release tag

<details>
<summary>For future reference if we want to fix that</summary>

```
> git tag -l 'v*' | xargs npx semver --preid rc
0.0.0-test.0
0.0.0-test.1
0.1.0
0.2.0
0.3.0
0.4.0
0.4.1
0.5.0-rc.0
0.5.0-rc.1
0.5.0-rc.2
0.5.0
0.6.0
0.6.1
0.7.0
0.8.0-rc.0
0.8.0-rc.1
0.8.0
0.9.0-rc.0
0.9.0-rc.1
0.9.0
0.10.0-rc.0
0.10.0
0.11.0-rc.0
0.11.0
0.12.0-rc.0
0.12.0-rc.1
0.12.0
```

</details>

### DIFF
--- a/.github/workflows/release-branch.yaml
+++ b/.github/workflows/release-branch.yaml
@@ -1,0 +1,76 @@
+name: Create a new release branch
+on:
+  workflow_dispatch:
+
+jobs:
+  compute-version:
+    name: Compute the next minor RC version
+    runs-on: ubuntu-22.04
+
+    permissions:
+      contents: read
+
+    outputs:
+      version: ${{ steps.next.outputs.version }}
+
+    steps:
+      - name: Fail the workflow if this is not the main branch
+        if: ${{ github.ref_name != 'main' }}
+        run: exit 1
+
+      - name: Checkout the code
+        uses: actions/checkout@v4.2.2
+
+      - name: Install Rust toolchain
+        run: |
+          rustup toolchain install stable
+          rustup default stable
+
+      - name: Extract the current version
+        id: current
+        run: echo "version=$(cargo metadata --format-version 1 | jq -r '.packages[] | select(.name == "mas-cli") | .version')" >> "$GITHUB_OUTPUT"
+
+      - name: Compute the new minor RC
+        id: next
+        run: echo "version=$(npx --yes semver@7.5.4 -i preminor --preid rc "${{ steps.current.outputs.version }}")" >> "$GITHUB_OUTPUT"
+
+  tag:
+    uses: ./.github/workflows/tag.yaml
+    needs: compute-version
+    with:
+      version: ${{ needs.compute-version.outputs.version }}
+    secrets:
+      BOT_GITHUB_TOKEN: ${{ secrets.BOT_GITHUB_TOKEN }}
+
+  branch:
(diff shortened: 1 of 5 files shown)


.github/workflows/release-branch.yaml
```diff
@@ -0,0 +1,104 @@
...
+        if: ${{ github.ref_name != 'main' }}
+        run: exit 1
+
+      - name: Checkout the code
+        uses: actions/checkout@v4.2.2
+
+      - name: Install Rust toolchain
+        run: |
+          rustup toolchain install stable
+          rustup default stable
```
reivilibre, line 28: I guess this is idiomatic for GHA but it feels suboptimal that we have to install Rust imperatively/slowly, particularly for just ... running `cargo metadata`.
When I used to self-host Drone/Woodpecker CI, this would have been done by running the command on a Rust container image that already had it installed.

Separately but also along the same lines: how intensive is `cargo metadata` — does it need to fetch all dependency crates? That would seem suboptimal just for reading a version.
Maybe there's a TOML version of `jq` that we could use to just pull it out of `mas-cli/Cargo.toml`? :-), or some `toml2json | jq`?

.github/workflows/release-bump.yaml
```diff
@@ -0,0 +1,76 @@
...
+        uses: actions/checkout@v4.2.2
+
+      - name: Install Rust toolchain
+        run: |
+          rustup toolchain install stable
+          rustup default stable
+
+      - name: Extract the current version
+        id: current
+        run: echo "version=$(cargo metadata --format-version 1 | jq -r '.packages[] | select(.name == "mas-cli") | .version')" >> "$GITHUB_OUTPUT"
```
reivilibre, line 40: (ditto)

reivilibre (approved):
This looks good.

Just a perfectionist point (though maybe there's a worthwhile speedup in there).

Maybe also: would it be worth having a dev doc page on how to build a release, even just bullet points of 'go and run X workflow, then merge the PR, then do this...'?

.github/workflows/release-branch.yaml
sandhose: That's fair, although the current release process took about 30s, including the rustup install, so I'm not too concerned about this slowing down the release :)

PR #3835, opened 2025-01-21
Publish nightly builds in a release

sandhose:
This should create and maintain a 'nightly' release, with the latest build artifacts, as well as up-to-date references to the docker images.

This should help people running the latest version of the service


sandhose:
Thanks for the review! Did rename nightly -> unstable

PR #3854, opened 2025-01-23
Upgrade Rust to 1.84 and fix new clippy lints

sandhose:


crates/storage-pg/src/queue/schedule.rs
```diff
@@ -69,7 +69,7 @@ impl QueueScheduleRepository for PgQueueScheduleRepository<'_> {
             ScheduleLookup,
             r#"
                 SELECT
-                    queue_schedules.schedule_name,
+                    queue_schedules.schedule_name as "schedule_name!",
```
sandhose, line 72: Not sure why, but my local sqlx kept complaining about this, even though it's supposed to be NOT NULL already ;(

PR #3878, opened 2025-01-24
syn2mas: Migrate threepids to MAS

reivilibre:
Last two commits only; need to merge #3805 and do a rebase

### DIFF
--- a/crates/syn2mas/src/migration.rs
+++ b/crates/syn2mas/src/migration.rs
@@ -17,12 +17,16 @@ use futures_util::StreamExt as _;
 use rand::RngCore;
 use thiserror::Error;
 use thiserror_ext::ContextInto;
+use tracing::Level;
 use ulid::Ulid;
 use uuid::Uuid;
 
 use crate::{
-    mas_writer::{self, MasNewUser, MasNewUserPassword, MasUserWriteBuffer, MasWriter},
-    synapse_reader::{self, ExtractLocalpartError, FullUserId, SynapseUser},
+    mas_writer::{
+        self, MasNewEmailThreepid, MasNewUnsupportedThreepid, MasNewUser, MasNewUserPassword,
+        MasThreepidWriteBuffer, MasUserWriteBuffer, MasWriter,
+    },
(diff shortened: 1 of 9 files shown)


crates/syn2mas/src/migration.rs
```diff
@@ -126,6 +139,63 @@ async fn migrate_users(
...
+            added_at,
+        } = threepid_res.into_synapse("reading threepid")?;
+        let created_at: DateTime<Utc> = created_at.into();
+
+        let username = synapse_user_id
+            .extract_localpart(server_name)
+            .into_extract_localpart(synapse_user_id.clone())?
+            .to_owned();
+        let Some(user_id) = user_localparts_to_uuid.get(username.as_str()).copied() else {
+            todo!()
```
sandhose, line 167: K-boom! This just made me think of something, what happens if one of task panics? Does everything shutdown properly?

crates/syn2mas/src/synapse_reader.rs
```diff
@@ -122,9 +122,9 @@ impl<'r> sqlx::Decode<'r, Postgres> for SecondsTimestamp {
     fn decode(
         value: <Postgres as sqlx::Database>::ValueRef<'r>,
     ) -> Result<Self, sqlx::error::BoxDynError> {
-        <i64 as sqlx::Decode<Postgres>>::decode(value).map(|milliseconds_since_epoch| {
+        <i64 as sqlx::Decode<Postgres>>::decode(value).map(|seconds_since_epoch| {
```
sandhose, line 125: 🙈

sandhose (approved):

crates/syn2mas/src/migration.rs
reivilibre: good thinking!

---

For the writer tasks:

It doesn't shut down properly, but depending on the exact failure:

- if all tasks panic, I suspect it will generally get stuck waiting for them to be ready to accept new tasks
- if not all of them panic, it will be detected that some panicked — at the end

I'll open an issue to solve that, I should have thought about it.

---

For the main task:

As far as I can see, we don't need to do anything special here, since the writer tasks will eventually finish their current operation. None of the transactions will get committed.

PR #3893, opened 2025-01-27
Upgrade axum to 0.8

sandhose:
This basically has three changes:

 - removing all the #[async_trait] attributes
 - replacing optional extractors with alternatives
 - changing the paths in the router to use the new `{notation}` instead of the old `:notation`


reivilibre (approved):
thanks for doing this, nice

PR #3909, opened 2025-01-29
Buffer reading of translation files

sandhose:
This speeds up loading of templates by a bunch (60ms -> 5ms), which affects startup and reload times


reivilibre (approved):
Always nice to improve startup/restart times :)

PR #3917, opened 2025-01-29
syn2mas: Support migrating external IDs as upstream OAuth2 providers

reivilibre:
Commit-by-commit suggested

### DIFF
--- a/crates/syn2mas/src/mas_writer/mod.rs
+++ b/crates/syn2mas/src/mas_writer/mod.rs
@@ -10,7 +10,7 @@
 use std::fmt::Display;
 
 use chrono::{DateTime, Utc};
-use futures_util::{future::BoxFuture, TryStreamExt};
+use futures_util::{future::BoxFuture, FutureExt, TryStreamExt};
 use sqlx::{query, query_as, Executor, PgConnection};
 use thiserror::Error;
 use thiserror_ext::{Construct, ContextInto};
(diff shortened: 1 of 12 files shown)


crates/syn2mas/src/mas_writer/mod.rs
```diff
@@ -843,6 +901,60 @@ impl<'writer, 'conn> MasThreepidWriteBuffer<'writer, 'conn> {
     }
 }
 
+/// A function that can accept and flush buffers from a `MasWriteBuffer`.
+/// Intended uses are the methods on `MasWriter` such as `write_users`.
+type WriteBufferFlusher<'conn, T> =
+    for<'a> fn(&'a mut MasWriter<'conn>, Vec<T>) -> BoxFuture<'a, Result<(), Error>>;
```
sandhose, line 776: Riiight, all the box future make sense now :)

crates/syn2mas/src/migration.rs
```diff
@@ -132,7 +132,8 @@ async fn migrate_users(
     server_name: &str,
     rng: &mut impl RngCore,
 ) -> Result<UsersMigrated, Error> {
-    let mut write_buffer = MasUserWriteBuffer::new(mas);
+    let mut user_buffer = MasWriteBuffer::new(MasWriter::write_users);
```
sandhose, line 135: not sure how I feel about passing function pointers instead of using traits… I guess that works, not sure how it would look like with traits

crates/syn2mas/src/mas_writer/mod.rs
```diff
@@ -716,7 +713,53 @@ impl<'conn> MasWriter<'conn> {
...
+                    subject,
+                    created_at,
+                } in links
+                {
+                    link_ids.push(link_id);
+                    user_ids.push(user_id);
+                    upstream_provider_ids.push(upstream_provider_id);
+                    subjects.push(subject);
+                    created_ats.push(created_at);
+                }
```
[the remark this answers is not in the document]
reivilibre: mrmm, it's hard to find a tool that works here other than doing it the boring way. I don't see that much point in destructuring to a tuple. Maybe it's a little better, but not by much

PR #3920, opened 2025-01-29
Fix the account page crash introduced in #3893

sandhose:
Looks like `Query<Option<T>>` is not really a valid extractor, which made Axum show a 500 error on the account page


reivilibre (approved):

reivilibre:
maybe `Option<Query<T>>` would work, but this way is more extensible anyway

sandhose:
> maybe Option<Query<T>> would work, but this way is more extensible anyway

That changed in axum 0.8: the extractor has to implement OptionalFromRequestParts for that. This is because Option<Extractor> would void most errors that you'd actually want to handle

PR #3992, opened 2025-02-07
Copy/paste attempt to make a "get user emails" API

turt2live:
Disclaimer: I don't know rust, and haven't tested this.


sandhose:
Thanks for this, I did a proper implementation in #4001

Main difference in the implementation is that it's not under the `/users` API, but rather in its own path, as a paginated list, with an (optional) `filter[user]=ID`. This will let us introduce other filters down the line more easily (like email provider domain?) and list all the emails on the whole instance

PR #4115, opened 2025-02-25
Admin API to dynamically set policy data

### LINKED ISSUE #3993: Dynamic data for policy engine
To support T&S tooling, the `registration` (and possibly `email`) policy engines will need to be expanded to accept `data` which is supplied over the MAS Admin API. Currently, some data may be provided by the config - these definitions are complimentary to the API approach described here.

It's proposed to have a simple get+set API rather than managing individual rules. Something like:

```
POST /api/admin/v1/policy/data
{"registration": {...}}

> 200 OK
```

```
GET /api/admin/v1/policy/data

> 200 OK
> {"registration": {...}}
```

Where the POST/GET body could be something like:

```json5
{
  "registration": {
    "banned_usernames": {
      "literals": ["alice", "bob"],
      "substrings": ["ali", "ob"],
      "regexes": ["(alice|bob)"]
    },
    "banned_emails": {
      "literals": ["alice@example.org", "bob@example.org"],
      "substrings": ["@example.org"],
      "regexes": ["ali.*@example\\.org"]
    },
    "banned_user_agents": {
      "literals": ["Pretend this is Real;"],
      "substrings": ["Chrome"],
      "regexes": ["Chrome 1.*;"]
    },
    "banned_ips": [
      "192.168.0.1",
      "192.168.1.1/24",
      "fe80::/64"
    ]
  }
}
```

Note that for string-based fields there are 3 possible match types: literals (exact matches), substrings ("does X contain Y?"), and regexs (substrings, but complicated). T&S uses all 3 on all described fields.

For the IP address check, T&S currently uses a [netaddr.IPSet](https://netaddr.readthedocs.io/en/latest/tutorial_03.html) using its constructor form (`IPSet([..strings..])`). CIDR support is of particular interest. 

This rule data is then sent to the policy engine as `data` for violation checks. Only the `registration` policy *really* needs checking, which is why the proposed endpoint nests the data accordingly. If `banned_emails` needs to/should be applied to the `emails` policy, that'd be fine.

If it's preferred to drop `substrings` in favour of regex, that'd be fine. Ideally, `literals` is kept though.

The current `registration` policy will need to be fed a request IP (or IPs, if multiple) and user agent. In future, this may be expanded with client IDs and other client metadata for stronger blocking.

sandhose:
Fixes #3993

This adds a way to dynamically set the policy data through an admin API call.
This data is then stored in the database, and merged with the policy data from the configuration file.
Note that it merges both objects, concatenates arrays, overridding any values with the ones from the dynamic data.

We keep an history in the database of the policy data, so that when looking at the logs, we can look which 'version' of the policy data was used during policy evaluation.

When setting the policy data through the API, it validates it and will refuse anything that it can't load in the WASM module. It will immediately load it in the current instance; other instances load it regularly (every minute) from the database, so there is maximum a 1 minute lag between the admin API being called and it being effective on all workers.


crates/storage-pg/migrations/20250225091000_dynamic_policy_data.sql
```diff
@@ -0,0 +1,11 @@
+-- Copyright 2025 New Vector Ltd.
+--
+-- SPDX-License-Identifier: AGPL-3.0-only
+-- Please see LICENSE in the repository root for full details.
+
+-- Add a table which stores the latest policy data
+CREATE TABLE IF NOT EXISTS policy_data (
+    policy_data_id UUID PRIMARY KEY,
```
reivilibre, line 8: I don't particularly see why we bother to keep around multiple rows and track an UUID on them.

Separately, I could see it being desirable to have different policy data for different policies, so surprised there isn't a stringy key here.

Overall slightly unsure why this isn't 'just' a `TABLE policy_data ( key TEXT PRIMARY KEY NOT NULL, value JSONB NOT NULL )` to be honest! You probably have good reasons but would be nice to put them down

crates/handlers/src/admin/v1/policy_data/get_latest.rs
```diff
@@ -0,0 +1,149 @@
...
+                "hello": "world"
+              }
+            },
+            "links": {
+              "self": "/api/admin/v1/policy-data/01FSHN9AG0MZAA6S4AF7CTV32E"
+            }
+          },
+          "links": {
+            "self": "/api/admin/v1/policy-data/01FSHN9AG0MZAA6S4AF7CTV32E"
+          }
```
reivilibre, line 122: the self-links are already a little verbose, but maybe fair enough. Slight eyebrow raise at it being duplicated, though

reivilibre (changes requested):
seems good, though not sure about the overkillness of the schema?

crates/storage-pg/migrations/20250225091000_dynamic_policy_data.sql
sandhose: I think the main motivation is being able to audit policy decisions? You get the `policy_data_id` logged every time it changes, so you can track what policy data was effective at that time

I also did that for consistency with other tables I guess 🤷 

Note that the policy is a single blob with multiple entrypoint. It shares the same `data` object for all entrypoints, so that's why there is no `key` in the schema here

crates/handlers/src/admin/v1/policy_data/get_latest.rs
sandhose: We're following the stuff from [JSON:API](https://jsonapi.org) here, that's why

reivilibre: to be fair, except for relationships the `links` field is a MAY (optional) in the spec for these cases. But sure, whatever. I don't particularly mind, I also don't really see why it's useful.

crates/storage-pg/migrations/20250225091000_dynamic_policy_data.sql
reivilibre: fair enough.

But we have this `prune` method; when does this get called? (I don't see any calls to it except in the tests for now)

Filling your disk because you're updating a large policy data blob frequently is also something to be avoided :-)

reivilibre (approved):

crates/storage-pg/migrations/20250225091000_dynamic_policy_data.sql
sandhose: Added a comment in 97d2b757c, and you're right, I've meant to add a job to regularly prune stale policy data, and failed to do so. Did that in 8581ca19e

PR #4116, opened 2025-02-26
maint: fix Authentik (>= 2024.2.1) upstream provider

waclaw66:
Since Authentik 2024.2.1 is required to use `offline_access` scope to obtain refresh tokens.

### DIFF
--- a/docs/setup/sso.md
+++ b/docs/setup/sso.md
@@ -190,7 +190,7 @@ upstream_oauth2:
       issuer: "https://<authentik-domain>/application/o/<app-slug>/" # TO BE FILLED
       client_id: "<client-id>" # TO BE FILLED
       client_secret: "<client-secret>" # TO BE FILLED
-      scope: "openid profile email"
+      scope: "openid profile email offline_access"
       claims_imports:
         localpart:
           action: require


docs/setup/sso.md
```diff
@@ -190,7 +190,7 @@ upstream_oauth2:
       issuer: "https://<authentik-domain>/application/o/<app-slug>/" # TO BE FILLED
       client_id: "<client-id>" # TO BE FILLED
       client_secret: "<client-secret>" # TO BE FILLED
-      scope: "openid profile email"
+      scope: "openid profile email offline_access"
```
sandhose, line 193: I'm a little confused by this: MAS doesn't *need* to get a refresh token from Authentik, as it only really needs to check whether the user is valid or not? Does it break right now without it?

waclaw66: Really? Because once I have enabled refresh tokens from OIDC upstream Element Web keeps the session.
I described that issue here https://github.com/element-hq/element-web/issues/27810#issuecomment-2677556581
Maybe it's a coincidence, I will disable refresh tokens again, to see what happens.

waclaw66: Yes, it's really a conincidence, Element Web's session survived tonight not using upstream OIDC refresh tokens without logging out. So please ignore this PR.

PR #4122, opened 2025-02-27
Properly ignore devices, threepids and access tokens from AS users

sandhose:
This reads the AS users from the Synapse database, doesn't import them, but keep them in memory, so that we can properly ignore threepids, external ids, devices and access tokens belonging to them.

### DIFF
--- a/crates/syn2mas/src/migration.rs
+++ b/crates/syn2mas/src/migration.rs
@@ -74,6 +74,7 @@ bitflags::bitflags! {
         const IS_SYNAPSE_ADMIN = 0b0000_0001;
         const IS_DEACTIVATED = 0b0000_0010;
         const IS_GUEST = 0b0000_0100;
+        const IS_APPSERVICE = 0b0000_1000;
     }
 }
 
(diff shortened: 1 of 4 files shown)


crates/syn2mas/src/migration.rs
```diff
@@ -177,6 +182,20 @@ async fn migrate_users(
...
         }
+        if user.appservice_id.is_some() {
+            flags |= UserFlags::IS_APPSERVICE;
+
+            // Special case for appservice users: we don't insert them into the database
+            // We just record the user's information in the state and continue
+            state.users.insert(
+                CompactString::new(&mas_user.username),
+                UserInfo {
+                    mas_user_id: Uuid::nil(),
```
sandhose, line 193: We *could* make this field an [`Option<NonNilUuid>`](https://docs.rs/uuid/latest/uuid/struct.NonNilUuid.html) (that would take the same space in memory) if we wanted to make sure that we never use a nil one?

reivilibre: I like this idea, even if it does mean you will probably need some unwraps or so, it feels more 'direct' to explicitly unwrap than potentially let some accident happen and find out later, probably in a harder to debug way.
That said, the current approach is defensible and not a big deal but avoiding sentinel values has been generally a pretty productive idea for me, so I prefer it as a maintenance safety improvement.

reivilibre (approved):

sandhose:
@reivilibre I did the thing of using NonNilUuids for MAS user IDs in 1e99a6b7e. wdyt? Should I extend that to all MAS ULIDs?

PR #4173, opened 2025-03-10
Properly accumulate form errors on the upstream register page

### LINKED ISSUE #4170: Erroneous error when selecting a username
**Describe the bug**
`This username is already taken` is displayed when an uppercase username value is rejected by matrix/synapse.

**To Reproduce**
1. Set up an external IdP with registration enabled
2. Register a new account using the IdP
3. When returning to matrix authentication service to select a username, try a new unique username containing an uppercase letter
4. Observe the erroneous error message `This username is already taken`
5. Remove the uppercase letter
6. Observe the expected behavior of registering the new username with matrix

**Expected behavior**
The error message should reflect the error from synapse (something like `User ID can only contain characters a-z, 0-9, or '=_-./+'` -- see below)

```
synapse-1  | 2025-03-08 20:19:57,259 - synapse.http.server - 130 - INFO - GET-2294 - <XForwardedForRequest at 0x7f0b3d7a83b0 method='GET' uri='/_synapse/admin/v1/username_available?username=Onecapitalletterisenough' clientproto='HTTP/1.1' site='8008'> SynapseError: 400 - User ID can only contain characters a-z, 0-9, or '=_-./+'
```

**Screenshots**

Image

 - OS: Linux
 - Browser: Firefox
 - Version: 136

**Smartphone (please complete the following information):**
N/A

**Additional context**
N/A

sandhose:
This changes things so that we don't return on the first error during validation, but rather accumulate them and show a proper errored form a the end.

This is mainly to fix #4170, as we weren't evaluating the policy if the homeserver told us the username isn't available.

This is how it looks like if you choose invalid chars in the username after this patch:

image

### DIFF
--- a/crates/handlers/src/upstream_oauth2/link.rs
+++ b/crates/handlers/src/upstream_oauth2/link.rs
@@ -722,7 +722,7 @@ pub(crate) async fn post(
                     homeserver_denied_username = true;
                 }
 
-                // If we need have a TOS in the config, make sure the user has accepted it
+                // If we have a TOS in the config, make sure the user has accepted it
                 if site_config.tos_uri.is_some() && !accept_terms {
                     form_state.add_error_on_field(
                         mas_templates::UpstreamRegisterFormField::AcceptTerms,
--- a/crates/handlers/src/upstream_oauth2/link.rs
+++ b/crates/handlers/src/upstream_oauth2/link.rs
@@ -672,7 +672,7 @@ pub(crate) async fn post(
                 ctx
             };
 
-            let forced_username = if provider.claims_imports.localpart.is_forced() {
+            let username = if provider.claims_imports.localpart.is_forced() {
                 let template = provider
                     .claims_imports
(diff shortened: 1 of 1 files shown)


crates/handlers/src/upstream_oauth2/link.rs
```diff
@@ -672,136 +672,116 @@ pub(crate) async fn post(
...
+                    // error from the policy, to avoid showing both
+                    homeserver_denied_username = true;
+                }
 
-            // Ask the homeserver to make sure the username is valid
-            let is_available = homeserver
-                .is_localpart_available(&username)
-                .await
-                .map_err(RouteError::HomeserverConnection)?;
+                // If we need have a TOS in the config, make sure the user has accepted it
```
reivilibre, line 725:
```suggestion
                // If we have a TOS in the config, make sure the user has accepted it
```

reivilibre (approved):

Issue #2089, opened 2024-09-10
Make it possible to auto-import existing users when the localpart matches in upstream OAuth 2.0 logins


matrixbot:
*This issue was originally created by [**@sandhose**](https://github.com/sandhose) at <https://github.com/matrix-org/matrix-authentication-service/issues/2089>.*

When there is a new upstream OAuth 2.0 link and the localpart imported matches an existing user, we have 4 choices:

 - do nothing, and fail the registration/login (corresponds to `allow_existing_users: false` in Synapse)
 - import the link *if and only if* there is no existing link for this user
 - import the link and *replace* any existing link for this user
 - import the link without overriding existing links for this user (corresponds to `allow_existing_users: true` in Synapse)

matrixbot:
*This comment was originally posted by [**@matrixbot**](https://github.com/matrixbot) at <https://github.com/matrix-org/matrix-authentication-service/issues/2089#issuecomment-2341164207>.*


For your information, this issue has been copied over to the Element fork of `matrix-authentication-service`: <https://github.com/element-hq/matrix-authentication-service/issues/2089>

mcalinghee:
@sandhose : I'm trying to tackle this issue with #4193, please let me know your view on it

scottwallacesh:
Where does `syn2mas` figure into this?

PR #4193, opened 2025-03-11
Allow existing user

mcalinghee:


odelcroi:
unit test added

crates/handlers/src/upstream_oauth2/link.rs
```diff
@@ -837,10 +840,22 @@ pub(crate) async fn post(
...
-            let user = repo.user().add(&mut rng, &clock, username).await?;
+            let user = if provider.allow_existing_users {
+                // If the provider allows existing users, we can use the existing user
+                let existing_user = repo.user().find_by_username(&username).await?;
+                if existing_user.is_some() {
+                    existing_user.unwrap()
+                } else {
+                    REGISTRATION_COUNTER
+                        .add(1, &[KeyValue::new(PROVIDER, provider.id.to_string())]);
+                    // This case should not happen
```
odelcroi, line 851: why this comment?

odelcroi:
## Wrong configuration of provider
When activating the option `allow_existing_users : true`, the claim `username` is used to check if users exist or not. If the attribute mapper `username` is set to `ignore` or `suggest `, the user existence is not checked.  Therefore to enable this option the `username` attribute mapper should be set to `required` or `force`.
- documentation should mention it
- an error should be raised at startup if the configuration does not respect this condition

## Impersonation risk

The documentation should also mention the risk of impersonation when activating this option. The attribute mapper `username` from each identity providers must guarantee that user accounts will be not mingled. 
Example : 

```yaml
# Provider A
      claims_imports:
        localpart:
          action: require 
          template: "{{ user.preferred_username }}"
```

Alice has an existing account with `username` "alice" . When using provider A with `preferred_username` as "alice",  her existing account gets linked to the external account. 

If users are free to choose their `preferred_username` in the `provider A` then they can impersonate any user that has not be linked already.

odelcroi:
## Impersonation risk
Details here : https://github.com/element-hq/synapse/issues/17837

The documentation should also mention the risk of impersonation when activating this option. The attribute mapper `username` from each identity providers must guarantee that user accounts will be not mingled. 
Example : 

```yaml
# Provider A
      claims_imports:
        localpart:
          action: require 
          template: "{{ user.preferred_username }}"
```

Alice has an existing account with `username` "alice" . When using provider A with `preferred_username` as "alice",  her existing account gets linked to the external account. 

If users are free to choose their `preferred_username` in the `provider A` then they can impersonate any user that has not be linked already.

- documentation should mention it

crates/handlers/src/upstream_oauth2/link.rs
mcalinghee: correct

odelcroi, line 851:
```suggestion
```

docs/reference/configuration.md
```diff
@@ -770,6 +770,9 @@ upstream_oauth2:
         # This helps end user identify what account they are using
         account_name:
           #template: "@{{ user.preferred_username }}"
+      # set to true to allow a user logging in via OIDC to match a pre-existing account instead of failing.
```
odelcroi, line 773:
```suggestion
      # set to true to allow a user logging in via OIDC to match a pre-existing account instead of failing. 
      # Account matching is based on the localpart template which needs to be set to require or force.
```

odelcroi (approved):

templates/pages/upstream_oauth2/do_register.html
```diff
@@ -11,7 +11,24 @@
...
+      </div>
+
+      <div class="header">
+        <h1 class="title">
+          {# todo translate #} 
+          Link your existing account
+        </h1>
+        <p class="text">
+          {# todo translate #} 
+          We have detected you have an existing account, confirm you want to use this account
```
odelcroi, line 27: needs translation

odelcroi (approved):

crates/config/src/sections/upstream_oauth2.rs
```diff
@@ -571,6 +583,13 @@ pub struct Provider {
     #[serde(default, skip_serializing_if = "ClaimsImports::is_default")]
     pub claims_imports: ClaimsImports,
 
+    /// Whether to allow a user logging in via OIDC to match a pre-existing
+    /// account instead of failing. This could be used if switching from
+    /// password logins to OIDC.
+    //Defaults to false.
+    #[serde(default)]
+    pub allow_existing_users: bool,
```
sandhose: In line with #2089, I would like the configuration flag to not just be a boolean, but rather an enum with:

 - `fail` (the default) which fails the registration on conflict
 - `add`/`append` which adds the link, regardless of whether there is an existing link or not
 - `replace`/`overwrite` which overwrites any existing link
 - `set`/`add-if-missing`/`set-new` which adds a link *only* if there are no existing link for that provider on the user

Not 100% sure about the naming, and I'm fine with only having the `fail` and `add` options for now

I think it also makes sense to put that in the claims_imports.localpart section; as this is stored as a JSON blob in the database, it even avoids the database migration for this

```yaml
claims_imports:
  localpart:
    on_conflict: add
```

What do you think?

crates/handlers/src/upstream_oauth2/link.rs
```diff
@@ -473,7 +473,9 @@ pub(crate) async fn get(
                             .await
                             .map_err(RouteError::HomeserverConnection)?;
 
-                        if maybe_existing_user.is_some() || !is_available {
+                        if !provider.allow_existing_users
```
sandhose: This logic is a bit hard to follow. In case we link to an existing user, I don't think we'd want to import the email address (we might end up with duplicate email addresses to deal with) or display name, *just* link the user.

This means I would rather branch out here with something like:

```rust
if let Some(existing_user) {
    if provider.allow_existing_users {
        // Render the UI to link the user
        return …
    }

    // Show the error about upstream account not being linked to this account
}

if !is_available {
    // Show an error about the user not being available on Synapse (could be an AS namespace for example)
}
```

This also implies its own view, rather than reusing the do_register template

It also means we don't run the 'registration' policy for an account link, which I think makes sense

crates/handlers/src/upstream_oauth2/link.rs
```diff
@@ -507,7 +509,11 @@ pub(crate) async fn get(
                             })
                             .await?;
 
-                        if res.valid() {
+                        if provider.allow_existing_users && maybe_existing_user.is_some() {
+                            // use existing user if allowed
+                            ctx.with_existing_user(maybe_existing_user.unwrap())
+                                .with_localpart(localpart, true)
+                        } else if res.valid() {
```
sandhose: Proposal up there would remove all this complexity

crates/handlers/src/upstream_oauth2/link.rs
```diff
@@ -837,26 +844,39 @@ pub(crate) async fn post(
...
+            //search and use existing users if allowed
+            if provider.allow_existing_users {
+                existing_user = repo.user().find_by_username(&username).await?;
             }
 
-            // And schedule the job to provision it
-            let mut job = ProvisionUserJob::new(&user);
+            let user = if existing_user.is_some() {
+                existing_user.unwrap()
+            } else {
```
sandhose: Proposal up there would remove all this complexity (and the potential email conflict)

docs/setup/sso.md
```diff
@@ -65,6 +65,23 @@ The template has the following variables available:
  - `user`: an object which contains the claims from both the `id_token` and the `userinfo` endpoint
  - `extra_callback_parameters`: an object with the additional parameters the provider sent to the redirect URL
 
+
+## Allow linking existing user accounts
```
sandhose, line 70: Really nice to see this, including the security warning. Just putting a note to check back that this gets adapted once you've changed the option to the enum

templates/pages/upstream_oauth2/do_register.html
```diff
@@ -11,7 +11,24 @@
...
+  {% if existing_user %}
+    <header class="page-heading">
+      <div class="icon">
+        {{ icon.download() }}
+      </div>
+
+      <div class="header">
+        <h1 class="title">
+          {# todo translate #} 
+          Link your existing account
```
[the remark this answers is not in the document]
sandhose: A very much valid comment. How you'd do that:

 - use the key you expect like `{{ _("mas.upstream_oauth2.do_register.headline", variable=foo) }}`
 - run `mas-i18n-scan` to update the translations/en.json; this is called in the `misc/update.sh` script

crates/config/src/sections/upstream_oauth2.rs
odelcroi: I like the idea of letting the field open to more business logic using an enum, what do you think of this naming : 
```yaml
upstream_oauth2:
  providers:
    - id: 
      claims_imports:
        localpart:
          on_conflict: add
```
`fail` (the default) which fails the sso login on conflict
`add` which adds the oauth identity link, regardless of whether there is an existing link or not
`replace` which overwrites any existing oauth identity link
`set` which adds a oauth identity link only if there are no existing link for that provider on the user

crates/handlers/src/upstream_oauth2/link.rs
odelcroi: > I don't think we'd want to import the email address (we might end up with duplicate email addresses to deal with) or display name, just link the user.

You're right — on first SSO login, we likely don't want to import or update the user's email address or display name, to avoid duplicates issue at this time.

However, we do have a use case with the French administration where the external SSO provider is considered the source of truth for user info. In that scenario, we want to update the user's email and display name on sso login, since those details can change over time (e.g., name changes or updated email addresses).

This was handled in Synapse — I’ll open an issue to track it here as well.

odelcroi: > This also implies its own view, rather than reusing the do_register template

I guess `suggest_link.html` does the job

odelcroi: >It also means we don't run the 'registration' policy for an account link, which I think makes sense

Yes indeed :)

translations/fr.json
sandhose, line 1: Translations for other languages should be done through Localazy after it being merged

translations/en.json
```diff
@@ -693,15 +693,29 @@
           "description": "Page shown when the user tries to link an upstream account that is already linked to another account"
         }
       },
+      "login_link": {
+        "action": "Continue",
+        "@action": {
+          "context": "pages/upstream_oauth2/login_link.html:29:28-70"
+        },
+        "description": "An account exists for the username : %(username)s, it will be linked to this new provider",
```
sandhose:
```suggestion
        "description": "An account exists for this username (%(username)s), it will be linked to this upstream account.",
```

templates/pages/upstream_oauth2/login_link.html
```diff
@@ -0,0 +1,33 @@
...
+      <h1 class="title">{{ _("mas.upstream_oauth2.login_link.heading") }}</h1>
+    </div>
+  </header>
+  <section class="flex flex-col gap-6 justify-center">
+    {{ _("mas.upstream_oauth2.login_link.description", username=linked_user.username) }}
+   
+    <form method="POST" class="cpd-form-root">
+      <input type="hidden" name="csrf" value="{{ csrf_token }}" />
+      <input type="hidden" name="action" value="associate" />
+      <input type="hidden" name="username" value="{{ linked_user.username }}" />
```
sandhose: Why do we need those hidden inputs? Shouldn't we re-run the templates to make sure the user doesn't temper with this?

templates/pages/upstream_oauth2/login_link.html
```diff
@@ -0,0 +1,33 @@
+{#
+Copyright 2024 New Vector Ltd.
+Copyright 2022-2024 The Matrix.org Foundation C.I.C.
```
sandhose:
```suggestion
Copyright 2025 New Vector Ltd.
```

frontend/schema.graphql
sandhose, line 1: This looks like it fails to generate

docs/reference/configuration.md
```diff
@@ -754,6 +754,12 @@ upstream_oauth2:
         localpart:
           #action: force
           #template: "{{ user.preferred_username }}"
+          
+          # How to handle when localpart already exists.
+          # Possible values are (default: fail):
+          # - `add` : Adds the oauth identity link to the existing user, regardless of whether there is an existing link or not.
+          # - `fail` : Fails the sso login.
```
sandhose:
```suggestion
          # - `add` : Adds the upstream account link to the existing user, regardless of whether there is an existing link or not.
          # - `fail` : Fails the upstream OAuth 2.0 login.
```

crates/templates/src/lib.rs
```diff
@@ -401,6 +401,9 @@ register_templates! {
     /// Render the upstream link mismatch message
     pub fn render_upstream_oauth2_link_mismatch(WithLanguage<WithCsrf<WithSession<UpstreamExistingLinkContext>>>) { "pages/upstream_oauth2/link_mismatch.html" }
 
+    /// Render the upstream link match
+    pub fn render_upstream_oauth2_login_link(WithLanguage<WithCsrf<UpstreamExistingLinkContext>>) { "pages/upstream_oauth2/login_link.html" }
```
sandhose, line 405: You should add this to the check down there as well, so that it gets run by the test suite

crates/data-model/src/upstream_oauth2/provider.rs
```diff
@@ -348,3 +351,22 @@ impl ImportAction {
         }
     }
 }
+
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize, Default)]
+#[serde(rename_all = "lowercase")]
+pub enum OnConflict {
+    /// Fails the sso login on claim conflict
```
sandhose:
```suggestion
    /// Fails the upstream OAuth 2.0 login
```

crates/data-model/src/upstream_oauth2/provider.rs
```diff
@@ -348,3 +351,22 @@ impl ImportAction {
...
+    #[default]
+    Fail,
+
+    /// Adds the oauth identity link, regardless of whether there is an existing
+    /// link or not
```
sandhose:
```suggestion
    /// Adds the upstream account link, regardless of whether there is an existing
    /// link or not
```

crates/config/src/sections/upstream_oauth2.rs
```diff
@@ -211,6 +245,10 @@ pub struct LocalpartImportPreference {
     /// If not provided, the default template is `{{ user.preferred_username }}`
     #[serde(default, skip_serializing_if = "Option::is_none")]
     pub template: Option<String>,
+
+    /// How to handle conflicts on the claim, default value is `Fail``
```
sandhose:
```suggestion
    /// How to handle conflicts on the claim, defaults to `fail`
```

crates/config/src/sections/upstream_oauth2.rs
```diff
@@ -211,6 +245,10 @@ pub struct LocalpartImportPreference {
...
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub on_conflict: Option<OnConflict>,
```
sandhose:
```suggestion
    #[serde(default, skip_serializing_if = "OnConflict::is_default")]
    pub on_conflict: OnConflict,
```

and define a 'is_default' method on OnConflict

crates/config/src/sections/upstream_oauth2.rs
```diff
@@ -183,6 +198,25 @@ impl ImportAction {
...
+    #[default]
+    Fail,
+
+    /// Adds the oauth identity link, regardless of whether there is an existing
+    /// link or not
+    Add,
+}
+
+impl OnConflict {
+    const fn add(&self) -> bool {
```
sandhose: Convention for this kind of method is `is_*`, and in this case, once we add other options (set/override), it makes more sense to use `is_fail`?

crates/cli/src/sync.rs
```diff
@@ -47,14 +61,17 @@ fn map_claims_imports(
         localpart: mas_data_model::UpstreamOAuthProviderImportPreference {
             action: map_import_action(config.localpart.action),
             template: config.localpart.template.clone(),
+            on_conflict: map_import_on_conflict(config.localpart.on_conflict),
         },
         displayname: mas_data_model::UpstreamOAuthProviderImportPreference {
             action: map_import_action(config.displayname.action),
             template: config.displayname.template.clone(),
+            on_conflict: None,
```
sandhose: Ideally the 'localpart' import preference would have its own type to avoid this, but err, I'm fine with keeping it like that since it's internal anyway

crates/handlers/src/upstream_oauth2/link.rs
```diff
@@ -473,11 +478,26 @@ pub(crate) async fn get(
...
-                                // linked to this upstream user.
-                                warn!(username = %localpart, user_id = %existing_user.id, "Localpart template returned an existing username");
+                        if let Some(existing_user) = maybe_existing_user {
+                            // The mapper returned a username which already exists, but isn't
+                            // linked to this upstream user.
+                            let on_conflict = provider
+                                .claims_imports
+                                .localpart
+                                .on_conflict
+                                .unwrap_or(DEFAULT_ON_CONFLICT);
```
sandhose:
```suggestion
                                .unwrap_or_default();
```

crates/handlers/src/upstream_oauth2/link.rs
```diff
@@ -473,11 +478,26 @@ pub(crate) async fn get(
...
+
+                            if on_conflict.is_add() {
```
sandhose: Can you make this a `match` instead? This way once we add more options, we're sure to add proper handling here

crates/handlers/src/upstream_oauth2/link.rs
```diff
@@ -618,6 +653,27 @@ pub(crate) async fn post(
             session
         }
 
+        (None, None, FormData::Associate { username }) => {
+            //user already exists, but it is not linked, neither connected
+            //proceed by associating the link and connect the user
+
+            let maybe_user = repo.user().find_by_username(&username.unwrap()).await?;
```
sandhose: We should not trust the user input like that! This should be pulled by rendering the localpart again, and we should check again that the provider allows 'adding' links here

sandhose (changes requested):

crates/config/src/sections/upstream_oauth2.rs
odelcroi: Agreed

crates/handlers/src/upstream_oauth2/link.rs
odelcroi: > We should not trust the user input like that! This should be pulled by rendering the localpart again,

yes, 100% agreed, it was planned :) Sorry should have switch the PR to Draft for the refactor

> and we should check again that the provider allows 'adding' links here

Indeed a malicious user can avoid the previous protection

mcalinghee:
@sandhose : are you ok for this [renaming](https://github.com/element-hq/matrix-authentication-service/pull/4193/commits/da0743a177566ab1f2318d305b3d6fdd00df5c7e)?

sandhose:
> @sandhose : are you ok for this [renaming](https://github.com/element-hq/matrix-authentication-service/pull/4193/commits/da0743a177566ab1f2318d305b3d6fdd00df5c7e)?

Yep, makes sense!

crates/config/src/sections/upstream_oauth2.rs
```diff
@@ -110,6 +110,17 @@ impl ConfigurationSection for UpstreamOAuth2Config {
                     }
                 }
             }
+
+            if !provider.claims_imports.localpart.on_conflict.is_default()
+                && !matches!(
```
sandhose: Minor code style, but I don't think we should rely on `is_default` but rather match

```suggestion
            if matches!(
                provider.claims_imports.localpart.on_conflict,
                OnConflict::Add
            ) && !matches!(
```

templates/pages/upstream_oauth2/login_link.html
```diff
@@ -0,0 +1,31 @@
+{#
+Copyright 2025 New Vector Ltd.
+
+SPDX-License-Identifier: AGPL-3.0-only
```
sandhose: The license header changed in the meantime :)

```suggestion
SPDX-License-Identifier: AGPL-3.0-only OR LicenseRef-Element-Commercial
```

crates/handlers/src/upstream_oauth2/link.rs
```diff
@@ -618,6 +652,79 @@ pub(crate) async fn post(
...
+            let localpart = render_attribute_template(&env, template, &context, true)?;
+
+            let maybe_user = repo.user().find_by_username(&localpart.unwrap()).await?;
+
+            if maybe_user.is_none() {
+                //user can not be None at this stage
+                return Err(RouteError::InvalidFormAction);
+            }
+
+            let user = maybe_user.unwrap();
```
sandhose:
```suggestion
            let Some(user) = maybe_user else {
                // user cannot be None at this stage
                return Err(RouteError::InvalidFormAction);
            }
```

crates/handlers/src/upstream_oauth2/link.rs
```diff
@@ -618,6 +652,79 @@ pub(crate) async fn post(
...
+            }
+            if let Some(extra_callback_parameters) = upstream_session.extra_callback_parameters() {
+                context = context.with_extra_callback_parameters(extra_callback_parameters.clone());
+            }
+            if let Some(userinfo) = upstream_session.userinfo() {
+                context = context.with_userinfo_claims(userinfo.clone());
+            }
+            let context = context.build();
+
+            // `is_forced` checks both if is it `force` or `require`
```
sandhose:
```suggestion
```

crates/handlers/src/upstream_oauth2/link.rs
```diff
@@ -473,19 +477,49 @@ pub(crate) async fn get(
...
+                                    // new oauth link is allowed
+                                    let ctx = UpstreamExistingLinkContext::new(existing_user)
+                                        .with_csrf(csrf_token.form_value())
+                                        .with_language(locale);
+
+                                    return Ok((
+                                        cookie_jar,
+                                        Html(templates.render_upstream_oauth2_login_link(&ctx)?)
+                                            .into_response(),
+                                    ));
```
sandhose, line 512: I'm now wondering if it makes sense to show this intermediate page or not? For the 'active session' linking I think it does, as it's user-initiated and we don't want it to happen automatically, whereas here the intent really is for HS admins to transition from local passwords to upstream IDPs, the users should not care about this.

I'm fine with leaving it like that for now, and I'll clean that up later, as we also would like to make import dialog to be optional in some cases

sandhose (commented):
We're almost there! A few nitpicks, and I also wonder whether the confirmation screen is needed in this case or not; we can also change that later

odelcroi (approved):

crates/handlers/src/upstream_oauth2/link.rs
```diff
@@ -1099,4 +1212,357 @@ mod tests {
...
+        let existing_username = format!("{}{}", "john", timestamp);
+        let existing_email = format!("{}@{}", existing_username, "example.com");
+
+        //existing username matches oidc username
+        let oidc_username = existing_username.clone();
+
+        //oidc email is different from existing email
+        let oidc_email: String = format!("{}{}@{}", "any_email", timestamp, "example.com");
+
+        let subject = format!("{}+{}", "subject", timestamp);
```
sandhose: The databases are isolated between each test, no need to add timestamps there!

crates/handlers/src/upstream_oauth2/link.rs
```diff
@@ -618,6 +652,76 @@ pub(crate) async fn post(
             session
         }
 
+        (None, None, FormData::Link) => {
+            // User already exists, but it is not linked, neither logged in
+            // Proceed by associating the link and log in the user
+            // Upstream_session is used to re-render the username as it is the only source
+            // of truth
```
sandhose: Just a little reflow and rewording of this comment
```suggestion
            // There is an existing user with the same username, but no link.
            // If the configuration allows it, the user is prompted to link the
            // existing account. Note that we cannot trust the user input here,
            // which is why we have to re-calculate the localpart, instead of
            // passing it through form data.
```

crates/handlers/src/upstream_oauth2/link.rs
```diff
@@ -618,6 +652,76 @@ pub(crate) async fn post(
...
+            let template = provider
+                .claims_imports
+                .localpart
+                .template
+                .as_deref()
+                .unwrap_or(DEFAULT_LOCALPART_TEMPLATE);
+
+            let localpart = render_attribute_template(&env, template, &context, true)?;
+
+            let maybe_user = repo.user().find_by_username(&localpart.unwrap()).await?;
```
sandhose: Better error out than crash (even if we recover from panics!)

```suggestion
            let Some(localpart) = render_attribute_template(&env, template, &context, true)? else {
                // This should never be the case at this point
                return Err(RouteError::InvalidFormAction);
            };

            let maybe_user = repo.user().find_by_username(&localpart).await?;
```

crates/handlers/src/upstream_oauth2/link.rs
```diff
@@ -1099,4 +1212,357 @@ mod tests {
...
+            .unwrap();
+        let cookie_jar = upstream_sessions.save(cookie_jar, &state.clock);
+        cookies.import(cookie_jar);
+
+        let user = create_user(
+            &mut rng,
+            &state.clock,
+            &mut repo,
+            existing_username.clone(),
+            existing_email.clone(),
```
sandhose: Why does this test need an email? I don't think the `create_user` utility is very helpful here

crates/data-model/src/upstream_oauth2/provider.rs
```diff
@@ -391,3 +411,22 @@ impl ImportAction {
...
+    /// existing link or not
+    Add,
+}
+
+impl OnConflict {
+    #[must_use]
+    pub fn is_add(&self) -> bool {
+        matches!(self, Self::Add)
+    }
+}
```
sandhose: I don't see it used?

```suggestion
```

crates/handlers/src/upstream_oauth2/link.rs
odelcroi: I wasn't sure, that's why I added those. Do the unit tests use the postgresql db or it is only in cache? I see the DATABASE_URL is mandatory for running the tests

sandhose: They use a real PostgreSQL database, but isolated, one for each test. It will create temporary databases names `_sqlx_test_XXX` on the `DATABASE_URL` you've provided, and destroy it at the end of the test

crates/handlers/src/upstream_oauth2/link.rs
odelcroi: We have some internal PR that tests matching by email. Let's clean those tests with optional artifacts

PR #4198, opened 2025-03-12
Better feedback on failure cases during login

### LINKED ISSUE #3728: Returning federated user prompted for password that does not exist
**Describe the bug**
When a user has signed onto the device previously, an attempt to sign in again is presented with a welcome back screen bearing the user's name and a password prompt.  However, if the was signed in via an OIDC provider (e.g. Google/Apple), they have no password and this attempt to login will fail with a "No password" error.  The form also does not offer them the choice to login with their previous OIDC provider at all.

It seems like this welcome-back behavior needs to be OIDC aware to be effective.

**To Reproduce**
1. Sign into a client using MAS and an OIDC provider (Google).
2. Complete login, then logout of the app.
3. Attempt to login again, you will be presented to return and reauthorize the account with a password.

**Expected behavior**
I would expect that a returning user would be prompted to use their previous method of 3rd party authorization if they do not have a password.

**Smartphone:**
 - Device: iOS
 - OS: 17.5
 - Browser: Authenticated Embedded Browser
 - Version: NA

sandhose:
This includes:

 - displaying the list of upstream providers on failed password logins
 - showing a proper 'account deactivated' or 'account locked' page
 - show the generic error page on 500-like errors

Fixes #3728

### DIFF
--- a/crates/handlers/src/views/login.rs
+++ b/crates/handlers/src/views/login.rs
@@ -15,9 +15,9 @@ use hyper::StatusCode;
 use mas_axum_utils::{
     FancyError, SessionInfoExt,
     cookies::CookieJar,
-    csrf::{CsrfExt, CsrfToken, ProtectedForm},
+    csrf::{CsrfExt, ProtectedForm},
 };
-use mas_data_model::{BrowserSession, UserAgent, oauth2::LoginHint};
+use mas_data_model::{UserAgent, oauth2::LoginHint};
 use mas_i18n::DataLocale;
 use mas_matrix::HomeserverConnection;
 use mas_router::{UpstreamOAuth2Authorize, UrlBuilder};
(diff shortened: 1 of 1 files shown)


crates/handlers/src/views/login.rs
```diff
@@ -146,181 +144,190 @@ pub(crate) async fn post(
...
+        let ctx = AccountInactiveContext::new(user)
+            .with_csrf(csrf_token.form_value())
+            .with_language(locale);
+        let content = templates.render_account_locked(&ctx)?;
+        return Ok((cookie_jar, Html(content)).into_response());
+    }
+
+    // At this point, we should have a 'valid' user. In case we missed something, we
+    // want it to crash in tests/debug builds
+    debug_assert!(user.is_valid());
```
reivilibre, line 297: tbh, I'd be in favour of making this a full-blown assert.

I'd only use `debug_assert` for perf reasons, which doesn't seem to apply here.

reivilibre (approved):

crates/handlers/src/views/login.rs
sandhose: I get that, but I also kinda want tasks to not panic as much as possible? Alternative is to bubble up a proper Error but 🤷

PR #4208, opened 2025-03-12
syn2mas: disable logging of slow statements, better access token query perf

reivilibre:
Follows: #4175

This is more cherry-picking from the tip branch. As far as I can tell, this is now everything except HACKs and the progress bar :).


sandhose (approved):
Can you rebase the tip branch after that? So that we have a build ready for m.org

reivilibre:
sure :)

PR #4215, opened 2025-03-12
syn2mas: Add progress reporting to log and to opentelemetry metrics

reivilibre:
The logging looks a bit like:

```
2025-03-12T18:40:05.936000Z  INFO mas_cli::commands::syn2mas: crates/cli/src/commands/syn2mas.rs:303: migrating users: 29/~3798454 (~0.0%)
2025-03-12T18:40:35.938238Z  INFO mas_cli::commands::syn2mas: crates/cli/src/commands/syn2mas.rs:303: migrating users: 59/~3798454 (~0.0%)
2025-03-12T18:41:05.939241Z  INFO mas_cli::commands::syn2mas: crates/cli/src/commands/syn2mas.rs:303: migrating users: 89/~3798454 (~0.0%)
```

(artificially slowed down)

I did not try the metrics. The HTTP listeners aren't active so Prometheus metrics are no good, you'd have to push them to an OpenTelemetry thing and I didn't bother to do that yet.
That said, I'm not convinced by the  value so perhaps it's better to just drop that aspect?

(edit maybe it'd be good to include the word 'progress' in the line there somewhere. Perhaps `target: "syn2mas::progress"` or something.... not sure what's best. Happy to hear your thoughts on that)

### DIFF
--- a/crates/cli/src/commands/syn2mas.rs
+++ b/crates/cli/src/commands/syn2mas.rs
@@ -10,7 +10,6 @@ use mas_config::{
 };
 use mas_storage::SystemClock;
 use mas_storage_pg::MIGRATOR;
-use opentelemetry::KeyValue;
 use rand::thread_rng;
 use sqlx::{Connection, Either, PgConnection, postgres::PgConnectOptions, types::Uuid};
 use syn2mas::{
@@ -18,10 +17,7 @@ use syn2mas::{
 };
 use tracing::{Instrument, error, info, info_span, warn};
 
-use crate::{
-    telemetry::METER,
-    util::{DatabaseConnectOptions, database_connection_from_config_with_options},
-};
+use crate::util::{DatabaseConnectOptions, database_connection_from_config_with_options};
 
 /// The exit code used by `syn2mas check` and `syn2mas migrate` when there are
 /// errors preventing migration.
@@ -258,7 +254,6 @@ impl Options {
 
                 let occasional_progress_logger_task =
                     tokio::spawn(occasional_progress_logger(progress.clone()));
-                let progress_telemetry_task = tokio::spawn(progress_telemetry(progress.clone()));
 
                 let mas_matrix = MatrixConfig::extract(figment)?;
                 eprintln!("\n\n");
@@ -274,7 +269,6 @@ impl Options {
                 .await?;
 
                 occasional_progress_logger_task.abort();
-                progress_telemetry_task.abort();
 
                 Ok(ExitCode::SUCCESS)
             }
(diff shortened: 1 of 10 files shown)


crates/cli/src/commands/syn2mas.rs
```diff
@@ -258,11 +269,75 @@ impl Options {
...
+/// Reports migration progress as OpenTelemetry metrics
+async fn progress_telemetry(progress: Progress) {
+    let migrated_data_counter = METER
+        .u64_gauge("migrated_data")
+        .with_description("How many entities have been migrated so far")
+        .build();
+    let max_data_counter = METER
+        .u64_gauge("max_data")
+        .with_description("How many entities of the given type exist (approximate)")
+        .build();
```
sandhose, line 325: Upping counters directly is probably performant enough? I would make those counters, static with a `std::sync::LazyLock` and just record every time we insert

crates/cli/src/commands/syn2mas.rs
```diff
@@ -258,11 +269,75 @@ impl Options {
                     &clock,
                     &mut rng,
                     provider_id_mappings,
+                    &progress,
                 )
                 .await?;
 
+                occasional_progress_logger_task.abort();
+                progress_telemetry_task.abort();
```
sandhose, line 277: Ideally this would be done through a CancellationToken, I don't particularly like aborting tasks, but since we haven't set that up in the rest of the migration process, I'm happy to keep it like this for now

crates/cli/src/commands/syn2mas.rs
```diff
@@ -258,11 +269,75 @@ impl Options {
...
+                info!(name: "progress", "still waiting for rebuild of constraint {constraint_name}");
+            }
+        }
+    }
+}
+
+/// Reports migration progress as OpenTelemetry metrics
+async fn progress_telemetry(progress: Progress) {
+    let migrated_data_counter = METER
+        .u64_gauge("migrated_data")
```
sandhose, line 319: Let's use a prefix, plus a common namespace, e.g. `syn2mas.data.migrated` and `syn2mas.data.total`. Note that when translated to Prometheus, dots will be replaced by underscores. Feel free to come up something else, this is just a suggestion

sandhose (changes requested):
Generally happy with the approach for the progress logs, I would just do the OTEL meters differently

crates/cli/src/commands/syn2mas.rs
reivilibre: `syn2mas.entity`... maybe. Don't like how vague these words are but something like 'migratee' is no better haha. I guess it doesn't exactly matter as long as it makes some sense.

crates/cli/src/commands/syn2mas.rs
reivilibre: I didn't make them static but maybe this is OK?

sandhose (approved):

PR #4234, opened 2025-03-17
Passkeys (experimental)

### LINKED ISSUE #4582: Support passkeys

Logging in (authentication).

Plug in your passkey. Press button. Done.

Logging in via username + password is user unfriendly, error prone and unsecure.

tonkku107:
Adds support for passkeys behind an experimental config option.

Includes:
- Adding, removing and renaming passkeys in the account management frontend
- Signing in with a passkey

Does not include (will probably have to be separate PRs later):
- Ability to remove the password to go completely passwordless
- Registering with a passkey instead of a password
- Creating a new passkey during recovery instead of a password

Can be reviewed commit per commit.
Frontend is hacked together from existing components pending designs.

Closes #4582


tonkku107:
I was expecting `webauthn-rs`'s dependence on openssl (3) to be a problem, but I wasn't expecting there to be a CI test for it 😁. Maybe `webauthn_rp` is a better alternative in this case.

When I'm running the frontend tests locally, they're only rendering error pages and I don't know why. Would appreciate some pointers as to what I'm missing.

Either way, in the meanwhile @sandhose could we request for the designs?

tonkku107:
> may still be small enough that you can avoid storing them in the database entirely

Storing in memory won't be an option since multiple instances can be run at once.

> The challenge they contain is based on a randomly-generated u128 which means they have the same amount of "entropy" as the 16-byte ID you plan on using.

I've added an Ulid to be consistent with the rest of the stuff in this project. There's actually only 10 bytes of entropy within the ID as Ulids reserve 6 bytes for a timestamp.

> you may be able to write something like `(&cred_id).into()`

Yeah I know and I wasn't a big fan of that and ultimately moved it out to be more explicit about the weird type stuff going on.

> but I'm guessing you know this and you're intentionally storing the JSON array of strings instead

I used JSON or strings wherever I could, but if database efficiency is a concern to the maintainers that could be changed.

> A few examples that will fail origin validation

Can you even get any of those examples out of a browser? 😄
The point there is to just have more lax verification for development purposes anyway. The real concerns about too strict or not strict enough would come from the case where the host of `public_base` is fed into `DomainOrigin::new`. Imo, it's just right for most deployments but some additional config options might be necessary for some people's needs.

tonkku107:
> Well that is not true since you do use `CredentialId::encode` and `UserHandle::encode`. 

I use `serde_json::to_string` and `serde_json::from_str` for CredentialIds. User handles I constructed from the Ulid's bytes.

> Additionally, you don't seem to care about `MetadataOwned::decode`; thus you could also call `Metadata::into_json` instead of `Metadata::encode`.

Did not notice that as I was looking for the serialization impls. The metadata is quite useless anyway.

crates/data-model/src/users.rs
```diff
@@ -215,3 +216,26 @@ pub struct UserRegistration {
     pub created_at: DateTime<Utc>,
     pub completed_at: Option<DateTime<Utc>>,
 }
+
+#[derive(Debug, Clone, PartialEq, Eq, Serialize)]
+pub struct UserPasskey {
+    pub id: Ulid,
+    pub user_id: Ulid,
+    pub credential_id: String,
```
sandhose: This is supposed to be a 'buffersource', so raw bytes? Shouldn't that be a `Vec<u8>` / `BYTEA`?

crates/data-model/src/users.rs
```diff
@@ -215,3 +216,26 @@ pub struct UserRegistration {
...
+    pub name: String,
+    pub transports: serde_json::Value,
```
sandhose: I think I would rather have that the exact type, even if that means making mas-data-model depend on webauthn-rp? This way, potential errors comes at the repository level, and there is no need for post-processing those later

crates/storage-pg/migrations/20250303142622_passkeys.sql
sandhose, line 1: Learned this the hard way: foreign keys don't create indexes!

So for each FK, please also do an explicit index, e.g.

```sql
CREATE INDEX user_passkeys_user_id_fk ON user_passkeys (user_id);
```

It's fine to create indexes non-concurrently on tables you've just created, but for FKs you're creating on existing tables, you need to create those with `CREATE INDEX CONCURRENTLY` to avoid locking the table during migration.

The other caveat, is that `CREATE INDEX CONCURRENTLY` *must* run outside of a transaction, and by default, migrations are run in transactions. You can opt-out of this by having `-- no-transaction` at the top of the file… which also means you can have a single statement in the migration.

See https://github.com/element-hq/matrix-authentication-service/pull/4385 for example migrations for FK indexes on existing tables

crates/handlers/Cargo.toml
```diff
@@ -88,6 +88,7 @@ rand.workspace = true
 rand_chacha.workspace = true
 headers.workspace = true
 ulid.workspace = true
+webauthn_rp = { version = "0.3.0", features = ["bin", "serde_relaxed", "custom", "serializable_server_state"] }
```
sandhose: Ideally can you put new dependencies at the workspace level?

crates/handlers/src/webauthn.rs
```diff
@@ -0,0 +1,248 @@
...
+        clock: &impl Clock,
+        conn: &impl HomeserverConnection,
+        user: &User,
+        browser_session: &BrowserSession,
+    ) -> Result<(String, UserPasskeyChallenge)> {
+        // Get display name or default to username
+        let matrix_user = conn.query_user(&conn.mxid(&user.username)).await?;
+        let display_name = matrix_user
+            .displayname
+            .unwrap_or_else(|| user.username.clone());
```
sandhose: I think this should either default to None if there is no displayname, or to the MXID instead. We already use the username as the user entity name

crates/handlers/src/webauthn.rs
```diff
@@ -0,0 +1,248 @@
...
+
+        let exclude_credentials = repo
+            .user_passkey()
+            .all(user)
+            .await?
+            .into_iter()
+            .map(|v| {
+                Ok(PublicKeyCredentialDescriptor {
+                    id: serde_json::from_str(&v.credential_id)?,
+                    transports: serde_json::from_value(v.transports)?,
```
sandhose: Making those the right type at the data-model level would avoid making this weirdly fallible

crates/handlers/src/webauthn.rs
```diff
@@ -0,0 +1,248 @@
...
+            })
+            .collect::<Result<Vec<PublicKeyCredentialDescriptor<Vec<u8>>>>>()?;
+
+        let options = PublicKeyCredentialCreationOptions::passkey(
+            &self.rpid,
+            user_entity,
+            exclude_credentials,
+        );
+
+        let (server_state, client_state) = options.start_ceremony()?;
```
sandhose, line 203: I don't like that the underlying library is using its own time source, but it is what it is

crates/handlers/src/webauthn.rs
```diff
@@ -0,0 +1,248 @@
...
+
+        let user_passkey_challenge = repo
+            .user_passkey()
+            .add_challenge_for_session(rng, clock, server_state.encode()?, browser_session)
+            .await?;
+
+        Ok((
+            serde_json::to_string(&client_state)?,
```
sandhose, line 211: Can we serialise up one level? Or is this because the client_state borrows on a bunch of stuff?

crates/handlers/src/webauthn.rs
```diff
@@ -0,0 +1,248 @@
...
+
+        let rpid = Arc::new(RpId::Domain(host.try_into()?));
+
+        Ok(Self { rpid })
+    }
+
+    #[must_use]
+    pub fn get_allowed_origin(&self) -> DomainOrigin {
+        let host = (*self.rpid).as_ref();
+        if host == "localhost" {
```
sandhose: I assume this is important for development, can you comment why?

crates/handlers/src/webauthn.rs
```diff
@@ -0,0 +1,248 @@
...
+    ///
+    /// # Errors
+    /// If the `public_base` has no valid host domain
+    pub fn new(public_base: &Url) -> Result<Self> {
+        let host = public_base
+            .host_str()
+            .context("Public base doesn't have a host")?
+            .to_owned();
+
+        let rpid = Arc::new(RpId::Domain(host.try_into()?));
```
sandhose: Looking at how RP matching works, looks like it could make sense to have that configurable? e.g. matrix.org registering credentials on matrix.org instead of account.matrix.org? This is not very important if it's potentially too annoying to do

crates/handlers/src/webauthn.rs
```diff
@@ -0,0 +1,248 @@
...
+    ) -> Result<(String, UserPasskeyChallenge)> {
+        // Get display name or default to username
+        let matrix_user = conn.query_user(&conn.mxid(&user.username)).await?;
+        let display_name = matrix_user
+            .displayname
+            .unwrap_or_else(|| user.username.clone());
+
+        let user_entity = PublicKeyCredentialUserEntity {
+            name: user.username.as_str().try_into()?,
+            id: &UserHandle::decode(user.id.to_bytes())?,
```
sandhose, line 182: Yay that we have static-length user IDs, but also that means that it's going to be annoying if we ever want to do passkey registration at registration time, because we don't know the user ID until way later in the registration process… anyway, problem for later

crates/handlers/src/webauthn.rs
```diff
@@ -0,0 +1,248 @@
...
+
+        let user_id = Ulid::from_bytes(credential.user_id().encode()?);
+        if user_id != user.id {
+            return Err(WebauthnError::UserMismatch.into());
+        }
+
+        let cred_id = serde_json::to_string(&credential.id())?;
+
+        // Webauthn requires that credential IDs be unique globally
+        if repo.user_passkey().find(&cred_id).await?.is_some() {
```
sandhose, line 260: This felt quite weird to me that a attester-provided ID was supposed to be globally unique, but it looks like it usually have enough entropy to never really be a problem

crates/handlers/src/webauthn.rs
```diff
@@ -0,0 +1,414 @@
...
+        rng: &mut (dyn RngCore + Send),
+        clock: &impl Clock,
+        user: &User,
+        user_passkey_challenge: UserPasskeyChallenge,
+        response: String,
+        name: String,
+    ) -> Result<UserPasskey> {
+        let server_state = RegistrationServerState::decode(&user_passkey_challenge.state)?;
+
+        let response = serde_json::from_str::<RegistrationRelaxed>(&response)?.0;
```
sandhose:
```suggestion
        let RegistrationRelaxed(response) = serde_json::from_str(&response).context("Invalid PublicKeyCredential sent by client")?;
```

crates/handlers/src/webauthn.rs
```diff
@@ -0,0 +1,414 @@
...
+    /// [`WebauthnError::UserMismatch`] if the passkey is tied to a different
+    /// user.
+    ///
+    /// The rest of the anyhow errors should be treated as internal errors
+    pub async fn discover_credential(
+        &self,
+        repo: &mut impl RepositoryAccess,
+        response: String,
+    ) -> Result<(DiscoverableAuthentication16, User, UserPasskey)> {
+        let response = serde_json::from_str::<DiscoverableAuthenticationRelaxed16>(&response)?.0;
```
sandhose:
```suggestion
        let DiscoverableAuthenticationRelaxed16(response) = serde_json::from_str(&response)?;
```

crates/handlers/src/views/login/passkey/mod.rs
sandhose, line 1: Ideally this shouldn't be its own page for the form, but on the login page.

The "sign in with passkey" should trigger the challenge directly; which also means we can unconditionally create a challenge on the login page, even if this means we get a bunch of "ghost" challenges.

This will allow us to do [conditional mediation](https://www.corbado.com/blog/webauthn-conditional-ui-passkeys-autofill) on the username form on the login page

frontend/src/template_passkey.ts
sandhose, line 1: This could render a React component as well. This can feel overkill, but I'd rather do this than manual DOM manipulation

Something like:

 - in the template, put in a root, which will not render anything if JS is disabled (which is fine)
   ```html
   <div id="passkey-root"></div>
   ```
 - then render a react component in it, like
   ```typescript
   createRoot(document.getElementById("passkey-root" as HTMLElement).render(
     <Suspense fallback="Loading…">
       <I18nextProvider i18n={i18n}>
         <WebAuthnButton options={options} />
       </I18nextProvider>
     </Suspense>
   )
   ```

or something like that. Vite should properly code-split and this shouldn't result into a giant JS payload for this page

frontend/src/utils/webauthn.ts
```diff
@@ -0,0 +1,128 @@
...
+  };
+
+  type AuthenticatorAssertionResponseJSON = {
+    clientDataJSON: Base64URLString;
+    authenticatorData: Base64URLString;
+    signature: Base64URLString;
+    userHandle?: Base64URLString;
+  };
+
+  if (!window.PublicKeyCredential.parseCreationOptionsFromJSON) {
```
sandhose, line 36: Looks like this polyfill exists to do that: https://github.com/MasterKale/webauthn-polyfills

sandhose (commented):
I haven't reviewed the frontend part in-depth yet, but most of this looks really solid! In terms of UI, for now my main request is that the webauthn flow starts on the login page; this is my very quick non-designer proposal:

image

crates/data-model/src/users.rs
tonkku107: Whichever works, the webauthn api provides both a base64 string and an ArrayBuffer (though both get sent as base64 thanks to toJSON). credential_id needs to be queried so a string format probably works better

crates/handlers/src/webauthn.rs
tonkku107: You'd need to know the user ID beforehand anyway for the discovery to work.

frontend/src/utils/webauthn.ts
tonkku107: I did use that at first, but it turned out that it didn't follow the spec and was missing fields.

crates/data-model/src/users.rs
tonkku107: I changed credential_id and transports to their exact types. There's still potential for the binary encoded ones, though the static state has a generic for the type of public key included and none of the data model structs so far have any generics so I'm just checking if that's desireable first. The metadata would be silly to decode each time since it's not being used so that can probably stay in its raw form.

crates/storage-pg/migrations/20250303142622_passkeys.sql
tonkku107: Done

crates/handlers/Cargo.toml
tonkku107: Done

crates/handlers/src/webauthn.rs
tonkku107: Done

crates/handlers/src/webauthn.rs
tonkku107: Maybe @zacknewman could give us an option to ignore the built-in time source?

crates/handlers/src/webauthn.rs
tonkku107: It does borrow a lot of stuff, at least during registration.

crates/handlers/src/webauthn.rs
tonkku107: Done

crates/handlers/src/webauthn.rs
tonkku107: Done

crates/handlers/src/views/login/passkey/mod.rs
tonkku107: I thought that the no js requirement on login so far meant the js should go on a dedicated page. I've moved everything to the login page and implemented the non-designer design

frontend/src/template_passkey.ts
tonkku107: While there was no official design yet I thought about experimenting with minimal basic js in case sceptical people would like to audit the js being loaded. The new version on the login page uses React now.

crates/handlers/src/webauthn.rs
zacknewman: Can you expand? By enforcing the timeout internally, it is less likely timeout enforcement will be done incorrectly. You can always use `NonZeroU32::MAX` as the timeout, and this should be more than large enough—this corresponds to just under 50 days—to ensure that a ceremony won't err due to a timeout. You can then enforce your own actual timeout before or after. The time source that is used is `std::time::SystemTime` iff `serializable_server_state` is enabled otherwise it's `std::time::Instant`. I'm unsure of the problem this causes.

I take great care in ensuring a `panic` doesn't occur by using methods like `SystemTime::checked_add`, but there are niche cases where `SystemTime::now` itself `panic`s (e.g., I believe it `panic`s on `wasm32-unknown-unknown` targets). Is that the problem you're trying to avoid?

PR #4241, opened 2025-03-17
Add `expires_in` to introspection responses

reivilibre:


crates/oauth2-types/src/requests.rs
```diff
@@ -767,6 +767,10 @@ pub struct IntrospectionResponse {
     #[serde_as(as = "Option<TimestampSeconds>")]
     pub exp: Option<DateTime<Utc>>,
 
+    /// Relative timestamp indicating when the token will expire,
+    /// in seconds from the current instant.
+    pub expires_in: Option<i64>,
```
sandhose, line 772: I would use a `std::time::Duration` (or `chrono::Duration`, depends which one is more convenient) and convert at serialisation time, like we do for the various timestamps here

```suggestion
    #[serde_as(as = "Option<serde_with::DurationSeconds<i64>>")]
    pub expires_in: Option<Duration>,
```

sandhose (changes requested):
Other than the small serialisation tweak, lgtm

crates/oauth2-types/src/requests.rs
reivilibre: nifty

sandhose (approved):

PR #4249, opened 2025-03-17
support RP initiated logout to upstream OIDC Provider when logging out from MAS

mcalinghee:


crates/handlers/src/upstream_oauth2/logout.rs
```diff
@@ -0,0 +1,148 @@
...
+    RouteError: std::convert::From<E>,
+{
+    let mut result: UpstreamLogoutInfo = UpstreamLogoutInfo::default();
+    // Set the post-logout redirect URI to our app's logout completion page
+    let post_logout_redirect_uri = url_builder
+        .absolute_url_for(&mas_router::Login::default())
+        .to_string();
+    result.post_logout_redirect_uri = Some(post_logout_redirect_uri.clone());
+
+    let sessions_cookie = UpstreamSessionsCookie::load(cookie_jar);
```
sandhose, line 84: Loading from the cookie is unreliable. You should instead:

 - lookup the authentications for the current `user_session` in the `user_session_authentications`
 - authentications that were done through an upstream will have a reference to a `upstream_oauth_authorization_session_id`
 - there you can grab the `id_token` received during the auth and extract the subject
 - you should also pass the id_token to the RP-initiated logout endpoint through the `id_token_hint`

crates/handlers/src/upstream_oauth2/logout.rs
```diff
@@ -0,0 +1,148 @@
...
+    }
+
+    // Check if we found a provider with allow_rp_initiated_logout
+    if let Some(provider) = provider {
+        // Look for end session endpoint
+        // In a real implementation, we'd have end_session_endpoint fields in the
+        // provider For now, we'll try to construct one from the issuer if
+        // available
+        if let Some(issuer) = &provider.issuer {
+            let end_session_endpoint = format!("{issuer}/protocol/openid-connect/logout");
```
sandhose, line 121: So that I don't forget: this must use the `end_session_endpoint` metadata. Please also make sure to allow overriding this, like all the other endpoints

crates/handlers/src/views/logout.rs
```diff
@@ -39,6 +41,22 @@ pub(crate) async fn post(
                     .record_browser_session(&clock, &session)
                     .await;
 
+                // First, get RP-initiated logout endpoints before actually finishing the
```
sandhose, line 44: We need to make sure that this logic also happens when the logout is done through the React frontend. I don't have a good suggestion on how to do so for now though

sandhose (changes requested):
Thanks for doing this! This isn't a full review, rather a quick look with a few suggestions

crates/handlers/src/upstream_oauth2/logout.rs
mcalinghee: Thank you for your review!
I have amend the code but I was not sure if I need to extract the subject to perform the logout

crates/handlers/src/views/logout.rs
mcalinghee: I am looking on how to integrate this [here](https://github.com/tchapgouv/matrix-authentication-service/blob/9b06223c1df940c15da512dd56659b4aa3dece57/crates/handlers/src/graphql/mutations/browser_session.rs#L87) but I am not sure this is the way to go

PR #4255, opened 2025-03-17
Admin API for adding and removing upstream oauth links

tonkku107:
Includes link removal storage API from #3245 with the review comment addressed by me

### DIFF
--- a/crates/handlers/src/admin/v1/upstream_oauth_links/add.rs
+++ b/crates/handlers/src/admin/v1/upstream_oauth_links/add.rs
@@ -1,0 +1,368 @@
+// Copyright 2025 New Vector Ltd.
+//
+// SPDX-License-Identifier: AGPL-3.0-only
+// Please see LICENSE in the repository root for full details.
+
+use aide::{NoApi, OperationIo, transform::TransformOperation};
+use axum::{Json, response::IntoResponse};
+use hyper::StatusCode;
+use mas_storage::{BoxRng, upstream_oauth2::UpstreamOAuthLinkFilter};
+use schemars::JsonSchema;
+use serde::Deserialize;
+use ulid::Ulid;
+
+use crate::{
+    admin::{
+        call_context::CallContext,
+        model::{Resource, UpstreamOAuthLink, User},
+        response::{ErrorResponse, SingleResponse},
+    },
+    impl_from_error_for_route,
+};
+
+#[derive(Debug, thiserror::Error, OperationIo)]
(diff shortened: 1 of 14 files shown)


crates/storage-pg/src/upstream_oauth2/link.rs
```diff
@@ -374,4 +374,53 @@ impl UpstreamOAuthLinkRepository for PgUpstreamOAuthLinkRepository<'_> {
...
+
+        // Then delete the link itself
+        let res = sqlx::query!(
+            r#"
+                DELETE FROM upstream_oauth_links
+                WHERE upstream_oauth_link_id = $1
+            "#,
+            Uuid::from(upstream_oauth_link.id),
+        )
+        .traced()
```
sandhose, line 418: Same here

crates/handlers/src/admin/v1/upstream_oauth_links/add.rs
```diff
@@ -0,0 +1,368 @@
...
+    let mut link = repo
+        .upstream_oauth_link()
+        .add(
+            &mut rng,
+            &clock,
+            &provider,
+            params.subject,
+            params.human_account_name,
+        )
+        .await?;
```
sandhose, line 143: I tried locally, and one thing this doesn't check is the `(provider, subject)` unique constraint. The filter+count up there checks if the user already have a link with the provider, even though it's (relatively) valid to have multiple links with the same provider

There is also the question of what to do if there is an existing link for a given subject, but with no user associated. I think in this case we should use the existing link and associate it to the user?

So, I would change this whole block to:

 - not check for existing links on the user+provider
 - use `UpstreamOAuthLinkRepository::find_by_subject` to find an existing link for the given subject
   - if it doesn't exist, create it
   - if it exists but with no existing user associated, use the existing one and associate the user
   - if it exists but with the same user associated, use the existing one
   - if it exists but with another user associated, error out
 - if we created the link, return a `201 CREATED`, else a `200 OK` if we reused one

tonkku107: I have no idea how I missed the unassociated link case despite it staring at me right in the face. I'll fix that asap tomorrow

tonkku107: Done in 7c4a9bf
> if it exists but with the same user associated, use the existing one

I didn't think this made sense though. In that case the endpoint wouldn't do anything and I think that should be considered a failure case.

sandhose (approved):
Thank you!

PR #4268, opened 2025-03-18
Don't delete devices marked as dehydrated devices

uhoreg:
MAS tries to delete all the devices that it doesn't know about.  This PR makes it ignore the dehydrated device (as reported by Synapse), so that it won't get deleted.

Companion PR: https://github.com/element-hq/synapse/pull/18252


crates/matrix-synapse/src/lib.rs
```diff
@@ -410,8 +414,12 @@ impl HomeserverConnection for SynapseConnection {
             .await
             .context("Failed to parse response while querying devices from Synapse")?;
 
-        let existing_devices: HashSet<String> =
-            body.devices.into_iter().map(|d| d.device_id).collect();
+        let existing_devices: HashSet<String> = body
+            .devices
+            .into_iter()
+            .filter(|d| d.dehydrated.is_none())
```
reivilibre, line 420: we could also alternatively consider hiding dehydrated devices from the admin API, but in principle I think it does make sense to show them there, so happy with this

reivilibre (approved):

crates/matrix-synapse/src/lib.rs
reivilibre: that said, shouldn't we have `d.dehydrated != Some(true)` here, or make `dehydrated: bool` with a default of false? (probably in favour of the latter since that's the simplest and the semantics match)

reivilibre (changes requested):

crates/matrix-synapse/src/lib.rs
reivilibre: though the struct is serialised again, so `Option<bool>` makes sense unless we split the usage — I'll change the filter condition, push, merge

PR #4289, opened 2025-03-24
Speed up local docker rebuilds & CI cache exports

sandhose:
We almost never cache the Rust build in Docker, because it always changes, because the version injected is always different.
The thing is, we always export the build cache in the CI, which always includes this huge layer, as it includes the `target/` directory.

This changes so that the `cargo build` uses a cache mount, which means that:

 - local rebuilds are faster
 - CI builds don't try to cache the `target/` directory, making them (in theory) faster to export. **Before: 214s to export cache, after: 27s**

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -114,20 +114,18 @@ ENV VERGEN_GIT_DESCRIBE=${VERGEN_GIT_DESCRIBE}
 
 # Network access: cargo auditable needs it
 RUN --network=default \
+  --mount=type=cache,target=/root/.cargo/registry \
+  --mount=type=cache,target=/app/target \
   cargo auditable build \
-  --locked \
-  --release \
-  --bin mas-cli \
-  --no-default-features \
-  --features docker \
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -114,20 +114,18 @@ ENV VERGEN_GIT_DESCRIBE=${VERGEN_GIT_DESCRIBE}
 
 # Network access: cargo auditable needs it
 RUN --network=default \
+  --mount=type=cache,target=/root/.cargo/registry \
+  --mount=type=cache,target=/app/target \
```
reivilibre, line 118: I can't easily tell, but should these have `id=mas-<unique ID>` so that they don't share the same cache with other dockerfiles that have caches in these locations?
(Under podman it seems like they might just hash the `target` option and use that as the cache key)

reivilibre (approved):

Dockerfile
sandhose: Erf, not sure it really matters honestly? 🤷 
We can always change that later

PR #4344, opened 2025-04-04
Allow setting unix socket mode

dblsaiko:


crates/cli/src/server.rs
```diff
@@ -329,8 +330,20 @@ pub fn build_listeners(
...
+                if let Some(mode) = mode {
+                    let mut permissions = fs::metadata(socket)
+                        .context("could not read socket metadata")?
+                        .permissions();
+                    let mode = u32::from_str_radix(mode, 8)
+                        .with_context(|| format!("could not parse mode: {mode}"))?;
+                    permissions.set_mode(mode);
+                    fs::set_permissions(socket, permissions)
+                        .context("could not set socket permissions")?;
+                }
```
sandhose, line 345: To avoid any potential race conditions, I would appreciate if we used the opened file descriptor directly.

This is a little bit annoying to do, but basically, you can:

 - keep opening the socket with `UnixListener::bind`
 - transform it [into an `OwnedFd`](https://doc.rust-lang.org/stable/std/os/unix/net/struct.UnixListener.html#impl-From%3CUnixListener%3E-for-OwnedFd)
 - transform it [into a `File`](https://doc.rust-lang.org/stable/std/fs/struct.File.html#impl-From%3COwnedFd%3E-for-File)
 - do the permissions operations using [`File::set_permissions`](https://doc.rust-lang.org/stable/std/fs/struct.File.html#method.set_permissions)
 - transform it back to an `OwnedFd` then back to a `UnixListener`

This way we're sure that if for some reason the socket gets deleted, gets moved or whatever, we don't end up doing operations on the wrong file. It only ever gets `open`ed once

crates/config/src/sections/http.rs
```diff
@@ -124,6 +124,9 @@ pub enum BindConfig {
         /// Path to the socket
         #[schemars(with = "String")]
         socket: Utf8PathBuf,
+
+        /// Socket file mode
```
sandhose, line 128: Please add in the comment what form you expect this to be. You could also use [schemars' `example` attribute](https://graham.cool/schemars/v0/deriving/attributes/#example), which will appear in the generate JSONSchema

docs/reference/configuration.md
```diff
@@ -58,6 +58,7 @@ http:
 
         # Third option: listen on the given UNIX socket
         - socket: /tmp/mas.sock
+          mode: "660" # optional
```
sandhose, line 61:
```suggestion
          mode: "660" # permissions to set on the socket, optional
```

sandhose (changes requested):
Thanks for adding this!

crates/cli/src/server.rs
2xsaiko: That unfortunately doesn't work, calling fchmod on a socket fd does nothing. See https://stackoverflow.com/a/15450698

(Still, change here: https://github.com/2xsaiko/matrix-authentication-service/commit/push-looktxnmnmsw)

sandhose (approved):
Erf, this really has fell down through the cracks, sorry about that. Even if there is the slight chance of race between the socket creation and mode setting, it's fine in practice; thanks for exploring the fchmod route still

sandhose:
@dblsaiko I actually need you to sign the CLA if we want to merge this

sandhose:
This got covered by #5858, closing this one

PR #4358, opened 2025-04-07
syn2mas: Allow skipping guest users

jaywink:
Passing `--skipGuest` will treat finding guest users in the Synapse database as warnings instead of aborting.

Needed for some cases where a Synapse host has had guest access enabled at some point in history and thus the database has some of these users. Skipping them in syn2mas is a much easier task than trying to clean them up from the database.


jaywink:
The old tool is gone so this has become unnecessary (at least against the old tool).

PR #4369, opened 2025-04-08
Allow requests to the compat login endpoint without a Content-Type header

### LINKED ISSUE #4340: compat login spec compliance: `Content-Type: application/json` should be optional, but MAS requires it
> Clients should supply a Content-Type header of application/json for all requests with JSON bodies, but this is not required.
> 
> — https://spec.matrix.org/v1.14/client-server-api/

The `/_matrix/client/v3/login` endpoint (and likely others, if there are any) fail if this header is not set.

sandhose:
Fixes #4340

### DIFF
--- a/crates/handlers/src/compat/mod.rs
+++ b/crates/handlers/src/compat/mod.rs
@@ -53,11 +53,8 @@ pub enum MatrixJsonBodyRejection {
     #[error("Failed to read request body")]
     BytesRejection(#[from] BytesRejection),
 
-    #[error("Body is not a valid JSON document")]
-    NotJson(#[source] serde_json::Error),
-
-    #[error("Request body has invalid parameters")]
-    BadJson(#[source] serde_json::Error),
+    #[error("Invalid JSON document")]
+    Json(#[from] serde_json::Error),
 }
 
 impl IntoResponse for MatrixJsonBodyRejection {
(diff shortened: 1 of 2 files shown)


crates/handlers/src/compat/mod.rs
```diff
@@ -27,3 +37,112 @@ impl IntoResponse for MatrixError {
...
+            }
+        }
+
+        let bytes = Bytes::from_request(req, state).await?;
+
+        // We first parse it as a JSON value so that we can distinguish between
+        // invalid JSON documents and invalid requests
+        let value: serde_json::Value =
+            serde_json::from_slice(&bytes).map_err(MatrixJsonBodyRejection::NotJson)?;
+        let value = serde_json::from_value(value).map_err(MatrixJsonBodyRejection::BadJson)?;
```
reivilibre, line 145: I guess these requests are tiny so this is fine. But if they were big, I'd rather avoid this and you could use `.is_syntax()` on the serde_json Error.

reivilibre (approved):

crates/handlers/src/compat/mod.rs
sandhose: You're right, better do that. I used `.is_data()` instead

7f4e97535

PR #4393, opened 2025-04-13
Require auth key file for Apple Sign-In

### LINKED ISSUE #4391: Apple SSO Provider Docs Incorrect/Incomplete
**Describe the bug**

- Missing `issuer` field in docs example (config will not sync otherwise)
- Missing `id` field in docs example (config will not sync otherwise)
- Fail to explain that `client_id` field must correspond to the Service ID
- `client_name` field is abitrary
- `human_name` is required to render HTML templates for login/register pages (missing in docs)

The following is just also straight up **not** working or is missing:
 - Even when an issuer is manually specified (https://appleid.apple.com), it will fail to authenticate you with apple when you do try to sign in. One must (as far as I have brute-forced so far) manually specify `authorization_endpoint` and `token_endpoint` fields in the config, which is also not documented anywhere.
 - Finally, even when you do get this to work, it silently just fails with `Error from the provider: none`, even though the POST done to MAS contains both `code` and `state` headers.

**To Reproduce**

Attempt to setup Apple SSO following the official documentation.

**Expected behavior**

That Apple SSO should work given the instructions in the official documentation.

**Screenshots**

N/A

 - OS: MacOS 15.1.1 (24B91)
 - Browser: Firefox
 - Version: 1.10.3b (Firefox 136.0.4) (aarch64)

**Additional context**

N/A

adisve:
This change requires using a private key file for Apple Sign-In authentication instead of directly pasting the key content.

- Using a file is more reliable and secure
- Copy-pasting key content can easily introduce formatting errors (extra spaces, line breaks, encoding issues)
- From experience, copy-pasting the key did not even work properly, only until I changed this myself and built a new binary
- Loading from a file ensures the key is parsed exactly as expected, reducing the chance of silent failures or invalid signature errors, as Apple does not even provide an error that this is what occurs unless you'd manually debug it like I did - verifying the issue is with the MAS binary.

Fixes #4391


crates/config/src/sections/upstream_oauth2.rs
```diff
@@ -383,8 +383,8 @@ fn signed_response_alg_default() -> JsonWebSignatureAlg {
 
 #[derive(Debug, Clone, Serialize, Deserialize, JsonSchema)]
 pub struct SignInWithApple {
-    /// The private key used to sign the `id_token`
-    pub private_key: String,
+    /// The private key file used to sign the `id_token`
+    pub private_key_file: String,
```
sandhose, line 387: This is a breaking change I'd like to avoid introducing. Could you do instead allow both properties? This is what we do for some secrets:

https://github.com/element-hq/matrix-authentication-service/blob/a3202a690ad677dacbed264c877e5e6831986012/crates/config/src/sections/secrets.rs#L34-L39

Then in the configuration section validation function, we check that only one of them is specified:

https://github.com/element-hq/matrix-authentication-service/blob/a3202a690ad677dacbed264c877e5e6831986012/crates/config/src/sections/secrets.rs#L145-L155

crates/handlers/src/upstream_oauth2/mod.rs
```diff
@@ -127,7 +128,11 @@ fn client_credentials_for_provider(
             let params = client_secret.ok_or(ProviderCredentialsError::MissingClientSecret)?;
             let params: SignInWithApple = serde_json::from_str(&params)?;
 
-            let key = elliptic_curve::SecretKey::from_pkcs8_pem(&params.private_key)?;
+            let private_key_pem = fs::read_to_string(&params.private_key_file)
```
sandhose, line 131: We usually avoid doing sync I/O but rather use Tokio. That means making this function asynchronous and use `tokio::fs::read_to_string` instead

docs/setup/sso.md
```diff
@@ -84,18 +84,19 @@ Sign-in with Apple uses special non-standard for authenticating clients, which r
 ```yaml
 upstream_oauth2:
   providers:
-    - client_id: 01JAYS74TCG3BTWKADN5Q4518C
-      client_name: "<Service ID>" # TO BE FILLED
+    - id: 01JAYS74TCG3BTWKADN5Q4518C
+      issuer: "https://appleid.apple.com"
+      authorization_endpoint: "https://appleid.apple.com/auth/authorize"
+      token_endpoint: "https://appleid.apple.com/auth/token"
```
sandhose, line 90:
```suggestion
```

Those are properly advertised by Apple in their discovery document, so no need to specify those: https://account.apple.com/.well-known/openid-configuration

docs/setup/sso.md
```diff
@@ -84,18 +84,19 @@ Sign-in with Apple uses special non-standard for authenticating clients, which r
...
+      human_name: "Apple"
```
sandhose, line 91: We also have support for `brand_name: apple`, which will show the right Apple logo on the login option

sandhose (changes requested):
Thanks! I don't know why the documentation was so far off the actual configuration 😅

crates/handlers/src/upstream_oauth2/mod.rs
```diff
@@ -124,10 +142,32 @@ fn client_credentials_for_provider(
...
+                return Err(ProviderCredentialsError::DuplicatePrivateKey);
+            }
 
-            let key = elliptic_curve::SecretKey::from_pkcs8_pem(&params.private_key)?;
+            let private_key_pem = if let Some(private_key) = params.private_key {
+                private_key
+            } else if let Some(private_key_file) = params.private_key_file {
+                tokio::fs::read_to_string(private_key_file)
+                    .await
+                    .map_err(|_| ProviderCredentialsError::MissingPrivateKey)?
```
sandhose, line 164: Sorry I missed that, but this should *not* be done when reading from the database. In MAS, the providers are stored in the database, and the config file is just there to sync on startup.

So:

 - don't touch this part, we should store in the database the key already read
 - read the file when we sync the config file with the database, probably somewhere around here:
https://github.com/element-hq/matrix-authentication-service/blob/a3202a690ad677dacbed264c877e5e6831986012/crates/cli/src/sync.rs#L192-L196

sandhose (changes requested):

adisve:
@sandhose I have now implemented the private key reading in the `sync.rs` file instead, and it will not break any current functionality - but if a private key file is given, it will be read. However, the raw private key takes precedence.

crates/cli/src/sync.rs
```diff
@@ -189,10 +189,18 @@ pub async fn config_sync(
...
+                    // private key to hold the content of the private key file.
+                    // private key (raw) takes precedence so both can be defined
+                    // without issues
+                    if siwa.private_key.is_none() {
+                        if let Some(private_key_file) = siwa.private_key_file.take() {
+                            let key = tokio::fs::read_to_string(private_key_file).await?;
+                            siwa.private_key = Some(key);
+                        }
+                    }
+                    let encoded = serde_json::to_vec(&siwa)?;
```
sandhose, line 203: This does mean that it's possible to end up in the database with *no* private key if none are set. But I'm fine with that, SiWA is sufficiently niche anyway :)

sandhose (approved):
Thanks for doing this!

Issue #4415, opened 2025-04-15
client_name not being upserted for static client registrations


defaultdino:
**Is your feature request related to a problem? Please describe.**
I have noticed that when using statically registered clients, the templated consent page does not properly display the client name. After taking a closer look at the sync logic, I can see that it is not part of the upsert method when syncing the clients with the db.

Is this behavior intended for some specific reason that I am missing?

**Describe the solution you'd like**
It would be nice if static clients also could have their name displayed when the user is prompted with a consent screen, instead of defaulting to the `client_id` on templating.

**Describe alternatives you've considered**
Nothing in particular aside from not using static registrations

**Additional context**
N/A

PR #4417, opened 2025-04-16
Insert client_name when upserting statically registered clients

defaultdino:
When using statically registered clients, the MAS consent screen fails to display a readable client_name when prompting the user for permission. It defaults to `client_id` instead of `client_name`, since the `client_name` field is not being populated during upsert (config sync).

I propose these changes so that statically registered clients are slightly more streamlined to dynamically registered ones, allowing the user to see an actual informative name instead of the ULID `client_id` for the client.

This would close my issue [here](https://github.com/element-hq/matrix-authentication-service/issues/4415)


sandhose (approved):
It would be nice to also being able to set other metadata like the `client_uri`, `logo_uri` etc., but I'm fine leaving it like that for now

defaultdino:
> It would be nice to also being able to set other metadata like the `client_uri`, `logo_uri` etc., but I'm fine leaving it like that for now

Easy to add, I can make a lil PR for that some day soon

Issue #2238, opened 2024-09-10
Feature: logging REST calls


matrixbot:
*This issue was originally created by [**@dklimpel**](https://github.com/dklimpel) at <https://github.com/matrix-org/matrix-authentication-service/issues/2238>.*

It is very difficult to debug or analyse MAS, because it does not log any REST calls.

It is a kind of webserver without logging.

matrixbot:
*This comment was originally posted by [**@matrixbot**](https://github.com/matrixbot) at <https://github.com/matrix-org/matrix-authentication-service/issues/2238#issuecomment-2341166054>.*


For your information, this issue has been copied over to the Element fork of `matrix-authentication-service`: <https://github.com/element-hq/matrix-authentication-service/issues/2238>

reivilibre:
A good idea might be to copy the Synapse log line format, for familiarity.

Ref: https://github.com/matrix-org/synapse/blob/49e1356ee3d5d72929c91f778b3a231726c1413c/docs/usage/administration/request_log.md

Not all the fields would be applicable (and some are probably hard to implement right now without extra tracking), so those could be left off for now, but at least keeping the format similar/inspired by Synapse should help keep the knowledge transferable.

PR #4424, opened 2025-04-17
New logging output

sandhose:
This completely overhauls how we log things.

First, it introduces a new task-level `LogContext`, which records task-level metrics, as well as a unique task ID.
This context is then logged when present, making it easier to correlate logs with the task that generated them.

For now, the log context only records the number of polls, the overall task duration and the CPU time used, but we could add more metrics in the future. I'm just trying to keep this PR "small".

Then, there is a new custom event formatter, which includes this log context, but also customizes how log looks in general.
Importantly, **it doesn't include context from the spans anymore**. This cleans up the log a lot, but may mean we're missing some context in some cases. I've done my best to restore said context, but it's not perfect.

One goal was to remove any `#[instrument(err)]` annotations. Two main reasons for this:

 - not all errors are 'critical', some of them need to be recorded as warnings. We have a lot of noise in Sentry because of this
 - `#[instrument(err)]` logs the error description and that's it. We loose a lot of information from the error stack

This is mostly replaced by a `record_error!` macro, which also decides whether the error is a warning or not.

For now, I've kept some `#[instrument(err)]` annotations in the `mas-matrix-synapse` and `mas-storage-pg` crates, as they are almost always 'internal'/bad errors that we want to log, and this PR is already massive.

Last but not least, we finally have a log entry for *every single HTTP request* 🎉

Fixes https://github.com/element-hq/matrix-authentication-service/issues/2238

This can be reviewed commit by commit, even though some things are not exactly in the right order.


crates/context/src/lib.rs
```diff
@@ -0,0 +1,126 @@
...
+    index: u64,
+
+    /// The time when the context was created
+    start: Instant,
+
+    /// The number of [`Future::poll`] recorded
+    polls: AtomicU64,
+
+    /// An approximation of the total CPU time spent in the context
+    cpu_time: AtomicU64,
```
reivilibre, line 58: in what units?

crates/context/src/lib.rs
```diff
@@ -0,0 +1,126 @@
...
+
+        #[expect(clippy::cast_precision_loss)]
+        let cpu_time_ms = self.inner.cpu_time.load(Ordering::Relaxed) as f64 / 1_000_000.;
+
+        let polls = self.inner.polls.load(Ordering::Relaxed);
+        let tag = &self.inner.tag;
+        let index = self.inner.index;
+        write!(
+            f,
+            "{tag}-{index} ({polls} polls, CPU: {cpu_time_ms:.3} ms, total: {elapsed:.3} ms)"
```
reivilibre: 3 places after the decimal point seems like way more granularity than we care about TBH, maybe `.1`?

reivilibre (approved):
exciting! Looks very plausible, if there's anything minor that needs improvement I'm sure this can happen in a subsequent PR

PR #4462, opened 2025-04-27
Add the missing space in zh translation

Kimiblock:
Problematic translation:

image

Signed-off-by: Kimiblock Moe <kimiblock@icloud.com>


sandhose:
Hello, thanks for this! The translations are managed in Localazy, so please fix it there and it will be included in the next release: https://localazy.com/p/matrix-authentication-service

PR #4517, opened 2025-05-07
Don't automatically insert the `openid` scope on upstream providers

sandhose:
This also includes another pass-by change: the `nonce` is no longer generated if the scope doesn't include `openid`.

This could technically be a breaking change in some cases, but that would have been relying on a weird behavior


crates/oidc-client/src/requests/authorization_code.rs
```diff
@@ -191,7 +191,7 @@ pub struct AuthorizationValidationData {
     pub state: String,
 
     /// A string to mitigate replay attacks.
-    pub nonce: String,
+    pub nonce: Option<String>,
```
reivilibre, line 194:
```suggestion
    /// Used when the `openid` scope is set (and therefore we are using OpenID Connect).
    pub nonce: Option<String>,
```

reivilibre (approved):

PR #4527, opened 2025-05-07
Don't hold database connections open when talking to the homeserver

sandhose:
- **Introduce a RepositoryFactory**
- **Move the pool acquisition metric logic to the PgRepositoryFactory**
- **Use the new RepositoryFactory everywhere**
- **Don't hold db conns when creating a device on the compat login API**

### DIFF
--- a/crates/handlers/src/compat/login.rs
+++ b/crates/handlers/src/compat/login.rs
@@ -14,11 +14,12 @@ use mas_axum_utils::record_error;
 use mas_data_model::{CompatSession, CompatSsoLoginState, Device, SiteConfig, TokenType, User};
 use mas_matrix::HomeserverConnection;
 use mas_storage::{
-    BoxClock, BoxRepository, BoxRng, Clock, RepositoryAccess,
+    BoxClock, BoxRepository, BoxRepositoryFactory, BoxRng, Clock, RepositoryAccess,
     compat::{
         CompatAccessTokenRepository, CompatRefreshTokenRepository, CompatSessionRepository,
         CompatSsoLoginRepository,
     },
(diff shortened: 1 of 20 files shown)


crates/handlers/src/compat/login.rs
```diff
@@ -368,12 +371,53 @@ pub(crate) async fn post(
...
+
+    // Now we can create the device on the homeserver, without holding the
+    // transaction
+    if let Err(err) = homeserver
+        .create_device(&user_id, device.as_str(), session.human_name.as_deref())
+        .await
+    {
+        // Something went wrong, let's end this session and schedule a device sync
+        let mut repo = repository_factory.create().await?;
+        let session = repo.compat_session().finish(&clock, session).await?;
```
sandhose, line 406: There is one slight difference in behaviour now: if the device creation failed on Synapse, before we'd just rollback the transaction, meaning that the client could retry the request with the same login token and it would work. This isn't the case anymore

It may also end another session, since we now end sessions with the same device ID

To be fair, this request was never really idempotent, so if the client lost the response it would have had the same effect

crates/storage/src/repository.rs
```diff
@@ -29,6 +30,17 @@ use crate::{
     },
 };
 
+/// A [`RepositoryFactory`] is a factory that can create a [`BoxRepository`]
+// XXX(quenting): this could be generic over the repository type, but it's annoying to make it dyn-safe
```
reivilibre: meh, don't overcomplicate it any more than it needs to be. For this, I don't think we should worry about the boxing

reivilibre (approved):

Issue #4528, opened 2025-05-07
MAS fails to support query params in redirect_uri - incompliant with MSC2966


t3chguy:
**Describe the bug**
MSC2966 explicitly says that query params in redirect_uri are valid

![Image](https://github.com/user-attachments/assets/2f775068-f1e8-48ec-b65a-4175fb688d4f)

MAS forbids them.

```
[...]

PR #4529, opened 2025-05-07
Fix MSC2966 compliance around redirect_uri validity

t3chguy:
Fixes https://github.com/element-hq/matrix-authentication-service/issues/4528

+ Allows URI without trailing slash on empty path
+ Allows URI containing a query segment at end


sandhose:
Whilst we're at it, could you also allow any port on https urls? It's in the secure_url util

t3chguy:
@sandhose done

sandhose (approved):
tyvm!

PR #4530, opened 2025-05-07
Docu: Fix Swagger UI callback path

lomion0815:
The config file example in the documentation was using the wrong path. Verified with the docker image with main tag.


sandhose (approved):
Thanks!

PR #4549, opened 2025-05-09
Support for experimental plan management tab in UI

hughns:
This PR adds a feature to allow a plan management tab to be added to the My Account screens.

When enabled it looks like this:

image

It attempts to resize the iframe to match the size of the embedded content. However, in current testing this has a habit of growing over time.


frontend/src/routes/_account.plan.index.tsx
```diff
@@ -0,0 +1,80 @@
...
+  }
+`);
+
+export const query = queryOptions({
+  queryKey: ["siteConfig"],
+  queryFn: ({ signal }) => graphqlRequest({ query: QUERY, signal }),
+});
+
+export const Route = createFileRoute("/_account/plan/")({
+  loader: ({ context }) => context.queryClient.ensureQueryData(query),
```
sandhose: nit perf thing, but we could preload the iframe document here, e.g

```suggestion
  loader: async ({ context }) => {
    const { planManagementIframeUri } = await context.queryClient.ensureQueryData(query);
    if (planManagementIframeUri) 
      preload(planManagementIframeUri, { as: "document" });
  }
```

with `import { preload } from "react-dom";`

frontend/src/routes/_account.plan.index.tsx
```diff
@@ -0,0 +1,80 @@
...
+  }
+`);
+
+const QUERY = graphql(/* GraphQL */ `
+  query SiteConfig {
+    siteConfig {
+      ...PlanManagement_siteConfig
+    }
+  }
+`);
```
sandhose: You don't need to have a fragment here, fragments are for "leaf" components

Also please name this query with a unique name, and unique query key that reflects that this is for the plan management page

frontend/src/routes/_account.plan.index.tsx
```diff
@@ -0,0 +1,80 @@
...
+  }, []);
+  useEffect(() => {
+    calculateHeight();
+
+    const interval = setInterval(() => {
+      calculateHeight();
+    }, 1000);
+
+    return () => clearInterval(interval);
+  }, [calculateHeight]);
```
sandhose: Instead of polling, we could use a MutationObserver to recalculate. Plus, to setup the observer with a ref callback, so that we get it as soon as we have the DOM node ready.

We also don't need to track the iframe height as state to avoid re-renders, so we end up with something like

```typescript
  const calculateHeight = useCallback((iframe: HTMLIFrameElement) => {
    const height =
      iframe.contentWindow?.document.body.parentElement?.scrollHeight;

    if (height) {
      iframe.height = `${height}px`;
    } else {
      iframe.height = "500px";
    }
  }, []);

  const ref: Ref<HTMLIFrameElement> = useCallback(
    (iframe: HTMLIFrameElement) => {
      calculateHeight(iframe);

      if (iframe.contentWindow) {
        const iframeDocument = iframe.contentWindow.document;

        const observer = new MutationObserver((_mutationsList) => {
          calculateHeight(iframe);
        });

        observer.observe(iframeDocument.body, {
          childList: true,
          subtree: true,
          attributes: true,
        });

        return () => observer.disconnect();
      }
    },
    [calculateHeight],
  );

  return (
    <iframe
      title="iframe" // no proper title as this is experimental feature
      ref={ref}
      onLoad={(e) => calculateHeight(e.target as HTMLIFrameElement)}
      src={planManagementIframeUri}
      scrolling="no"
    />
  );
```

frontend/src/routes/_account.plan.index.tsx
sandhose: also you should `throw redirect({ to: "/" })` if there is no plan management URI here, this should allow you to avoid loading the site config in the `_account.index.tsx` and just redirect to `/plan` there?

frontend/src/routes/_account.tsx
```diff
@@ -26,6 +27,7 @@ const QUERY = graphql(/* GraphQL */ `
 
     siteConfig {
       ...UserGreeting_siteConfig
+      ...PlanManagement_siteConfig
```
sandhose: Using the fragment here is unnecessary, just get the property

crates/config/src/sections/experimental.rs
```diff
@@ -75,6 +76,10 @@ pub struct ExperimentalConfig {
     /// Disabled by default
     #[serde(skip_serializing_if = "Option::is_none")]
     pub inactive_session_expiration: Option<InactiveSessionExpirationConfig>,
+
+    /// Experimental feature to show a plan management tab and iframe
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub plan_management_iframe_uri: Option<Url>,
```
sandhose: Not convinced that this *has* to be a URL? Like if we want to have a relative link here, it won't get accepted?

sandhose (changes requested):
Looks alright, a few nitpicks on how things should be on the react side

crates/config/src/sections/experimental.rs
hughns: I suppose someone might want to configure it with a relative URL.

Are you thinking that it is just a string that is passed through without validation? Or are you thinking the value is parsed from config relative to `http.public_base` and then the site config still return a full URL? Or something else entirely?

frontend/src/routes/_account.plan.index.tsx
hughns: > this should allow you to avoid loading the site config in the _account.index.tsx and just redirect to /plan there?

In an earlier version of the PR the router logic in `_account.index.tsx` always redirected to `/plan` and then got bounced back if the wasn't actually a plan management URI configured. I changed it to check preemptively based on a verbal conversation that we had... I would happily revert if you confirm that is what you prefer?

hughns: I've gone ahead and reverted to the simpler version with no check at the query param handling level.

frontend/src/routes/_account.plan.index.tsx
hughns: I can't get the observer to fire after set up. I've tried with a ResizeObserver too. Testing on Chrome and Safari. I will try again tomorrow.

hughns: I found two problems:

1. The node corresponding to the iframe document body changed, so we weren't always observing the right thing. The fix was to re-observe after the iframe has loaded.
2. Some rendering changes do not correspond to a mutation on the iframe document body DOM. This is fixed by doing an extra calculation after a delay.

I tried using the resize observer again, but it still didn't fire as expected.

crates/config/src/sections/experimental.rs
sandhose: I'm indeed thinking of just passing it as as string without validation

frontend/src/routes/_account.plan.index.tsx
sandhose: > I changed it to check preemptively based on a verbal conversation that we had... I would happily revert if you confirm that is what you prefer?

Sorry I think there was a small misunderstanding here 😅 
I think I was thinking about using `throw redirect` in the loader of the plan route, so that it never tries to display the route

frontend/src/routes/_account.plan.index.tsx
```diff
@@ -0,0 +1,116 @@
...
+export const Route = createFileRoute("/_account/plan/")({
+  loader: async ({ context }) => {
+    const { planManagementIframeUri } = (
+      await context.queryClient.ensureQueryData(query)
+    ).siteConfig;
+    if (planManagementIframeUri) {
+      preload(planManagementIframeUri, { as: "document" });
+    } else {
+      throw redirect({ to: "/" });
+    }
```
sandhose: Style nitpick + using `replace` on the redirect

```suggestion
    const { siteConfig: { planManagementIframeUri } } =
      await context.queryClient.ensureQueryData(query);

    if (!planManagementIframeUri)
      throw redirect({ to: "/", replace: true });

    preload(planManagementIframeUri, { as: "document" });
```

sandhose (commented):
Grr, I had this not submitted since yesterday

crates/config/src/sections/experimental.rs
hughns: > I'm indeed thinking of just passing it as as string without validation

I've converted into a `String` with some additional doc comments in https://github.com/element-hq/matrix-authentication-service/commit/d6dd647f79258d2ff951a1d5752e37b141101cfd

PR #4563, opened 2025-05-13
Fix client_registration URI regex not accepting full query string grammar

t3chguy:
The grammar is pretty hellish and parsing a URL by using a regex is awful =(

https://www.cl.cam.ac.uk/~jac22/books/www/book/node166.html


sandhose (approved):
Tested locally and looks like it works well, thanks!

PR #4568, opened 2025-05-15
email login_hint support when login_with_email_allowed is activated

mcalinghee:
It is now possible to forward the login_hint with a mxid to the upstream providers thanks to https://github.com/element-hq/matrix-authentication-service/pull/4512

This PR allows us to support the login_hint with an email when `login_with_email_allowed` is activated.
In particular this PR will display the login_hint ie. email on the MAS login page.


mcalinghee:
Might be re-adapted if this PR goes through : https://github.com/element-hq/matrix-authentication-service/pull/4571

crates/data-model/src/oauth2/authorization_grant.rs
```diff
@@ -142,6 +144,7 @@ impl AuthorizationGrantStage {
 
 pub enum LoginHint<'a> {
     MXID(&'a UserId),
+    EMAIL(lettre::Address),
```
sandhose: In the case before, MXID is an ~acronym, hence the caps, but for emails it should follow the regular naming convention
```suggestion
    Email(lettre::Address),
```

crates/data-model/src/oauth2/authorization_grant.rs
```diff
@@ -197,6 +200,16 @@ impl AuthorizationGrant {
 
                 LoginHint::MXID(mxid)
             }
+            "email" => {
```
sandhose: I'm not convinced "email" should be a prefix. We've defined that in the MSC specifically for mxid:, but more generally, having emails as-is for login_hint is somewhat supported?

crates/data-model/Cargo.toml
```diff
@@ -28,3 +28,5 @@ ruma-common.workspace = true
 mas-iana.workspace = true
 mas-jose.workspace = true
 oauth2-types.workspace = true
+# Emails
+lettre.workspace = true
```
sandhose: Can you put that in the list of external crates up there without the comment for consistency?

crates/data-model/src/oauth2/authorization_grant.rs
```diff
@@ -172,14 +175,31 @@ impl std::ops::Deref for AuthorizationGrant {
...
+    ///
+    /// Returns `LoginHint::MXID` for valid mxid 'mxid:@john.doe:example.com'
+    ///
+    /// Returns `LoginHint::Email` for valid email 'john.doe@example.com' if
+    /// email supports is enabled
+    ///
+    /// Otherwise returns `LoginHint::None`
     #[must_use]
-    pub fn parse_login_hint(&self, homeserver: &str) -> LoginHint {
+    pub fn parse_login_hint(&self, homeserver: &str, login_with_email_allowed: bool) -> LoginHint {
```
sandhose: I think it makes more sense to have the upper level handle this logic: always parse the email login hint, and let the caller decide what to do with it

crates/data-model/src/oauth2/authorization_grant.rs
```diff
@@ -306,11 +326,47 @@ mod tests {
...
         assert!(matches!(hint, LoginHint::MXID(mxid) if mxid.localpart() == "example-user"));
     }
 
+    #[test]
+    fn valid_login_hint_with_email() {
+        #[allow(clippy::disallowed_methods)]
+        let mut rng = thread_rng();
+
+        #[allow(clippy::disallowed_methods)]
+        let now = Utc::now();
```
sandhose: Please use a static RNG and the MockClock in tests instead

```suggestion
        let now = MockClock::default().now();
        let mut rng = rand_chacha::ChaChaRng::seed_from_u64(42);
```

There is a reason for those methods are not allowed: we want tests to have a stable outcome, so that we can use snapshots when needed

crates/handlers/src/views/login.rs
```diff
@@ -380,8 +387,12 @@ fn handle_login_hint(
...
 
     if let PostAuthContextInner::ContinueAuthorizationGrant { ref grant } = next.ctx {
-        let value = match grant.parse_login_hint(homeserver.homeserver()) {
+        let value = match grant.parse_login_hint(
+            homeserver.homeserver(),
+            site_config.login_with_email_allowed,
+        ) {
             LoginHint::MXID(mxid) => Some(mxid.localpart().to_owned()),
+            LoginHint::Email(email) => Some(email.to_string()),
             LoginHint::None => None,
```
sandhose: the logic would then be something like

```rust
let value = match grant.parse_login_hint(homeserver.homeserver()) {
    LoginHint::MXID(mxid) => Some(mxid.localpart().to_owned()),
    LoginHint::Email(email) if site_config.login_with_email_allowed => Some(email.to_string()),
    _ => None,
};
```

crates/data-model/src/oauth2/authorization_grant.rs
mcalinghee: agreed but I would need to add a dependency on `mas-storage` on the `data-model` crate and this would end up as a cyclic dependency.
If this is ok, I would extract these static function into `data-model` to be able to use them in both `mas-storage` and `data-model` but there will be a bit of code refactor in this PR

sandhose:
@mcalinghee after an initial review, I would appreciate if you don't force-push to the branch, as it often makes GitHub confused and makes it harder to understand what changed. If you ever need to update from the main branch, prefer merging it into your branch instead of rebasing :)

crates/data-model/src/oauth2/authorization_grant.rs
sandhose: Hah, of course! If you feel like doing it, you can move `Clock`/`MockClock`/`SystemClock`/`BoxClock`/`BoxRng` to `mas-data-model`, either in a separate PR or in this one

I'm fine with keeping it like that if it's too annoying to refactor now

mcalinghee: Ok job done.

It was a bit annoying but I have done most of the refactoring in c314802fcdf6db6cd2108af9e9aa6d87dc084c81 and brought minor corrections afterwards. I hope this is ok.

Hopefully, it will not be too painful to perform the review.
Let me know if you prefer to have this in another PR as the refacto is quite heavy.

crates/data-model/src/lib.rs
```diff
@@ -4,10 +4,19 @@
 // SPDX-License-Identifier: AGPL-3.0-only OR LicenseRef-Element-Commercial
 // Please see LICENSE files in the repository root for full details.
 
+//!
+//! This crate defines a [`Clock`] trait that can be used to abstract the
+//! way the current time is retrieved. It has two implementation:
+//! [`SystemClock`] that uses the system time and [`MockClock`] which is useful
+//! for testing.
+//!
+//! [`MockClock`]: crate::clock::MockClock
```
sandhose: This comment feels a bit out of place as there are no other documentation for the crate. It is redundant with the module comment in the `clock` module, so I'm fine with dropping it

crates/data-model/src/oauth2/authorization_grant.rs
```diff
@@ -172,15 +175,26 @@ impl std::ops::Deref for AuthorizationGrant {
...
+    ///
+    /// Otherwise returns `LoginHint::None`
     #[must_use]
     pub fn parse_login_hint(&self, homeserver: &str) -> LoginHint<'_> {
         let Some(login_hint) = &self.login_hint else {
             return LoginHint::None;
         };
 
-        // Return none if the format is incorrect
         let Some((prefix, value)) = login_hint.split_once(':') else {
```
sandhose: I've simplified the whole logic a little bit in fb5c4de88

mcalinghee: This makes more sense indeed.
Thank you for this.

PR #4571, opened 2025-05-15
remove prefix when forwarding the login_hint

mcalinghee:
It is now possible to forward the login_hint with a mxid to the upstream providers thanks to https://github.com/element-hq/matrix-authentication-service/pull/4512

We are using a prefix `mxid` in the login_hint (login_hint is in the form of `mxid:@localpart:homeserver`) to know the type of the login_hint.
This prefix is useful between the MAS and their clients but on the upstream providers side, it is usually not supported unless we customize them.


mcalinghee:
We can change this PR to add a configuration to `forward_login_hint_with_prefix` or not

sandhose:
I feel like this is now covered by https://github.com/element-hq/matrix-authentication-service/pull/5677 ; closing

PR #4576, opened 2025-05-19
Add missing `synapse_idp_id` to docs

jaywink:
This is documented in the config schema but confusingly not visible in the docs, when it is an integral part of running `syn2mas` for a Synapse with upstream OIDC providers.

Added also a short mention in the `syn2mas` page for clarity.


mcalinghee (commented):
Do you think it is a good idea to add reference to the [migration documentation](https://element-hq.github.io/matrix-authentication-service/setup/migration.html#map-any-upstream-sso-providers) (../setup/migration.md#map-any-upstream-sso-providers)? 
This is only place where `synapse_idp_id` has been documented.

sandhose:
Sorry I missed that ; I've covered that in another PR, closing that one in favor of #5748

PR #4602, opened 2025-06-02
config: Refactor parsing of secrets section

V02460:
Code refactor of the config’s `secrets` section. Can be reviewed commit-by-commit.

### Benefits

- Makes some invalid states unrepresentable
- Eliminates now-unnecessary checks in the `validate` function
- Adds concurrent loading of key files
- Creates a cleaner separation between deserialization and creation of key store

### Noteworthy changes

- This PR unifies the loading of keys from the config and file by no longer assuming a specific key format in either case.
- The `futures` crate is added as a dependency to have access to the standard async tools `try_join` and `try_join_all`.

### DIFF
--- a/crates/config/Cargo.toml
+++ b/crates/config/Cargo.toml
@@ -15,7 +15,7 @@ workspace = true
 tokio.workspace = true
 tracing.workspace = true
 anyhow.workspace = true
-futures = "0.3.31"
+futures-util.workspace = true
 
 camino = { workspace = true, features = ["serde1"] }
 chrono.workspace = true
--- a/crates/config/Cargo.toml
+++ b/crates/config/Cargo.toml
@@ -15,6 +15,7 @@ workspace = true
 tokio.workspace = true
 tracing.workspace = true
 anyhow.workspace = true
+futures = "0.3.31"
 
 camino = { workspace = true, features = ["serde1"] }
 chrono.workspace = true
--- a/crates/config/src/sections/secrets.rs
+++ b/crates/config/src/sections/secrets.rs
@@ -42,7 +42,9 @@ pub enum Password {
 #[derive(JsonSchema, Serialize, Deserialize, Clone, Debug)]
 struct PasswordRaw {
     #[schemars(with = "Option<String>")]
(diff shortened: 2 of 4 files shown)


crates/config/Cargo.toml
```diff
@@ -15,6 +15,7 @@ workspace = true
 tokio.workspace = true
 tracing.workspace = true
 anyhow.workspace = true
+futures = "0.3.31"
```
sandhose: Please use `futures-util` instead, it should provide the utility you need
```suggestion
futures-util.workspace = true
```

crates/config/src/sections/secrets.rs
```diff
@@ -27,23 +28,156 @@ fn example_secret() -> &'static str {
...
+
+/// A single key with its key ID and optional password.
+#[serde_as]
+#[derive(JsonSchema, Serialize, Deserialize, Clone, Debug)]
+pub struct KeyConfig {
+    kid: String,
+
+    #[schemars(with = "PasswordRaw")]
+    #[serde_as(as = "serde_with::TryFromInto<PasswordRaw>")]
+    #[serde(flatten)]
```
sandhose, line 139: This is really neat! Main reason I switched away from `serde(flatten)` was because the generated scheme was crap :)

crates/config/src/sections/secrets.rs
```diff
@@ -127,40 +221,7 @@ impl SecretsConfig {
...
-            if key.password.is_some() && key.password_file.is_some() {
-                return annotate(figment::Error::from(
-                    "Cannot specify both `password` and `password_file`".to_owned(),
-                ));
-            }
-        }
-
+    fn validate(&self, _figment: &figment::Figment) -> Result<(), figment::Error> {
         Ok(())
     }
```
sandhose: You can remove that, there is a default implementation for this trait

crates/config/src/sections/secrets.rs
V02460: Glad to hear you like it! 😊

PR #4604, opened 2025-06-02
Reduce the cache time for 404s on the assets

sandhose:


crates/cli/src/server.rs
```diff
@@ -246,17 +243,28 @@ pub fn build_router(
...
-                let cache_layer = SetResponseHeaderLayer::overriding(
-                    CACHE_CONTROL,
-                    HeaderValue::from_static("public, max-age=31536000, immutable"),
+                let add_cache_headers = axum::middleware::map_response(
+                    async |mut res: Response<ServeFileSystemResponseBody>| {
+                        let cache_control = if res.status() == StatusCode::NOT_FOUND {
+                            // Cache 404s for 5 minutes
+                            CacheControl::new()
+                                .with_public()
+                                .with_max_age(Duration::from_secs(5 * 60))
```
reivilibre, line 252: honestly 5 min seems excessive, I would have gone for 30 seconds, maybe even less

but otherwise fine

reivilibre (approved):

PR #4617, opened 2025-06-02
Add secrets.encryption_file config option

V02460:
Adds the `secrets.encryption_file` config option. This allows users to pass the encryption secret via file. Includes documentation.

### DIFF
--- a/docs/config.schema.json
+++ b/docs/config.schema.json
@@ -1515,18 +1515,7 @@
     "SecretsConfig": {
       "description": "Application secrets",
       "type": "object",
-      "required": [
-        "encryption"
-      ],
       "properties": {
-        "encryption": {
-          "description": "Encryption key for secure cookies",
-          "examples": [
-            "0000111122223333444455556666777788889999aaaabbbbccccddddeeeeffff"
-          ],
-          "type": "string",
-          "pattern": "[0-9a-fA-F]{64}"
-        },
         "keys": {
           "description": "List of private keys to use for signing and encrypting payloads",
           "default": [],
@@ -1534,6 +1523,19 @@
           "items": {
             "$ref": "#/definitions/KeyConfig"
           }
+        },
+        "encryption_file": {
+          "description": "File containing the encryption key for secure cookies.",
+          "type": "string"
+        },
+        "encryption": {
+          "description": "Encryption key for secure cookies.",
+          "default": null,
+          "examples": [
+            "0000111122223333444455556666777788889999aaaabbbbccccddddeeeeffff"
+          ],
+          "type": "string",
+          "pattern": "[0-9a-fA-F]{64}"
         }
       }
     },
--- a/docs/config.schema.json
+++ b/docs/config.schema.json
@@ -1530,7 +1530,6 @@
         },
         "encryption": {
           "description": "Encryption key for secure cookies.",
-          "default": null,
           "examples": [
             "0000111122223333444455556666777788889999aaaabbbbccccddddeeeeffff"
           ],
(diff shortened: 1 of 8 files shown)


docs/config.schema.json
```diff
@@ -1515,25 +1515,27 @@
...
             "$ref": "#/definitions/KeyConfig"
           }
+        },
+        "encryption_file": {
+          "description": "File containing the encryption key for secure cookies.",
+          "type": "string"
+        },
+        "encryption": {
+          "description": "Encryption key for secure cookies.",
+          "default": null,
```
V02460: I was not able to get rid of this default keyword and have no idea where it’s coming from. It doesn’t even validate!

docs/reference/configuration.md
```diff
@@ -232,6 +232,21 @@ secrets:
...
+via file, set `secrets.encryption_file` to the file path; alternatively use
+`secrets.encryption` for declaring the secret inline. The options
+`secrets.encryption_file` and `secrets.encryption` are mutually exclusive.
+
+If given via file, the encyption secret is only read at application startup.
+The secret is not updated when the content of the file changes.
+
+> ⚠️ **Warning** – Do not change the encryption secret after the initial start.
+> Changing the encryption secret afterwards will lead to a loss of all
+> information stored in the database.
```
V02460: I am not very sure about this warning. Can you please double-check it’s correct?

docs/config.schema.json
sandhose: I wonder if this would go away if you do `#[serde(skip_serializing_if = "Option::is_none")]`?

docs/reference/configuration.md
sandhose: Not all, but some things will not behave correctly. The most important one is that all cookies sessions will get invalidated

docs/reference/configuration.md
```diff
@@ -232,6 +232,21 @@ secrets:
         -----END EC PRIVATE KEY-----
 ```
 
+### `secrets.encryption{_file}`
+
+The encryption secret used for encrypting cookies and database fields. It takes
+the form of a 32-bytes-long hex-encoded string. To provide the encryption secret
```
sandhose, line 238: Hex-encoded is right for the string variant, not for the file variant though? Could you clarify that? Also it would be nice to have instructions on how to generate such file, e.g. `openssl rand 32 -out file`

sandhose (changes requested):

docs/reference/configuration.md
V02460: I changed the code to match the documentation by expecting hex characters in the file. This better matches behavior of similar features and makes file generation trivial.

docs/config.schema.json
V02460: That did the trick – and the annotation was missing anyway :+1:

docs/reference/configuration.md
V02460: So is changing the secret like *bad* bad, never do it, things will break? Or more like: One can choose to perform secret rotation every other month and accept that users will be slightly inconvenienced, because they have to log in again?

sandhose: Yeah no, it's *bad bad*, don't do it :)

We don't encrypt much in the database now (except client secrets), but I don't want to have to deal with users regularly rotating this one if we start introduce encryption for other more crucial fields

sandhose:
@V02460 whenever you've finished fixing stuff on your PRs, please re-request review so that I know it's ready for me to have another look :)

docs/reference/configuration.md
V02460: Ok, I rewrote the warning to slightly increase accuracy and scariness ;)

PR #4619, opened 2025-06-03
Registration tokens

### LINKED ISSUE #3956: Moderated registration
Currently running an open matrix server with "just" email and CAPTCHA still attracts a lot of spammers, as they can send messages into rooms directly after signup.

I'd really like to see a moderated signup option. This would be similar to how Mastodon allows the users to put in a reason for the account and then the moderators to review and manually approve each account.

An alternative would be to implement / configure this in an upstream OAuth provider and only use that in MAS. But this would compliate social login and migration from existing non-MAS to MAS scenarios.

**Additional context**
In non-MAS times this was done via eg. https://github.com/krombel/matrix-register-bot or https://github.com/ZerataX/matrix-registration to limit spam registrations

### LINKED ISSUE #4584: Registration tokens

Registration tokens (https://element-hq.github.io/synapse/latest/usage/administration/admin_api/registration_tokens.html#registration-tokens) in Synapse are used by both our customers and community servers to mitigate abusive account registrations.

Implement registration tokens in MAS.

Other than not using MAS and by extension not use Element X, no good alternatives are available. Email allow list is not a practical replacement for community instances.

sandhose:
Fixes #3956 
Fixes #4584 

This can be reviewed commit by commit.

It includes an admin API to list/get/create/revoke tokens, as well as a CLI tool to issue a token.

- **Data model and repository for user registration tokens**
- **Add config flag to require registration tokens for password registrations**
- **Registration token step view**
- **Enforce registration token on registration**
- **List and count methods on the UserRegistrationTokenRepository**
- **Admin API to list user registration tokens**
- **Admin API to get individual user registration tokens**
- **Admin API to create a new user registration token**
- **Admin API to revoke user registration tokens**
- **CLI tool to issue user registration tokens**

### DIFF
--- a/crates/cli/src/commands/manage.rs
+++ b/crates/cli/src/commands/manage.rs
@@ -7,6 +7,7 @@
 use std::{collections::BTreeMap, process::ExitCode};
 
 use anyhow::Context;
+use chrono::Duration;
 use clap::{ArgAction, CommandFactory, Parser};
 use console::{Alignment, Style, Term, pad_str, style};
 use dialoguer::{Confirm, FuzzySelect, Input, Password, theme::ColorfulTheme};
@@ -28,7 +29,10 @@ use mas_storage::{
     user::{BrowserSessionFilter, UserEmailRepository, UserPasswordRepository, UserRepository},
 };
 use mas_storage_pg::{DatabaseError, PgRepository};
-use rand::{RngCore, SeedableRng};
+use rand::{
+    RngCore, SeedableRng,
+    distributions::{Alphanumeric, DistString as _},
+};
 use sqlx::{Acquire, types::Uuid};
 use tracing::{error, info, info_span, warn};
 use zeroize::Zeroizing;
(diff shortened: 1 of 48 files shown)


crates/cli/src/commands/manage.rs
```diff
@@ -95,6 +99,24 @@ enum Subcommand {
...
 
+    /// Create a new user registration token
+    IssueUserRegistrationToken {
+        /// Specific token string to use. If not provided, a random token will
+        /// be generated.
+        #[arg(long)]
+        token: Option<String>,
+
+        /// Maximum number of times this token can be used.
+        /// If not provided, the token can be used an unlimited number of times.
```
reivilibre: hmm.. kind of feels like the sensible default would be 1, to be honest

(the word 'token' feels like something you 'spend', not something that would be infinite by default)

reivilibre (approved):

crates/cli/src/commands/manage.rs
sandhose: Fair enough. I've changed the behaviour of the CLI tool in e53e2666b

olmari:
Very +1 to this concept. After MAS we have basically lost the "not SSO" registation method of https://github.com/zeratax/matrix-registration/ we used to use, as we have 12+ different hackerspaces to cater and not everyone have an SSO available for an reason or another, so _a_ method of registation tokens would be awesome to support while MAS is at use.

To say it most broadly, that Zeratax thing would still be just perfect as concept, but obviously it would need complete backend overhaul to use MAS instead.

PR #4682, opened 2025-06-12
Upgrade schemars to 0.9

sandhose:
This upgrades schemars and aide to their latest versions.

This changes a little bit the generated config and adminapi schemas, but at a first glance nothing major.


crates/handlers/src/admin/mod.rs
```diff
@@ -171,8 +172,16 @@ where
     aide::generate::infer_responses(false);
 
     aide::generate::in_context(|ctx| {
-        ctx.schema =
-            schemars::r#gen::SchemaGenerator::new(schemars::r#gen::SchemaSettings::openapi3());
+        ctx.schema = schemars::generate::SchemaGenerator::new(
+            schemars::generate::SchemaSettings::openapi3().with(|settings| {
+                // Remove the transform which adds nullable fields, as it's not
+                // valid with OpenAPI 3.1. For some reason, aide/schemars output
+                // an OpenAPI 3.1 schema with this nullable transform.
```
reivilibre, line 179: should this be reported & issue tracked here?

reivilibre (approved):
not too bad actually

crates/handlers/src/admin/mod.rs
sandhose: I don't know if it's an `aide` or a `schemars` thing… I can't really find related issues, but also there is a bunch of cleaning up happening on the `aide` side happening on the main branch; maybe we should just wait for the next release and see?

PR #4688, opened 2025-06-16
Add logging to `/login` failure

erikjohnston:
Two commits:

1. Simple logging for which error case we're hitting
2. Refactor of password verification result so that we can differentiate between "password mismatch" vs. other errors.

The second has ended up being a fairly large change, as I opted to add a new struct `PasswordVerificationResult` for the return type. I did look instead into making the returned error being structured, but that proved a bit of a PITA, and generally I think we want to handle the error vs password mismatch cases differently anyway.

### DIFF
--- a/crates/data-model/src/users.rs
+++ b/crates/data-model/src/users.rs
@@ -4,7 +4,7 @@
 // SPDX-License-Identifier: AGPL-3.0-only OR LicenseRef-Element-Commercial
 // Please see LICENSE files in the repository root for full details.
 
-use std::{fmt::Display, net::IpAddr};
+use std::net::IpAddr;
 
 use chrono::{DateTime, Utc};
 use rand::Rng;
--- a/crates/data-model/src/users.rs
+++ b/crates/data-model/src/users.rs
@@ -47,14 +47,6 @@ impl User {
     }
 }
 
-impl Display for User {
-    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
-        // Format the user as their username, this is useful for logging and
-        // debugging.
-        self.username.fmt(f)
-    }
-}
-
 #[derive(Debug, Clone, PartialEq, Eq, Serialize)]
 pub struct Password {
     pub id: Ulid,
--- a/crates/data-model/src/users.rs
+++ b/crates/data-model/src/users.rs
@@ -4,7 +4,7 @@
 // SPDX-License-Identifier: AGPL-3.0-only OR LicenseRef-Element-Commercial
 // Please see LICENSE files in the repository root for full details.
 
-use std::net::IpAddr;
+use std::{fmt::Display, net::IpAddr};
 
 use chrono::{DateTime, Utc};
 use rand::Rng;
@@ -44,6 +44,14 @@ impl User {
             deactivated_at: None,
             can_request_admin: false,
         }]
+    }
+}
+
+impl Display for User {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        // Format the user as their username, this is useful for logging and
+        // debugging.
+        self.username.fmt(f)
     }
 }
 
--- a/crates/handlers/src/compat/login.rs
+++ b/crates/handlers/src/compat/login.rs
@@ -32,7 +32,8 @@ use zeroize::Zeroizing;
 use super::{MatrixError, MatrixJsonBody};
 use crate::{
     BoundActivityTracker, Limiter, METER, RequesterFingerprint, impl_from_error_for_route,
-    passwords::PasswordManager, rate_limit::PasswordCheckLimitedError,
+    passwords::{PasswordManager, PasswordVerificationResult},
+    rate_limit::PasswordCheckLimitedError,
 };
 
 static LOGIN_COUNTER: LazyLock<Counter<u64>> = LazyLock::new(|| {
@@ -193,7 +194,7 @@ pub enum RouteError {
     NoPassword,
 
     #[error("password verification failed")]
-    PasswordVerificationFailed(#[source] anyhow::Error),
+    PasswordVerificationFailed,
 
     #[error("request rate limited")]
     RateLimited(#[from] PasswordCheckLimitedError),
@@ -209,6 +210,12 @@ pub enum RouteError {
 }
 
 impl_from_error_for_route!(mas_storage::RepositoryError);
+
+impl From<anyhow::Error> for RouteError {
+    fn from(err: anyhow::Error) -> Self {
+        Self::Internal(err.into())
+    }
+}
 
 impl IntoResponse for RouteError {
     fn into_response(self) -> axum::response::Response {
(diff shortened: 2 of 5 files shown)


crates/handlers/src/passwords.rs
```diff
@@ -345,16 +389,16 @@ impl Algorithm {
         hashed_password: &str,
         password: &[u8],
         pepper: Option<&[u8]>,
-    ) -> Result<(), anyhow::Error> {
-        match self {
+    ) -> Result<PasswordVerificationResult, anyhow::Error> {
```
sandhose, line 392: This should be simplified with just a boolean

crates/handlers/src/views/login.rs
```diff
@@ -166,6 +166,7 @@ pub(crate) async fn post(
     }
 
     if !form_state.is_valid() {
+        tracing::warn!("Invalid login form: {form_state:?}");
         PASSWORD_LOGIN_COUNTER.add(1, &[KeyValue::new(RESULT, "error")]);
```
sandhose, line 170: I think it makes sense to add an additional label on the metric to attach the reason on each failure, which will be easier to look in graphs rather than logs

crates/handlers/src/views/login.rs
```diff
@@ -189,6 +190,7 @@ pub(crate) async fn post(
     // First, lookup the user
     let Some(user) = get_user_by_email_or_by_username(site_config, &mut repo, username).await?
     else {
+        tracing::warn!("User not found: {username}");
```
sandhose: We usually prefer to log attributes structured
```suggestion
        tracing::warn!(%username, "User not found");
```

crates/handlers/src/views/login.rs
```diff
@@ -269,10 +272,11 @@ pub(crate) async fn post(
                 )
                 .await?
         }
-        Ok(None) => user_password,
-        Err(_) => {
+        Ok(PasswordVerificationResult::Success(None)) => user_password,
+        Ok(PasswordVerificationResult::Failure) => {
+            tracing::warn!("Failed to verify/upgrade password for user: {user}");
```
sandhose:
```suggestion
            tracing::warn!(%user.id, "Failed to verify/upgrade password for user");
```

or `%user.username`

crates/data-model/src/users.rs
```diff
@@ -47,6 +47,14 @@ impl User {
     }
 }
 
+impl Display for User {
```
sandhose: I don't like the idea of adding a display implementation for things in datamodel, unless the whole thing has an obvious string representation

crates/handlers/src/compat/login.rs
```diff
@@ -193,7 +194,7 @@ pub enum RouteError {
     NoPassword,
 
     #[error("password verification failed")]
-    PasswordVerificationFailed(#[source] anyhow::Error),
+    PasswordVerificationFailed,
```
sandhose: To be consistent with the rest, split this into `PasswordMismatch` and `PasswordVerificationFailed(#[source] anyhow::Error)`, and mark the second as an internal server error like we do for `ProvisionDeviceFailed`

crates/handlers/src/passwords.rs
```diff
@@ -186,7 +230,7 @@ impl PasswordManager {
         scheme: SchemeVersion,
         password: Zeroizing<String>,
         hashed_password: String,
-    ) -> Result<Option<(SchemeVersion, String)>, anyhow::Error> {
+    ) -> Result<PasswordVerificationResult<Option<(SchemeVersion, String)>>, anyhow::Error> {
```
sandhose, line 233: Not the prettiest, but this could return `Result<bool, Option<(SchemaVersion, String)>>`. I don't think adding this new result type makes sense

crates/handlers/src/views/login.rs
```diff
@@ -228,6 +230,7 @@ pub(crate) async fn post(
     let Some(user_password) = repo.user_password().active(&user).await? else {
         // There is no password for this user, but we don't want to disclose that. Show
         // a generic 'invalid credentials' error instead
+        tracing::warn!("No password for user: {user}");
```
sandhose:
```suggestion
        tracing::warn!(%user.id, "No password for user");
```

crates/handlers/src/views/login.rs
```diff
@@ -286,11 +290,13 @@ pub(crate) async fn post(
             )
             .await;
         }
+        Err(err) => return Err(InternalError::from_anyhow(err)),
     };
 
     // Now that we have checked the user password, we now want to show an error if
     // the user is locked or deactivated
     if user.deactivated_at.is_some() {
+        tracing::warn!("User is deactivated: {user}");
```
sandhose:
```suggestion
        tracing::warn!(%user.id, "User is deactivated");
```

crates/handlers/src/views/login.rs
```diff
@@ -301,6 +307,7 @@ pub(crate) async fn post(
     }
 
     if user.locked_at.is_some() {
+        tracing::warn!("User is locked: {user}");
```
sandhose:
```suggestion
        tracing::warn!(%user.id, "User is locked");
```

sandhose (changes requested):

crates/handlers/src/passwords.rs
erikjohnston: The reason for using `PasswordVerificationResult` is so that we can use `#[must_use]`, which a) helped a lot in the refactor to ensure we changed all call sites, and b) ensures future uses don't forget to actually check the result (this is a common issue with verify functions returning bools).

crates/handlers/src/views/login.rs
erikjohnston: Heh, after I keep making the same point in all the synapse PRs

crates/data-model/src/users.rs
erikjohnston: Mmm, have removed for now. Although I do suspect we'll run into this later where we want to ergonomically log the user, so not sure what the best plan there is

crates/handlers/src/compat/login.rs
erikjohnston: Have renamed. Errors raised during verification are currently going to `InternalError`, is it helpful to split that up into a separate variant? I wouldn't have thought so as they're unexpected internal errors at that point.

crates/handlers/src/views/login.rs
erikjohnston: I'm not sure. It'd only be useful if we can capture enough information in the metrics to be able to debug it, but we also can't add much cardinality to the metrics without causing issues for Prometheus.

TBH, so long as we actually have what is going wrong in the logs then its probably easy enough to just check in there (or in tracing, etc) :shrug:

crates/handlers/src/passwords.rs
erikjohnston: We could potentially change it to be `type PasswordVerificationResult<T> = Result<T, ()>` too, but that just might be even more confusing to use.

sandhose (approved):

sandhose:
Sorry that I've missed the re-review request

PR #4743, opened 2025-07-03
WIP: backchannel logout

sandhose:
- **Record the decoded ID token claims on upstream auth sessions**
- **storage: list and count methods for upstream oauth sessions**
- **Allow filtering upstream sessions by sub and sid claims**
- **Backchannel logout behavior settings on upstream providers**
- **fixup! Record the decoded ID token claims on upstream auth sessions**
- **Receive and validate backchannel logout requests**
- **storage: allow filtering browser sessions by which upstream session authd them**
- **fixup! Backchannel logout behavior settings on upstream providers**
- **Log out browser sessions when receiving a backchannel logout notification**


reivilibre (approved):
The semantics/opinions on desirable behaviour themselves seem messy, but as an implementation of what it says it will implement, this seems correct and reasonable

PR #4744, opened 2025-07-03
Allow skipping GDPR-erasure when deactivating a user through the admin API

### LINKED ISSUE #4742: REST Admin API to deactivate user without erasing it
The GraphQL API allows choosing whether or not to erase a deactivated user ([see here](https://github.com/element-hq/matrix-authentication-service/blob/b4c3de411c44956a90387471a250ac1aa58010f6/frontend/schema.graphql#L611)), but the REST Admin API's [`/deactivate` endpoint](https://element-hq.github.io/matrix-authentication-service/api/index.html#/user/deactivateUser) does not & always erases the user ([see here](https://github.com/element-hq/matrix-authentication-service/blob/b4c3de411c44956a90387471a250ac1aa58010f6/crates/handlers/src/admin/v1/users/deactivate.rs#L93)).

The REST Admin API's `/deactivate` endpoint should accept a JSON body with an `erase` boolean property, like the Synapse Admin API does.  For backwards compatibility, the endpoint should continue to accept requests with no body.

There could be a separate endpoint for deactivating without erasing, instead of adding a request body to the existing endpoint.

**Additional context**
Having the option to not erase deactivated users would help achieve parity with the Synapse Admin API.

AndrewFerr:
Fixes #4742 

This allows using the endpoint to deactivate a user without GDPR-erasing it


AndrewFerr:
The failure is due to the schema updater marking the request body of `/deactivate` as required, but I want it to be optional.  The unwanted change is:
```
diff --git a/docs/api/spec.json b/docs/api/spec.json
index cb597e14..c0e2ee51 100644
--- a/docs/api/spec.json
+++ b/docs/api/spec.json
@@ -1367,7 +1367,7 @@
               }
             }
           },
-          "required": false
+          "required": true
         },
         "responses": {
           "200": {
```

In `deactivate.rs`, I've already allowed the handler to accept an optional body (with a parameter of `body: Option<Json<Request>>`), so in practice the body really is optional (and a unit test confirms that).

Is there a way to make schema generation aware of the body being optional?

The reason I want it to be optional is to keep the endpoint backwards-compatible with its current form, which expects no body.

crates/handlers/src/admin/v1/users/deactivate.rs
```diff
@@ -138,6 +173,17 @@ mod tests {
...
+        // XXX: we don't have a good way to look for the deactivation job
+        let job: Json<serde_json::Value> = sqlx::query_scalar(
+            "SELECT payload FROM queue_jobs WHERE queue_name = 'deactivate-user'",
+        )
+        .fetch_one(&pool)
+        .await
+        .expect("Deactivation job to be scheduled");
+        assert_eq!(job["user_id"], serde_json::json!(user.id));
+        assert_eq!(job["hs_erase"], serde_json::json!(erase.unwrap_or(true)));
+
```
AndrewFerr: I kept this in to check for the erase parameter on the job, as the response to `GET /users` doesn't indicate whether the user was erased or not.

PR #4745, opened 2025-07-04
Add the JWT being validated to the log message

sandhose:


sandhose:
This was useful for debugging at some point but we don't need that anymore, closing

Issue #4740, opened 2025-07-03
REST Admin API to reactivate user


Twi1ightSparkle:
**Is your feature request related to a problem? Please describe.**

Account reactivation was implemented in https://github.com/matrix-org/matrix-authentication-service/pull/2970 but only in the GraphQL API

**Describe the solution you'd like**

Implement a REST Admin API to reactivate a user

**Describe alternatives you've considered**

No permanent solution since the GraphQL API will be deprecated at some point

**Additional context**

User reactivation is used frequently by enterprise customers

PR #4768, opened 2025-07-09
New admin API endpoint to reactivate users & align what lock/unlock/deactivate does in the admin API

AndrewFerr:
Fixes https://github.com/element-hq/matrix-authentication-service/issues/4740

This changes the lock/unlock/deactivate endpoints to:

 - only do what they claim to do (unlock doesn't reactivate, deactivate doesn't lock)
 - do most of it right away, not in a background task

### DIFF
--- a/crates/handlers/src/graphql/mutations/user.rs
+++ b/crates/handlers/src/graphql/mutations/user.rs
@@ -144,9 +144,6 @@ impl LockUserPayload {
 struct UnlockUserInput {
     /// The ID of the user to unlock
     user_id: ID,
-
-    /// Reactivate the user if it had been deactivated
-    reactivate: Option<bool>,
 }
 
 /// The status of the `unlockUser` mutation.
@@ -566,7 +563,7 @@ impl UserMutations {
         Ok(LockUserPayload::Locked(user))
     }
 
-    /// Unlock a user. This is only available to administrators.
+    /// Unlock and reactivate a user. This is only available to administrators.
     async fn unlock_user(
         &self,
         ctx: &Context<'_>,
@@ -588,18 +585,12 @@ impl UserMutations {
             return Ok(UnlockUserPayload::NotFound);
         };
 
-        let user = if input.reactivate.unwrap_or(false) {
-            // Call the homeserver synchronously to reactivate the user
-            let mxid = matrix.mxid(&user.username);
-            matrix.reactivate_user(&mxid).await?;
+        // Call the homeserver synchronously to reactivate the user
+        let mxid = matrix.mxid(&user.username);
+        matrix.reactivate_user(&mxid).await?;
 
-            // Now reactivate the user in our database
-            repo.user().reactivate(user).await?
-        } else {
-            user
-        };
-
-        // Now unlock the user in our database
+        // Now reactivate & unlock the user in our database
+        let user = repo.user().reactivate(user).await?;
         let user = repo.user().unlock(user).await?;
 
         repo.save().await?;
--- a/frontend/schema.graphql
+++ b/frontend/schema.graphql
@@ -886,7 +886,7 @@ type Mutation {
   """
   lockUser(input: LockUserInput!): LockUserPayload!
   """
-  Unlock a user. This is only available to administrators.
+  Unlock and reactivate a user. This is only available to administrators.
   """
   unlockUser(input: UnlockUserInput!): UnlockUserPayload!
   """
@@ -1842,10 +1842,6 @@ input UnlockUserInput {
   The ID of the user to unlock
   """
   userId: ID!
-  """
-  Reactivate the user if it had been deactivated
-  """
-  reactivate: Boolean
 }
 
 """
--- a/frontend/src/gql/graphql.ts
+++ b/frontend/src/gql/graphql.ts
@@ -604,7 +604,7 @@ export type Mutation = {
   setPrimaryEmail: SetPrimaryEmailPayload;
   /** Start a new email authentication flow */
   startEmailAuthentication: StartEmailAuthenticationPayload;
-  /** Unlock a user. This is only available to administrators. */
+  /** Unlock and reactivate a user. This is only available to administrators. */
   unlockUser: UnlockUserPayload;
 };
 
@@ -1347,8 +1347,6 @@ export type StartEmailAuthenticationStatus =
 
 /** The input for the `unlockUser` mutation. */
 export type UnlockUserInput = {
-  /** Reactivate the user if it had been deactivated */
-  reactivate?: InputMaybe<Scalars['Boolean']['input']>;
   /** The ID of the user to unlock */
   userId: Scalars['ID']['input'];
 };


AndrewFerr:
Marking this as draft until it makes use of https://github.com/element-hq/matrix-authentication-service/pull/4775 to test the updating of the `deactivated_at` timestamp on {d,r}eactivation.

crates/storage-pg/src/user/mod.rs
```diff
@@ -417,6 +417,40 @@ impl UserRepository for PgUserRepository<'_> {
...
+    #[tracing::instrument(
+        name = "db.user.reactivate_and_unlock",
+        skip_all,
+        fields(
+            db.query.text,
+            %user.id,
+        ),
+        err,
+    )]
+    async fn reactivate_and_unlock(&mut self, mut user: User) -> Result<User, Self::Error> {
```
sandhose: Considering we're doing things in a single transaction anyway, I wouldn't be worried with minimising the number of database roundtrips, and have consumers call `reactivate` and `unlock` together instead of adding a specialised method on the repository

crates/handlers/src/admin/v1/users/unlock.rs
```diff
@@ -50,7 +52,25 @@ impl IntoResponse for RouteError {
...
 }
 
-pub fn doc(operation: TransformOperation) -> TransformOperation {
+/// # JSON payload for the `POST /api/admin/v1/users/:id/unlock` endpoint
+#[derive(Default, Deserialize, JsonSchema)]
+#[serde(rename = "UnlockUserRequest")]
+pub struct Request {
+    /// Whether to skip ensuring the user is active upon being unlocked.
+    #[serde(default)]
+    skip_reactivate: bool,
```
sandhose: I'd prefer for each endpoint to do one thing and explicitly state in the docs 'this endpoints unset the 'locked' state of the user, but doesn't reactivate it. Call /reactivate to do so'

crates/handlers/src/admin/v1/users/unlock.rs
```diff
@@ -73,23 +93,27 @@ pub async fn handler(
...
-
-    // Now unlock the user in our database
-    let user = repo.user().unlock(user).await?;
+    let user = if !params.skip_reactivate {
+        // Call the homeserver synchronously to reactivate the user
+        let mxid = homeserver.mxid(&user.username);
+        homeserver
+            .reactivate_user(&mxid)
+            .await
+            .map_err(RouteError::Homeserver)?;
```
sandhose: I'm guessing that you're trying to not break the previous behaviour?

I think the reason this would reactivate is that previously we didn't track whether the user was deactivated or not on the MAS side, so you maaaay end up with users that are deactivated on the Synapse side but not on the MAS side.

I think there are two ways out of this:

 - allow calling /reactivate even if the user is not deactivated (I think this is already the case?) and expect clients to handle this weird edge-case
 - when you /unlock, reactivate on the HS if it's not deactivated on the MAS side to make sure it's not in an inconsistent state

Whatever way you're choosing make sure to include a comment explaining this caveat!

crates/handlers/src/admin/v1/users/deactivate.rs
```diff
@@ -49,7 +51,25 @@ impl IntoResponse for RouteError {
...
 }
 
-pub fn doc(operation: TransformOperation) -> TransformOperation {
+/// # JSON payload for the `POST /api/admin/v1/users/:id/deactivate` endpoint
+#[derive(Default, Deserialize, JsonSchema)]
+#[serde(rename = "DeactivateUserRequest")]
+pub struct Request {
+    /// Whether to skip locking the user before deactivation.
+    #[serde(default)]
+    skip_lock: bool,
```
sandhose: If this is to keep the previous behaviour by default, I'm fine with breaking the API and not have it lock

crates/handlers/src/admin/v1/users/deactivate.rs
```diff
@@ -76,15 +96,17 @@ pub async fn handler(
...
+    let Json(params) = body.unwrap_or_default();
     let id = *id;
     let mut user = repo
         .user()
         .lookup(id)
         .await?
         .ok_or(RouteError::NotFound(id))?;
 
-    if user.locked_at.is_none() {
+    if !params.skip_lock && user.locked_at.is_none() {
```
sandhose: I'm not sure what the better behaviour is. Obviously, deactivation is expansive, so you kinda want it to happen in the background. At the same time, if we don't flip the deactivated bit on the user right away, it will still be valid for some time?

Maybe this should just flip the deactivated bit right away and schedule a deactivation job?

sandhose (changes requested):

crates/handlers/src/admin/v1/users/deactivate.rs
AndrewFerr: Thanks for the go-ahead on API-breaking changes.  So I'll opt for making {d,r}eactivate and {,un}lock do only their namesake action.

crates/handlers/src/admin/v1/users/deactivate.rs
AndrewFerr: Agreed that flipping the deactivate flag (as locked flag is flipped now) before launching the job is the safest thing to do here.

crates/handlers/src/admin/v1/users/unlock.rs
AndrewFerr: I'll go with the former, so that unlocking just unlocks.  This will let MAS have unlocked + reactivated accounts, like Synapse does.

Should the CLI's unlock handler be updated too, to match this new behaviour?

sandhose: Yes please!

AndrewFerr: Done with a841312

crates/handlers/src/admin/v1/users/deactivate.rs
AndrewFerr: a841312

crates/handlers/src/graphql/mutations/user.rs
```diff
@@ -585,9 +588,16 @@ impl UserMutations {
             return Ok(UnlockUserPayload::NotFound);
         };
 
-        // Call the homeserver synchronously to unlock the user
```
sandhose, line 588: Maybe this should have the 'old' behaviour of reactivating. This is an internal API, and no one should be relying on this except us, so to avoid breaking our internal tooling, I would not introduce new flags, and make this API reactivate users properly

sandhose (changes requested):
Almost there! I just want to avoid changes in behaviour on the GraphQL API, as it's used in our tools that haven't migrated to the admin API yet… unless you want to also fix those tools :D

crates/handlers/src/graphql/mutations/user.rs
AndrewFerr: Completely reasonable!  Done with 43806cdf (along with a slight update to documentation)

sandhose (approved):

PR #4775, opened 2025-07-09
Allow running jobs from the job queue in tests

sandhose:
- **Make the task State::clock() return a &dyn Clock instead of a BoxClock**
- **Remove the duplicate clock and rng in QueueWorker**
- **Allow setting a custom clock on the QueueWorker & add one to the TestState**
- **Test helper to run all tests in the job queue**
- **Update the deactivate admin API test to run the deactivation job**

### DIFF
--- a/crates/handlers/src/test_utils.rs
+++ b/crates/handlers/src/test_utils.rs
@@ -115,7 +115,7 @@ pub(crate) struct TestState {
     pub rng: Arc<Mutex<ChaChaRng>>,
     pub http_client: reqwest::Client,
     pub task_tracker: TaskTracker,
-    queue_worker: Arc<Mutex<QueueWorker>>,
+    queue_worker: Arc<tokio::sync::Mutex<QueueWorker>>,
 
     #[allow(dead_code)] // It is used, as it will cancel the CancellationToken when dropped
     cancellation_drop_guard: Arc<DropGuard>,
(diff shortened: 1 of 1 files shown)


crates/handlers/src/admin/v1/users/deactivate.rs
```diff
@@ -137,15 +138,37 @@ mod tests {
...
-        // XXX: we don't have a good way to look for the deactivation job
-        let job: Json<serde_json::Value> = sqlx::query_scalar(
-            "SELECT payload FROM queue_jobs WHERE queue_name = 'deactivate-user'",
-        )
-        .fetch_one(&pool)
-        .await
-        .expect("Deactivation job to be scheduled");
-        assert_eq!(job["user_id"], serde_json::json!(user.id));
+        // Make sure to run the jobs in the queue
+        state.run_jobs_in_queue().await;
```
AndrewFerr, line 142: Shouldn't the jobs have finished running by this point?  If the job finishes after the response to `/deactivate` is generated, then it's possible for some side effects of the job to be missing from the response (namely the value of the `deactivated_at` timestamp).

sandhose: With this, in test they don't run in the background, you *have* to call `State::run_jobs_in_queue`

reivilibre (approved):

PR #4776, opened 2025-07-10
Fix link title

Johennes:
I guess this should not mention Synapse since it's the MAS contribution guide?


sandhose (approved):
Oh, I totally did not copy-paste this from the Synapse contributing docs… thanks!

PR #4788, opened 2025-07-14
Make the admin API provision users synchronously

AndrewFerr:
as opposed to always launching an asynchronous worker job.

This allows callers to have a guarantee that the user is fully created by the time it receives the response to the user creation request.

### DIFF
--- a/crates/handlers/src/admin/v1/users/add.rs
+++ b/crates/handlers/src/admin/v1/users/add.rs
@@ -11,10 +11,7 @@ use axum::{Json, extract::State, response::IntoResponse};
 use hyper::StatusCode;
 use mas_axum_utils::record_error;
 use mas_matrix::{HomeserverConnection, ProvisionRequest};
-use mas_storage::{
-    BoxRng,
-    queue::{ProvisionUserJob, QueueJobRepositoryExt as _},
-};
+use mas_storage::BoxRng;
 use schemars::JsonSchema;
 use serde::Deserialize;
 use tracing::warn;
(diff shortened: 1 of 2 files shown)


crates/handlers/src/admin/v1/users/add.rs
```diff
@@ -168,9 +172,19 @@ pub async fn handler(
 
     let user = repo.user().add(&mut rng, &clock, params.username).await?;
 
-    repo.queue_job()
-        .schedule_job(&mut rng, &clock, ProvisionUserJob::new(&user))
-        .await?;
+    if params.add_synchronously {
```
sandhose: I think I wouldn't mind making this always synchronous? Especially when called with skip_homeserver_check, you may end up with a user that is not valid, and fails to provision on the HS without you noticing.

The counter-argument of this is that it prevents creating users if Synapse is not healthy… but that's also probably fine?

AndrewFerr: Sounds reasonable.

Regarding your second point, how about waiting to create the MAS user until after the homeserver user is created?  Is there any danger in not creating the MAS user first?

AndrewFerr: Ah of course -- the homeserver user creation needs the numeric ID of the MAS user, so the latter must be created first.

AndrewFerr: 0d613a52

sandhose: >  how about waiting to create the MAS user until after the homeserver user is created

if you keep it in the transaction, this is what's going to happen anyway! :)

AndrewFerr: The magic of transactions! Thanks!

PR #4789, opened 2025-07-14
Support M_USER_LOCKED error for compat sessions

AndrewFerr:


crates/handlers/src/compat/login.rs
```diff
@@ -506,7 +514,13 @@ async fn token_login(
             browser_session.id = %browser_session_id,
             "Attempt to exchange login token but browser session is not active"
         );
-        return Err(RouteError::InvalidLoginToken);
```
sandhose, line 509: I think it would be more readable to check if the user is locked earlier

```rust
if browser_session.user.locked_at.is_some() {
  return Err(RouteError::UserLocked);
}

if !browser_session.active() || !browser_session.user.is_valid() {
  // ...
}
```

It changes a little bit what error we show in what condition, but it's an edge-case anyway, as usually we would show a user comprehensible error earlier in the UI

crates/handlers/src/compat/login.rs
```diff
@@ -565,9 +579,13 @@ async fn user_password_login(
         .user()
         .find_by_username(username)
         .await?
-        .filter(mas_data_model::User::is_valid)
+        .filter(|user| user.deactivated_at.is_none())
```
sandhose, line 584: I wonder if we should add User::is_deactivate/User::is_locked, it would make things more readable?

crates/handlers/src/compat/login.rs
```diff
@@ -565,9 +579,13 @@ async fn user_password_login(
...
         .ok_or(RouteError::UserNotFound)?;
 
+    if user.locked_at.is_some() {
+        return Err(RouteError::UserLocked);
```
sandhose, line 588: This is quite easy to test, could you add a test for this case?

sandhose (commented):
Nothing major, thanks for this!

crates/handlers/src/compat/login.rs
AndrewFerr: This is the only place that would benefit from `is_none` being wrapped a method, so I'm willing to make this take a readability hit.

crates/handlers/src/compat/login.rs
AndrewFerr: 0d7fde0e

crates/handlers/src/compat/login.rs
AndrewFerr: d31af23e

crates/handlers/src/compat/login.rs
AndrewFerr: On second thought, I've reverted this with 05827d1e (the commit message explains why).  The logic is more precise this time, though.

sandhose: Alright, works for me!

sandhose (approved):

PR #4801, opened 2025-07-18
Use the new dedicated Synapse API

sandhose:
This adds support for using the new dedicated 'MAS' API in Synapse for MAS->Synapse communication introduced in https://github.com/element-hq/synapse/pull/18520

This removes the dependency on the Synapse admin API, making it more focused, and will let us remove the fact that the shared secret is effectively a site-wide admin token.

I've kept the old implementation around so that we can gradually roll this out. The plan is:

 - include the client and server part of this API in the next MAS (0.20.0) and Synapse (1.135.0) versions
 - but opt-in by default (with the `synapse_modern` homeserver type)
 - make it the default in the next version, requiring Synapse 1.135.0 or later (but keep the old one under the `synapse_legacy` homeserver type)
 - remove the legacy API in the version after that


crates/matrix-synapse/src/modern.rs
```diff
@@ -0,0 +1,576 @@
...
+            matrix.homeserver = self.homeserver,
+            matrix.localpart = localpart,
+        ),
+        err(Debug),
+    )]
+    async fn allow_cross_signing_reset(&self, localpart: &str) -> Result<(), anyhow::Error> {
+        #[derive(Serialize)]
+        struct Request {
+            localpart: String,
+            password: String, // Required by the API but not used in this context
```
reivilibre: curious why this Synapse-side API needs a password anyway?

docs/config.schema.json
```diff
@@ -1707,18 +1707,32 @@
...
-          "description": "Homeserver is Synapse",
+          "description": "Homeserver is Synapse, using the legacy API\n\nThis will switch to using the modern API in a few releases.",
           "type": "string",
           "enum": [
             "synapse"
           ]
         },
         {
-          "description": "Homeserver is Synapse, in read-only mode\n\nThis is meant for testing rolling out Matrix Authentication Service with no risk of writing data to the homeserver.",
+          "description": "Homeserver is Synapse, using the legacy API, in read-only mode\n\nThis is meant for testing rolling out Matrix Authentication Service with no risk of writing data to the homeserver.",
```
reivilibre: should we threaten to switch this to the modern API in a few releases too?

reivilibre (approved):
seems fair!

crates/matrix-synapse/src/modern.rs
sandhose: woops, for some reason Claude generated that field in Synapse, so when I asked it to do the implementation on the MAS side, it still had it

sandhose: Fixed in fbcfa25cb

docs/config.schema.json
sandhose: Fixed in 832244675

PR #4807, opened 2025-07-21
docs: Remove requirement for arbitrary KIDs

V02460:
The docs currently note that the `kid` property specified for `secrets.keys` items has to be arbitrary. This PR removes that requirement and documents more loose ones instead.

The JWT spec itself does have [only few restrictions on KIDs](https://datatracker.ietf.org/doc/html/rfc7517#section-4.5). Having this additional requirement of arbitrary KIDs in MAS would disqualify the use of a hash function or a running number to create KIDs.

The PR also clarifies that it is ok for KIDs to change between restarts.

The suggested documentation would make automatic key management easier for me. I think these more loose requirements don’t break anything in MAS, but please correct me if I’m wrong.

PS: I’m thinking about making the `kid` property optional and auto-generating it by MAS if missing, e.g., by hashing the corresponding key. Would you take such a PR?

### DIFF
--- a/docs/reference/configuration.md
+++ b/docs/reference/configuration.md
@@ -257,8 +257,10 @@ The following key types are supported:
 - ECDSA with the P-384 (`secp384r1`) curve
 - ECDSA with the K-256 (`secp256k1`) curve
 
-Each entry must have a unique (and arbitrary) `kid`, plus the key itself.
-The key can either be specified inline (with the `key` property), or loaded from a file (with the `key_file` property).
+Each entry must have a unique `kid`, plus the key itself. The `kid` can be any
+case-sensitive string value as long as it is unique to this list; `kid` values
+must not be stable across restarts. The key can either be specified inline (with
+the `key` property), or loaded from a file (with the `key_file` property).
 The following key formats are supported:
 
 - PKCS#1 PEM or DER-encoded RSA private key


docs/reference/configuration.md
```diff
@@ -257,8 +257,10 @@ The following key types are supported:
 - ECDSA with the P-384 (`secp384r1`) curve
 - ECDSA with the K-256 (`secp256k1`) curve
 
-Each entry must have a unique (and arbitrary) `kid`, plus the key itself.
-The key can either be specified inline (with the `key` property), or loaded from a file (with the `key_file` property).
+Each entry must have a unique `kid`, plus the key itself. The `kid` can be any
+case-sensitive string value as long as it is unique to this list; `kid` values
+must not be stable across restarts. The key can either be specified inline (with
+the `key` property), or loaded from a file (with the `key_file` property).
```
sandhose: Could you instead use [one sentence per line](https://nick.groenen.me/notes/one-sentence-per-line/) instead of wrapping at 80 characters?

sandhose (changes requested):
I think the 'arbitrary' here was meant to say 'any value you want, doesn't matter'. That's probably something that sounds about right in french, but not in english :)

I'd be happy to have the `kid` derived from a hash; the reason I didn't do it back then was that I couldn't really find a standard key fingerprinting mechanism, but admittedly I didn't look very far

docs/reference/configuration.md
```diff
@@ -257,8 +257,11 @@ The following key types are supported:
 - ECDSA with the P-384 (`secp384r1`) curve
 - ECDSA with the K-256 (`secp256k1`) curve
 
-Each entry must have a unique (and arbitrary) `kid`, plus the key itself.
-The key can either be specified inline (with the `key` property), or loaded from a file (with the `key_file` property).
+Each entry must have a unique `kid`, plus the key itself.
+The `kid` can be any case-sensitive string value as long as it is unique to this list;
+`kid` values don’t need to be stable across restarts.
```
sandhose: That is actually not true? If we issue an id_token and change the kid of the key that was used to sign it, the client won't be able to verify it

V02460: Ok, I hope this is better now?

sandhose (approved):

PR #4808, opened 2025-07-22
Allow the homeserver to perform introspection using a shared secret

sandhose:
The intent is to simplify the setup process by only requiring a shared secret between the HS and MAS, instead of having to provision a static client for the HS.

### DIFF
--- a/crates/axum-utils/src/client_authorization.rs
+++ b/crates/axum-utils/src/client_authorization.rs
@@ -9,13 +9,12 @@ use std::collections::HashMap;
 use axum::{
     BoxError, Json,
     extract::{
-        Form, FromRequest, FromRequestParts,
+        Form, FromRequest,
         rejection::{FailedToDeserializeForm, FormRejection},
     },
     response::IntoResponse,
 };
-use axum_extra::typed_header::{TypedHeader, TypedHeaderRejectionReason};
-use headers::{Authorization, authorization::Basic};
+use headers::authorization::{Basic, Bearer, Credentials as _};
 use http::{Request, StatusCode};
 use mas_data_model::{Client, JwksOrJwksUri};
 use mas_http::RequestBuilderExt;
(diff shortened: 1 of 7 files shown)


crates/axum-utils/src/client_authorization.rs
```diff
@@ -360,25 +373,36 @@ where
...
+        enum Authorization {
+            Basic(String, String),
+            Bearer(String),
+        }
+
+        // Sadly, the typed-header 'Authorization' doesn't let us check for both
+        // Basic and Bearer at the same time, so we need to parse them manually
+        let authorization = if let Some(header) = req.headers().get(http::header::AUTHORIZATION) {
+            let bytes = header.as_bytes();
+            if bytes.len() >= 6 && bytes[..6].eq_ignore_ascii_case(b"Basic ") {
```
reivilibre, line 385:
```suggestion
            if bytes.starts_with(b"Basic ") {
```

crates/axum-utils/src/client_authorization.rs
```diff
@@ -360,25 +373,36 @@ where
...
+                    return Err(ClientAuthorizationError::InvalidHeader);
+                };
 
-        // Reconstruct the request from the parts
-        let req = Request::from_parts(parts, body);
+                Some(Authorization::Basic(
+                    decoded.username().to_owned(),
+                    decoded.password().to_owned(),
+                ))
+            } else if bytes.len() >= 7 && bytes[..7].eq_ignore_ascii_case(b"Bearer ") {
```
reivilibre, line 394:
```suggestion
            } else if bytes.starts_with(b"Bearer ") {
```

crates/handlers/src/oauth2/introspection.rs
```diff
@@ -205,28 +222,41 @@ pub(crate) async fn post(
...
-        .fetch(&mut repo)
-        .await?
-        .ok_or(RouteError::ClientNotFound)?;
-
-    let method = match &client.token_endpoint_auth_method {
-        None | Some(OAuthClientAuthenticationMethod::None) => {
-            return Err(RouteError::NotAllowed(client.id));
+    if let Some(token) = credentials.bearer_token() {
+        // If the client presented a bearer token, we check with the homeserver
+        // connection if it is allowed to use the introspection endpoint
```
reivilibre:
```suggestion
        // configuration if it is allowed to use the introspection endpoint
```
I think makes a bit more sense?
We aren't actually making a request to the HS to validate this token. (Maybe it shouldn't be async)

reivilibre (approved):
Seems fair enough

crates/axum-utils/src/client_authorization.rs
sandhose: We want to make a case-insensitive comparison, hence the `eq_ignore_ascii_case`

crates/handlers/src/oauth2/introspection.rs
sandhose: it's just that this trait is called `HomeserverConnection`, but no strong opinion here

I made it async thinking that maybe we may want to do something else for verifying the token but maybe I'm optimising a future change too early 🤷

crates/axum-utils/src/client_authorization.rs
reivilibre: ah derp, fine :)

PR #4818, opened 2025-07-24
Clarify that the order of secrets.keys is not significant

### LINKED ISSUE #4806: Clarify significance of secrets.keys order
Please clarify whether the ordering of keys in the [secrets.keys](https://element-hq.github.io/matrix-authentication-service/reference/configuration.html#secretskeys) array is of any significance.

The JWK spec assumes by default [no significance of ordering](https://datatracker.ietf.org/doc/html/rfc7517#section-5.1), but allows for it. The MAS doc comments of the `JsonWebKeySet` type are conflicting, with [`find_key`](https://github.com/element-hq/matrix-authentication-service/blob/5eb8e780cfd8ab165d020109cc58528bdef4e5b6/crates/jose/src/jwk/mod.rs#L309-L311) returning “the best key given the constraints”, but [`signing_key_for_algorithm`](https://github.com/element-hq/matrix-authentication-service/blob/5eb8e780cfd8ab165d020109cc58528bdef4e5b6/crates/jose/src/jwk/mod.rs#L327-L330) returning “a key for the given algorithm”.

V02460:
I think the order of keys in the `secrets.keys` list does not matter, but would like to have this confirmed. Please double-check the suggested wording.

Closes #4806

### DIFF
--- a/docs/reference/configuration.md
+++ b/docs/reference/configuration.md
@@ -270,6 +270,8 @@ The following key formats are supported:
 
 For PKCS#8 encoded keys, the `password` or `password_file` properties can be used to decrypt the key.
 
+The order of keys in the list is of no significance.
+
 ## `passwords`
 
 Settings related to the local password database


sandhose:
Actually as explained in https://github.com/element-hq/matrix-authentication-service/issues/4806, they are *technically* significant, but not really in practice. I'm not really sure it makes sense to state anything in the docs though; please reopen if you think otherwise

PR #4829, opened 2025-07-27
fix: don't attempt to set TCP_NODELAY on accepted unix streams

networkException:
This pull request fixes accepting streams from unix sockets resulting in operation not supported errors.


sandhose (approved):
Thank you!

PR #4847, opened 2025-07-31
feat: add support for mandatory upstream pushed authorization requests

Johennes:
This adds support for upstream [PARs](https://www.rfc-editor.org/rfc/rfc9126.html) in the situation that the provider enforces them.

This will probably need some tests but I wanted to get some feedback on the general direction first.


sandhose:
This has bit-rotten quite a bit. I'd be still interested at some point with introducing PAR for both the upstream OAuth 2.0 bits and on the Matrix side though, but for now I'll close this PR to clear a bit the backlog 📉

PR #4858, opened 2025-08-05
Use the modern Synapse API (>= 1.135.0) by default

sandhose:
This changes the Synapse 'integration' to use the new API introduced in Synapse 1.135.0 by default. This is needed to use the new MAS integration in Synapse that will be introduced in 1.136.0


anoadragon453 (approved):
While it seems rather soon to default to Synapse v1.135.0, I understand wanting this to be the default for MAS 1.0.0.

Synapse v1.136.0 should be out once MAS 1.0.0 comes out, at least.

PR #4873, opened 2025-08-07
Docs: Run local tests via `cargo-nextest`

jaywink:
Mention this specifically in contributing.md since CI also uses this test runner. My attempt to use `cargo test` failed due to PG connection pooling. In any case, unless there are reasons not to, we should keep close to mirroring what we do in CI.

### DIFF
--- a/docs/development/contributing.md
+++ b/docs/development/contributing.md
@@ -91,7 +91,7 @@ While you're developing and before submitting a patch, you'll want to test your 
 
 If you haven't already, install [Cargo-Nextest](https://nexte.st/docs/installation/pre-built-binaries/).
 
-- Run the tests to the backend by running `cargo-nextest nextest run --workspace`. This requires a connection to a PostgreSQL database, set via the `DATABASE_URL` environment variable.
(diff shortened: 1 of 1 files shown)


docs/development/contributing.md
```diff
@@ -89,7 +89,9 @@ While you're developing and before submitting a patch, you'll want to test your
 
 ### Run the tests
 
-- Run the tests to the backend by running `cargo test --workspace`. This requires a connection to a PostgreSQL database, set via the `DATABASE_URL` environment variable.
+If you haven't already, install [Cargo-Nextest](https://nexte.st/docs/installation/pre-built-binaries/).
+
+- Run the tests to the backend by running `cargo-nextest nextest run --workspace`. This requires a connection to a PostgreSQL database, set via the `DATABASE_URL` environment variable.
```
sandhose:
```suggestion
- Run the tests to the backend by running `cargo nextest run --workspace`. This requires a connection to a PostgreSQL database, set via the `DATABASE_URL` environment variable.
```

PR #4876, opened 2025-08-08
Automatically derive the `kid` from the key fingerprint if missing

V02460:
Implements deriving a key’s ID if no explicit `kid` is provided. Contains unit tests and documentation.

`secrets.keys.[].kid` can now be omitted and in this case the kid is automatically derived. If the field is set, the behavior is the same as before.

Edit: The derived kid is now the key’s [RFC 7638 JWK Thumbprint](https://www.rfc-editor.org/rfc/rfc7638.html).

~~The derived kid is based on a key’s fingerprint: it simply takes the first four bytes and hex-encodes them.~~

~~For creating the fingerprinting function I took inspiration from how [OpenSSH](https://superuser.com/questions/421997/what-is-a-ssh-key-fingerprint-and-how-is-it-generated/714195#714195) and [AWS](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/verify-keys.html) are doing their key fingerprinting. They take some canonical byte-representation and hash them, but differ in the used hashing function and whether the private key or corresponding public key is used. I ended up using sha256 hashing on the corresponding public key encoded as PSCK#8 DER.~~

### DIFF
--- a/crates/config/src/sections/secrets.rs
+++ b/crates/config/src/sections/secrets.rs
@@ -9,13 +9,9 @@ use std::borrow::Cow;
 use anyhow::{Context, bail};
 use camino::Utf8PathBuf;
 use futures_util::future::{try_join, try_join_all};
-use mas_jose::jwk::{JsonWebKey, JsonWebKeySet};
+use mas_jose::jwk::{JsonWebKey, JsonWebKeySet, Thumbprint};
 use mas_keystore::{Encrypter, Keystore, PrivateKey};
-use rand::{
-    Rng, SeedableRng,
-    distributions::{Alphanumeric, DistString, Standard},
-    prelude::Distribution as _,
-};
+use rand::{Rng, SeedableRng, distributions::Standard, prelude::Distribution as _};
 use schemars::JsonSchema;
 use serde::{Deserialize, Serialize};
 use serde_with::serde_as;
@@ -134,8 +130,7 @@ impl From<Key> for KeyRaw {
 pub struct KeyConfig {
     /// The key ID `kid` of the key as used by JWKs.
     ///
-    /// If not given, `kid` will be derived from the key by hex-encoding the
-    /// first four bytes of the key’s fingerprint.
+    /// If not given, `kid` will be the key’s RFC 7638 JWK Thumbprint.
     #[serde(skip_serializing_if = "Option::is_none")]
     kid: Option<String>,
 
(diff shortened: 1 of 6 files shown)


crates/config/src/sections/secrets.rs
```diff
@@ -178,12 +183,24 @@ impl KeyConfig {
...
-            .with_kid(self.kid.clone())
+            .with_kid(kid)
             .with_use(mas_iana::jose::JsonWebKeyUse::Sig))
     }
 }
 
+/// Returns a kid derived from the given key.
+fn kid_from_key(private_key: &PrivateKey) -> anyhow::Result<String> {
+    let fingerprint = private_key.fingerprint()?;
+    let head = fingerprint.first_chunk::<4>().unwrap();
```
sandhose: We're in a context where it's easy to return an error, so even if it really shouldn't happen
```suggestion
    let head = fingerprint.first_chunk::<4>()
      .context("Key fingerprint must be at least 4 bytes long!")
```

crates/config/src/sections/secrets.rs
```diff
@@ -322,7 +339,7 @@ impl SecretsConfig {
         .await
         .context("could not join blocking task")?;
         let rsa_key = KeyConfig {
-            kid: Alphanumeric.sample_string(&mut rng, 10),
+            kid: Some(Alphanumeric.sample_string(&mut rng, 10)),
```
sandhose: We could skip that now from the config generation?

crates/config/src/sections/secrets.rs
```diff
@@ -383,7 +400,7 @@ impl SecretsConfig {
 
     pub(crate) fn test() -> Self {
         let rsa_key = KeyConfig {
-            kid: "abcdef".to_owned(),
+            kid: Some("abcdef".to_owned()),
```
sandhose: Same from the test config

crates/keystore/src/lib.rs
```diff
@@ -179,6 +183,24 @@ impl PrivateKey {
         }
     }
 
+    /// Returns the fingerprint of the private key.
+    ///
+    /// The fingerprint is calculated as the SHA256 sum over the PKCS#8 ASN.1
```
sandhose: Technically I think it's the `SPKI`/`SubjectPublicKeyInfo`, not `PKCS#8`? PKCS#8 is a wrapper for private keys

crates/keystore/src/lib.rs
```diff
@@ -179,6 +183,24 @@ impl PrivateKey {
...
+    /// DER-encoded bytes of the private key’s corresponding public key.
```
sandhose: We coooould also use RFC 7638 JWK thumbprint: https://datatracker.ietf.org/doc/html/rfc7638

What do you think? I'm also fine with keeping using the SPKI digest

crates/config/src/sections/secrets.rs
sandhose: Actually do we really need to trim that? We don't really care if the `kid` is a  little bit long, do we?

crates/keystore/src/lib.rs
V02460: That’s completely right!

For my record: SPKI is from [RFC5280 section 4.1](https://datatracker.ietf.org/doc/html/rfc5280#section-4.1)

crates/keystore/src/lib.rs
V02460: Of course I missed that >.< It’s thumbprints now.

PR #4877, opened 2025-08-08
Add `secrets.keys_dir` config option.

V02460:
Implements the `secrets.keys_dir` config option as an alternative to `secrets.keys`. Contains unit tests and documentation.

Currently managing MAS key material is intertwined with managing MAS configuration: it is not possible to provide signing keys to MAS without some prior knowledge of their content, filename or count. This PR resolves that limitation by introducing the `secrets.keys_dir` config option.

The `secrets.keys_dir` config option allows specifying a single directory that is expected to contain one file per MAS signingkey. The key files are read from the directory when a `KeyStore` is created via `SecretsConfig::key_store()`.

Requires #4876 to derive a key’s kid.

### DIFF
--- a/crates/config/src/sections/secrets.rs
+++ b/crates/config/src/sections/secrets.rs
@@ -240,28 +240,6 @@ impl From<Encryption> for EncryptionRaw {
     }
 }
 
-/// Description of signing keys.
-///
-/// It either holds the key config values directly or references a directory
-/// where each file contains a key.
-#[derive(Debug, Clone)]
-pub enum Keys {
-    Values(Vec<KeyConfig>),
-    Directory(Utf8PathBuf),
-}
-
-impl Keys {
-    /// Returns a list of key configs.
-    ///
-    /// If `keys_dir` was given, the keys are read from file.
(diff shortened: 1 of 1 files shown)


crates/config/src/sections/secrets.rs
```diff
@@ -255,8 +335,10 @@ pub struct SecretsConfig {
     encryption: Encryption,
 
     /// List of private keys to use for signing and encrypting payloads
-    #[serde(default)]
-    keys: Vec<KeyConfig>,
+    #[schemars(with = "KeysRaw")]
+    #[serde_as(as = "serde_with::TryFromInto<KeysRaw>")]
+    #[serde(flatten)]
+    keys: Keys,
```
sandhose: I think they should be additive, not exclusive in this case, since it's a list

V02460: Ok, done :) In this implementation keys defined inline take precedence over keys from the directory, but I think that isn’t a problem either way.

sandhose (approved):
tyvm, looks good to me now!

PR #4882, opened 2025-08-11
Add `upstream_oauth2.providers.[].client_secret_file` config option

networkException:
This pull request factors out the previously introduced config wrapper for client secrets to also use it for upstream oauth providers.

See a7e7c3caa18266f962c1135c1d4a42d5967d5896

Disclosure: I work for the German government agency FITKO where I also participate in the Matrix Community. This pull request is written and published by me personally without my work hat on.


crates/config/src/sections/mod.rs
sandhose, line 1: Even though the two 'client secrets' are used in very different context, I'm fine with this refactoring

sandhose (approved):
Looks good to me, tyvm!

PR #4902, opened 2025-08-18
Use the stable MAS integration of Synapse in the documentation and in `mas-cli doctor`

sandhose:
- **Update `mas-cli doctor` to suggest the stable MAS integration**
- **Document Synapse integration with the stable feature**


crates/cli/src/commands/doctor.rs
```diff
@@ -278,62 +275,37 @@ Check that the homeserver is running."#
...
-                            r#"❌ A Synapse admin API endpoint at "{server_version}" replied with {status}.
-Make sure MAS can reach the admin API, and that the homeserver is running.
-"#
-                        );
-                    }
-                }
-                Err(e) => error!(
-                    r#"❌ Can't reach the Synapse admin API at "{server_version}".
-Make sure MAS can reach the admin API, and that the homeserver is running.
-
```
anoadragon453, line 305: Why is this no longer useful? Doesn't MAS still need to use Synapse's Admin API in several cases?

anoadragon453 (changes requested):

crates/cli/src/commands/doctor.rs
sandhose: It's using a dedicated API now, and doesn't have full admin access anymore

sandhose: This is also now just testing for the dedicated MAS API, which should cover both the 'it's unreachable' and the 'it's the wrong secret' cases

reivilibre (approved):

PR #4961, opened 2025-08-30
Trim the secret read by the `matrix.secret_file` configuration option

mweinelt:
Treat the secret the same as synapse, so the same secret file can be reused between mas and synapse.

https://github.com/element-hq/synapse/blob/v1.137.0/synapse/config/experimental.py#L50

### DIFF
--- a/crates/config/src/sections/matrix.rs
+++ b/crates/config/src/sections/matrix.rs
@@ -131,7 +131,11 @@ impl MatrixConfig {
     /// Returns an error when the shared secret could not be read from file.
     pub async fn secret(&self) -> anyhow::Result<String> {
         Ok(match &self.secret {
-            Secret::File(path) => tokio::fs::read_to_string(path).await?,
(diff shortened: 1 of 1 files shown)


crates/config/src/sections/matrix.rs
```diff
@@ -130,10 +130,12 @@ impl MatrixConfig {
     ///
     /// Returns an error when the shared secret could not be read from file.
     pub async fn secret(&self) -> anyhow::Result<String> {
-        Ok(match &self.secret {
+        let raw = match &self.secret {
             Secret::File(path) => tokio::fs::read_to_string(path).await?,
```
sandhose: I think we should `trim` only if the secret is being read from a file to match synapse behaviour. It would also be nice to stick in a comment to explicit this behaviour

mweinelt: Updated.

PR #5014, opened 2025-09-15
Expose the legacy guest/non-guest flag on users in the admin API

sandhose:
- **Surface the user guest flag in the admin API**
- **Allow filtering guest/non-guest users**


crates/storage-pg/src/user/mod.rs
```diff
@@ -114,6 +116,10 @@ impl Filter for UserFilter<'_> {
             .add_option(self.can_request_admin().map(|can_request_admin| {
                 Expr::col((Users::Table, Users::CanRequestAdmin)).eq(can_request_admin)
             }))
+            .add_option(
```
reivilibre, line 119: is the index on `(is_guest, can_request_admin)` intended to be a future PR?

sandhose: Yes, but also low-priority for now, it is performant enough in most cases

PR #5015, opened 2025-09-15
Admin API filter to search users by username

sandhose:


crates/storage-pg/migrations/20250915092000_pgtrgm_extension.sql
```diff
@@ -0,0 +1,10 @@
+-- no-transaction
+-- Copyright 2025 New Vector Ltd.
+--
+-- SPDX-License-Identifier: AGPL-3.0-only OR LicenseRef-Element-Commercial
+-- Please see LICENSE in the repository root for full details.
+
+-- This enables the pg_trgm extension, which is used for search filters
+-- Starting Posgres 16, this extension is marked as "trusted", meaning it can be
+-- installed by non-superusers
+CREATE EXTENSION IF NOT EXISTS pg_trgm;
```
sandhose, line 10: The only tricky thing is that before PG 16 you have to be superuser to execute that. Which is probably fine in most Docker environments for example?

reivilibre: hmm yikes. Feels like something we should at least call out somewhere, v15 still has another 2 years on it and I'm not sure people will know how to run this manually as a superuser.

reivilibre (changes requested):
two things to do:

- add documentation to the database page and upgrade notes for this migration, for Pg <= 15;
- check that if the superuser on Pg 15 already created the extension, whether a non-superuser is then able to complete the migration as a no-op?

sandhose:
@reivilibre I've added things in the documentation in a3ecdc9 and put what should be added to the release notes in the PR body, wdyt?

PR #5091, opened 2025-09-30
Add admin APIs to finish individual sessions

sandhose:
This adds 3 endpoints to finish sessions, one for each kind of session.

### DIFF
--- a/crates/handlers/src/admin/v1/compat_sessions/finish.rs
+++ b/crates/handlers/src/admin/v1/compat_sessions/finish.rs
@@ -99,17 +99,14 @@ pub async fn handler(
         return Err(RouteError::AlreadyFinished(id));
     }
 
-    // Load the user to schedule a device sync job
-    let user = repo
-        .user()
-        .lookup(session.user_id)
-        .await?
-        .ok_or_else(|| RouteError::Internal("User not found for session".into()))?;
-
     // Schedule a job to sync the devices of the user with the homeserver
-    tracing::info!(user.id = %user.id, "Scheduling device sync job for user");
+    tracing::info!(user.id = %session.user_id, "Scheduling device sync job for user");
     repo.queue_job()
-        .schedule_job(&mut rng, &clock, SyncDevicesJob::new(&user))
+        .schedule_job(
+            &mut rng,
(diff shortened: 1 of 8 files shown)


crates/handlers/src/admin/v1/compat_sessions/finish.rs
```diff
@@ -0,0 +1,246 @@
...
+    // Finish the session
+    let session = repo.compat_session().finish(&clock, session).await?;
+
+    // Get the SSO login info for the response
+    let sso_login = repo.compat_sso_login().find_for_session(&session).await?;
+
+    repo.save().await?;
+
+    Ok(Json(SingleResponse::new(
+        CompatSession::from((session, sso_login)),
```
reivilibre, line 121: slight eyebrow raise at `CompatSession::from((CompatSession, Option<CompatSsoLogin>))` but I guess it's OK and matches what we do in other places

crates/handlers/src/admin/v1/oauth2_sessions/finish.rs
```diff
@@ -0,0 +1,234 @@
...
+
+    // Check if the session is already finished
+    if session.finished_at().is_some() {
+        return Err(RouteError::AlreadyFinished(id));
+    }
+
+    // If the session has a user associated with it, schedule a job to sync devices
+    if let Some(user_id) = session.user_id {
+        tracing::info!(user.id = %user_id, "Scheduling device sync job for user");
+        let job = SyncDevicesJob::new_for_id(user_id);
```
reivilibre, line 105: weirdly inconsistent with how it's done for CompatSessions; why use the db-loaded `User` in one place and the ID in the other?

I think just using the `user_id` should win, though; should be no need to verify the user exists in the DB given the FK constraint should do that for us

reivilibre (approved):
Seems fine overall

crates/handlers/src/admin/v1/oauth2_sessions/finish.rs
sandhose: Good q, dunno why I did that on the compat sessions. Fixed that in 7e3df5585

Issue #4492, opened 2025-05-02
Personal access tokens


sandhose:
The new auth paradigm broke a few workflows for our end users, especially for working with bots and automated tools.

It broke in two main ways:

 - the main OAuth 2.0 flow, which can grant admin access to the Synapse and MAS admin APIs uses short-lived tokens, making it unfit for automated tools
 - the compatibility flow doesn't let apps access the Synapse or MAS admin APIs, making it unfit for administrative tools. Also since EW now uses the OAuth 2.0 flow, the original 'just copy the token from EW settings' suggestion doesn't work

One workaround is to use the `mas-cli manage issue-compatibility-token` CLI tool to issue a long-lived, Synapse admin token (but not MAS admin)

This is why we need a feature which lets users create long-lived (potentially infinite) access tokens for automation purposes

Taking inspiration from how it works on GitHub:

 - user chooses what access this token has. For non-admin, it's just the full C-S API
 - admins can request access to the MAS and Synapse admin APIs. Syn admin requires C-S API access
 - admins can request access tokens for other users (?)
 - tokens have a separate owner and actor. Owner manage the token, actor is the user for which the token acts. Actors can't remove the token but they see them in the MAS UI
 - when requesting C-S API access, it generates a device ID
 - tokens have a (mandatory) name and an optional description
 - a token has an expiration, with suggested 7 days, 30 days, 60 days, 90 days, custom, no expiration
 - server config can cap the max expiration date (optional for MVP)
 - a token can be re-generated with the same characteristics, keeping the device ID
 - we don't delete devices from expired tokens, only 'deleted' ones
 - we record and show when it was last used

### How to implement this?

- Introduce a new token format, with prefix `mpt_`
- Introduce a new table, `personal_access_token_families`, with the following:
   - `scope`
   - `created_at`
   - `revoked_at`
   - `owner_user_id`
   - `actor_user_id`
   - `last_active_at`
   - `last_active_ip`
- Introduce a new table, `personal_access_tokens`, with the following:
   - `personal_access_token_family_id`
   - `created_at`
   - `expires_at`
   - `revoked_at`
   - `token`
- Make those work in the token introspection
- Add an admin API to create them
- Show the tokens in the UI:
  - Have a separate list with those tokens
  - Make sure the deep-link to a device works
  - Add an option to regenerate a token from there
  - Add a form to create a new token

Lazywave:
[...]

PR #5106, opened 2025-10-06
Add storage for Personal Access Tokens

reivilibre:
This PR defines a token format for PATs and implements the base storage functionality for PATs and 'Personal Sessions' which are conceptually the parent containers of PATs.

Personal Sessions survive across regenerations of PATs and are the entities associated with the device ID. In virtually every way they are functionally the same as Compat or OAuth2 Sessions.

Future work that will follow soon after this PR:

- admin API
- support PATs in token introspection
- store activity tracking rows for PATs
- ability to regenerate PATs

Part of: #4492

### DIFF
--- a/crates/storage-pg/src/personal/access_token.rs
+++ b/crates/storage-pg/src/personal/access_token.rs
@@ -143,7 +143,7 @@ impl PersonalAccessTokenRepository for PgPersonalAccessTokenRepository<'_> {
         rng: &mut (dyn RngCore + Send),
         clock: &dyn Clock,
         session: &PersonalSession,
-        access_token: String,
+        access_token: &str,
         expires_after: Option<chrono::Duration>,
     ) -> Result<PersonalAccessToken, Self::Error> {
         let created_at = clock.now();
--- a/crates/storage-pg/src/personal/mod.rs
+++ b/crates/storage-pg/src/personal/mod.rs
@@ -227,7 +227,7 @@ mod tests {
                 &mut rng,
                 &clock,
                 &session,
-                FIRST_TOKEN.to_owned(),
+                FIRST_TOKEN,
                 Some(Duration::try_minutes(1).unwrap()),
             )
             .await
@@ -246,7 +246,7 @@ mod tests {
                         &mut rng,
                         &clock,
                         &session,
-                        FIRST_TOKEN.to_owned(),
+                        FIRST_TOKEN,
                         Some(Duration::try_minutes(1).unwrap()),
                     )
                     .await
@@ -288,7 +288,7 @@ mod tests {
         // Add a second access token, this time without expiration
         let token = repo
             .personal_access_token()
-            .add(&mut rng, &clock, &session, SECOND_TOKEN.to_owned(), None)
+            .add(&mut rng, &clock, &session, SECOND_TOKEN, None)
             .await
             .unwrap();
         assert_eq!(token.session_id, session.id);
--- a/crates/storage/src/personal/access_token.rs
+++ b/crates/storage/src/personal/access_token.rs
@@ -71,7 +71,7 @@ pub trait PersonalAccessTokenRepository: Send + Sync {
         rng: &mut (dyn RngCore + Send),
         clock: &dyn Clock,
         session: &PersonalSession,
-        access_token: String,
+        access_token: &str,
         expires_after: Option<Duration>,
     ) -> Result<PersonalAccessToken, Self::Error>;
 
@@ -107,7 +107,7 @@ repository_impl!(PersonalAccessTokenRepository:
         rng: &mut (dyn RngCore + Send),
         clock: &dyn Clock,
         session: &PersonalSession,
-        access_token: String,
+        access_token: &str,
         expires_after: Option<Duration>,
     ) -> Result<PersonalAccessToken, Self::Error>;
 


crates/iana/src/oauth.rs
```diff
@@ -284,6 +284,11 @@ impl schemars::JsonSchema for OAuthAuthorizationEndpointResponseType {
 
 /// OAuth Token Type Hint
 ///
+/// A parameter used in token introspection requests to indicate the type
```
sandhose: unfortunately this is code gen'd by `mas-iana-codegen`

crates/storage-pg/migrations/20250924132713_personal_access_tokens.sql
```diff
@@ -0,0 +1,55 @@
+-- Copyright 2025 New Vector Ltd.
+--
+-- SPDX-License-Identifier: AGPL-3.0-only OR LicenseRef-Element-Commercial
+-- Please see LICENSE in the repository root for full details.
+
+-- A family of personal access tokens. This is a long-lived wrapper around the personal access tokens
+-- themselves, allowing tokens to be regenerated whilst still retaining a persistent identifier for them.
+CREATE TABLE personal_sessions (
+    personal_session_id UUID NOT NULL PRIMARY KEY,
+    owner_user_id UUID NOT NULL REFERENCES users(user_id),
```
sandhose: I wonder if this could be nullable because it could be possible to create a personal session using the client credentials? In this case the owner is the client I guess?

crates/storage-pg/migrations/20250924132713_personal_access_tokens.sql
```diff
@@ -0,0 +1,55 @@
...
+    revoked_at TIMESTAMP WITH TIME ZONE,
+    last_active_at TIMESTAMP WITH TIME ZONE,
+    last_active_ip INET
+);
+
+-- Individual tokens.
+CREATE TABLE personal_access_tokens (
+    -- The family this access token belongs to.
+    personal_access_token_id UUID NOT NULL PRIMARY KEY,
+    personal_session_id UUID NOT NULL REFERENCES personal_sessions(personal_session_id),
```
sandhose:
```suggestion
    personal_access_token_id UUID NOT NULL PRIMARY KEY,
    -- The family this access token belongs to.
    personal_session_id UUID NOT NULL REFERENCES personal_sessions(personal_session_id),
```

crates/handlers/src/oauth2/introspection.rs
```diff
@@ -625,6 +625,11 @@ pub(crate) async fn post(
                 device_id: session.device.map(Device::into),
             }
         }
+
+        TokenType::PersonalAccessToken => {
+            // TODO
+            return Err(RouteError::UnknownToken(TokenType::PersonalAccessToken));
```
sandhose, line 631: A few things I have in mind: this needs to check that both the actor and the owner are still active.

Deactivating (not locking) users should revoke all their tokens

Validating those tokens should also be done for requests to the GraphQL endpoint (optional maybe, we want to deprecate 'admin' access to the GraphQL API) and for the admin API access

crates/iana/src/oauth.rs
reivilibre: fair enough, will just remove

crates/storage-pg/migrations/20250924132713_personal_access_tokens.sql
reivilibre: added support for OAuth2 clients as owners in 3df22983378116a7acd90d8ad6ecfb37a3ab35ce

crates/handlers/src/oauth2/introspection.rs
reivilibre: Noted down. Given our plans with GraphQL I'm tempted to say that we just don't accept PATs there; cuts off one more opportunity for people to get locked into a deprecated API.

crates/data-model/src/personal/session.rs
```diff
@@ -0,0 +1,132 @@
...
+    }
+}
+
+#[derive(Debug, Clone, PartialEq, Eq, Serialize)]
+pub struct PersonalSession {
+    pub id: Ulid,
+    pub state: SessionState,
+    pub owner: PersonalSessionOwner,
+    pub actor_user_id: Ulid,
+    pub human_name: String,
```
sandhose, line 79: I'm wondering if this should be optional… but I guess most tools force you to have a name

crates/data-model/src/personal/session.rs
```diff
@@ -0,0 +1,132 @@
...
+        }
+    }
+}
+
+#[derive(Debug, Clone, PartialEq, Eq, Serialize)]
+pub struct PersonalSession {
+    pub id: Ulid,
+    pub state: SessionState,
+    pub owner: PersonalSessionOwner,
+    pub actor_user_id: Ulid,
```
sandhose, line 78: I'm now thinking that potentially you could have no actor (like you don't need one when talking to the MAS admin API) but then the main reason for you to use PATs (and not just the client credentials grant) is if you need to access Synapse and MAS, so probably not worth supporting

crates/storage-pg/src/personal/access_token.rs
```diff
@@ -0,0 +1,216 @@
...
+            %session.id,
+        ),
+        err,
+    )]
+    async fn add(
+        &mut self,
+        rng: &mut (dyn RngCore + Send),
+        clock: &dyn Clock,
+        session: &PersonalSession,
+        access_token: String,
```
sandhose: we're hashing under the hood and not returning it in the data model, so maybe take a `&str`? especially since in the admin API/GraphQL handle, you'll likely

 - t = gen()
 - repo.add(t)
 - return t

so taking a reference would avoid an unnecessary clone

crates/data-model/src/personal/session.rs
reivilibre: That was my thought really; other tools make you put in a name and it seems self-sabotaging to not have a label for your PATs, so it seems like a good idea to just have one for simplicity.

crates/data-model/src/personal/session.rs
reivilibre: Also thought the same really. Seems like extra kerfuffle for no apparent benefit; could be relaxed later if it turns out to be useful though

crates/storage-pg/src/personal/access_token.rs
reivilibre: → 277e8e84b019

PR #5135, opened 2025-10-10
Expose end_session endpoint to perform rp_initiated_logout

mcalinghee:
### Contexte
Add `end_session` endpoint to perform a RP-Initiated Logout according to the spec : https://openid.net/specs/openid-connect-rpinitiated-1_0.html

This endpoint will allow clients to disconnect from the MAS when they perform a log out.

NOTE : 
- no logout of upstream oauth2 session is performed

### DIFF
--- a/crates/handlers/src/oauth2/end_session.rs
+++ b/crates/handlers/src/oauth2/end_session.rs
@@ -5,9 +5,10 @@
 // Please see LICENSE files in the repository root for full details.
 use axum::{
     Json,
-    extract::{Query, State},
+    extract::State,
     response::{IntoResponse, Redirect, Response},
 };
+use axum_extra::extract::Query;
 use hyper::StatusCode;
 use mas_axum_utils::{SessionInfoExt, cookies::CookieJar, record_error};
 use mas_data_model::{BoxClock, BoxRng, Clock};
--- a/crates/handlers/src/oauth2/end_session.rs
+++ b/crates/handlers/src/oauth2/end_session.rs
@@ -1,0 +1,210 @@
+// Copyright 2024, 2025 New Vector Ltd.
+// Copyright 2023, 2024 The Matrix.org Foundation C.I.C.
+//
+// SPDX-License-Identifier: AGPL-3.0-only OR LicenseRef-Element-Commercial
+// Please see LICENSE files in the repository root for full details.
+use axum::{
+    Json,
+    extract::{Query, State},
+    response::{IntoResponse, Redirect, Response},
+};
+use hyper::StatusCode;
+use mas_axum_utils::{SessionInfoExt, cookies::CookieJar, record_error};
+use mas_data_model::{BoxClock, BoxRng, Clock};
+use mas_keystore::Keystore;
+use mas_oidc_client::{
+    error::IdTokenError,
+    requests::jose::{JwtVerificationData, verify_id_token},
+};
+use mas_router::UrlBuilder;
+use mas_storage::{
+    BoxRepository, RepositoryAccess,
+    queue::{QueueJobRepositoryExt as _, SyncDevicesJob},
+    user::BrowserSessionRepository,
+};
+use oauth2_types::errors::{ClientError, ClientErrorCode};
+use serde::{Deserialize, Serialize};
+use thiserror::Error;
+
+use crate::{BoundActivityTracker, impl_from_error_for_route};
+
+#[derive(Debug, Deserialize, Serialize)]
+pub(crate) struct EndSessionParam {
+    id_token_hint: String,
+    post_logout_redirect_uri: String,
+}
+
+#[derive(Debug, Error)]
+pub(crate) enum RouteError {
+    #[error(transparent)]
+    Internal(Box<dyn std::error::Error + Send + Sync + 'static>),
+
+    #[error("bad request")]
+    BadRequest,
+
+    #[error("client not found")]
+    ClientNotFound,
+
+    #[error("client is unauthorized")]
+    UnauthorizedClient,
+
+    // #[error("unsupported token type")]
+    // UnsupportedTokenType,
+    #[error("unknown token")]
+    UnknownToken,
+}
+
+impl_from_error_for_route!(mas_storage::RepositoryError);
+
+impl IntoResponse for RouteError {
+    fn into_response(self) -> Response {
+        let sentry_event_id = record_error!(self, Self::Internal(_));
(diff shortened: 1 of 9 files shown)


crates/handlers/src/oauth2/end_session.rs
```diff
@@ -0,0 +1,211 @@
+// Copyright 2024, 2025 New Vector Ltd.
+// Copyright 2023, 2024 The Matrix.org Foundation C.I.C.
```
sandhose:
```suggestion
// Copyright 2025 New Vector Ltd.
```

crates/handlers/src/oauth2/end_session.rs
```diff
@@ -0,0 +1,211 @@
...
+    BadRequest,
+
+    #[error("client not found")]
+    ClientNotFound,
+
+    #[error("client is unauthorized")]
+    UnauthorizedClient,
+
+    // #[error("unsupported token type")]
+    // UnsupportedTokenType,
```
sandhose:
```suggestion
```

crates/handlers/src/oauth2/end_session.rs
```diff
@@ -0,0 +1,211 @@
...
+            )
+                .into_response(),
+
+            Self::ClientNotFound => (
+                StatusCode::UNAUTHORIZED,
+                Json(ClientError::from(ClientErrorCode::InvalidClient)),
+            )
+                .into_response(),
+
+            // Self::ClientNotAllowed |
```
sandhose:
```suggestion
```

crates/handlers/src/oauth2/end_session.rs
```diff
@@ -0,0 +1,211 @@
...
+                Json(ClientError::from(ClientErrorCode::UnauthorizedClient)),
+            )
+                .into_response(),
+
+            // Self::UnsupportedTokenType => (
+            //     StatusCode::BAD_REQUEST,
+            //     Json(ClientError::from(ClientErrorCode::UnsupportedTokenType)),
+            // )
+            //     .into_response(),
+
```
sandhose:
```suggestion
```

crates/handlers/src/oauth2/end_session.rs
```diff
@@ -0,0 +1,211 @@
...
+        .browser_session()
+        .lookup(browser_session_id)
+        .await?
+        .ok_or(RouteError::BadRequest)?;
+
+    let oauth_session = repo
+        .oauth2_session()
+        .find_by_browser_session(browser_session.id)
+        .await?
+        .ok_or(RouteError::BadRequest)?;
```
sandhose, line 126: You can have many OAuth session started from a single browser session, so this way is not the right approach here.

Instead, best is to decode the JWT (but not necessarily verify it, maybe `verify_id_token` should be changed a little bit?) to get the claims we need and go from there? I'm unclear what we should be checking exactly on that JWT though, I don't know whether there is something that would let us link it to the current browser session or not 🤔

crates/handlers/src/oauth2/end_session.rs
```diff
@@ -0,0 +1,211 @@
...
+            .ok_or(RouteError::UnknownToken)?;
+
+        // Schedule a job to sync the devices of the user with the homeserver
+        repo.queue_job()
+            .schedule_job(&mut rng, &clock, SyncDevicesJob::new(&user))
+            .await?;
+    }
+
+    // Now that we checked everything, we can end the session.
+    repo.oauth2_session().finish(&clock, oauth_session).await?;
```
sandhose, line 185: Not sure about that semantic. I don't know whether we should finish the oauth session at this point or let the client still call the revocation endpoint.

Either way, we need a better way to find the right OAuth session, and I think this should be done by injecting the OAuth session ID in the id_token `sid` claim when we give out the id_token

crates/handlers/src/oauth2/end_session.rs
```diff
@@ -0,0 +1,211 @@
...
+        .finish(&clock, browser_session)
+        .await?;
+
+    repo.save().await?;
+
+    // We always want to clear out the session cookie, even if the session was
+    // invalid
+    let cookie_jar = cookie_jar.update_session_info(&session_info.mark_session_ended());
+
+    Ok((cookie_jar, Redirect::to(&params.post_logout_redirect_uri)).into_response())
```
sandhose, line 200: The spec says we really should be validating the post_logout_redirect_uri, it should be registered by the client first

> The value MUST have been previously registered with the OP, either using the post_logout_redirect_uris Registration parameter or via another mechanism.

crates/router/src/endpoints.rs
```diff
@@ -155,6 +155,14 @@ impl SimpleRoute for OAuth2AuthorizationEndpoint {
     const PATH: &'static str = "/authorize";
 }
 
+/// `POST /oauth2/end_session`
+#[derive(Default, Debug, Clone)]
+pub struct OAuth2EndSession;
+
+impl SimpleRoute for OAuth2EndSession {
+    const PATH: &'static str = "/oauth2/end_session";
```
sandhose: nit: I don't like underscores in URLs :)
```suggestion
    const PATH: &'static str = "/oauth2/end-session";
```

crates/handlers/src/oauth2/end_session.rs
```diff
@@ -0,0 +1,211 @@
...
+                .into_response(),
+
+            // Self::UnsupportedTokenType => (
+            //     StatusCode::BAD_REQUEST,
+            //     Json(ClientError::from(ClientErrorCode::UnsupportedTokenType)),
+            // )
+            //     .into_response(),
+
+            // If the token is unknown, we still return a 200 OK response.
+            Self::UnknownToken => StatusCode::OK.into_response(),
```
sandhose, line 86: We should have a nice view in some of those cases, in the end that is going to be a redirect, potentially where it doesn't go back to the client

sandhose (changes requested):
Thanks for working on this! Generally agree with having this, but there are a few things to get right

sandhose (changes requested):
The open threads are still valid, requesting changes to pop it off the review queue

sandhose:
With no activity for a while, I'll close this to clear a bit the PR list 📉

plui29989:
Not a developer myself, but I would appreciate if anyone implemented this in MAS. 
Currently, all our users assume, when logging out of element-web, a logout in MAS is implicated...

PR #5141, opened 2025-10-14
Personal Sessions: add create, list, get, revoke, regenerate Admin APIs 

reivilibre:
Follows: #5106

This PR introduces some admin API endpoints for Personal Sessions.

* add: Creates a personal session along with its first personal access token, returning both. This is currently the only way to get a personal access token.
* get: Shows the information about a personal session
* list: Shows many personal sessions
* revoke: Revokes a personal session, so it can't be used anymore
* regenerate: Revoke the active personal access token for a session and issue a new one to replace it.

### DIFF
--- a/docs/api/spec.json
+++ b/docs/api/spec.json
@@ -999,6 +999,20 @@
           },
           {
             "in": "query",
+            "name": "filter[scope]",
+            "description": "Retrieve the items with the given scope",
+            "schema": {
+              "description": "Retrieve the items with the given scope",
+              "default": [],
+              "type": "array",
+              "items": {
+                "type": "string"
+              }
+            },
+            "style": "form"
+          },
+          {
+            "in": "query",
             "name": "filter[status]",
             "description": "Filter by session status",
             "schema": {
@@ -1028,6 +1042,17 @@
               "description": "Filter by access token expiry date",
               "type": "string",
               "format": "date-time",
+              "nullable": true
+            },
+            "style": "form"
+          },
+          {
+            "in": "query",
+            "name": "filter[expires]",
+            "description": "Filter by whether the access token has an expiry time",
+            "schema": {
+              "description": "Filter by whether the access token has an expiry time",
+              "type": "boolean",
               "nullable": true
             },
             "style": "form"
@@ -5138,6 +5163,14 @@
             "$ref": "#/components/schemas/ULID",
             "nullable": true
           },
+          "filter[scope]": {
+            "description": "Retrieve the items with the given scope",
+            "default": [],
+            "type": "array",
+            "items": {
+              "type": "string"
+            }
+          },
           "filter[status]": {
             "description": "Filter by session status",
             "$ref": "#/components/schemas/PersonalSessionStatus",
@@ -5153,6 +5186,11 @@
             "description": "Filter by access token expiry date",
             "type": "string",
             "format": "date-time",
+            "nullable": true
+          },
+          "filter[expires]": {
+            "description": "Filter by whether the access token has an expiry time",
+            "type": "boolean",
             "nullable": true
           }
         }
@@ -5230,8 +5268,6 @@
           "actor_user_id",
           "created_at",
           "human_name",
-          "owner_client_id",
-          "owner_user_id",
           "scope"
         ],
         "properties": {
@@ -5248,11 +5284,13 @@
           },
           "owner_user_id": {
             "description": "The ID of the user who owns this session (if user-owned)",
-            "$ref": "#/components/schemas/ULID"
+            "$ref": "#/components/schemas/ULID",
+            "nullable": true
           },
           "owner_client_id": {
             "description": "The ID of the `OAuth2` client that owns this session (if client-owned)",
-            "$ref": "#/components/schemas/ULID"
+            "$ref": "#/components/schemas/ULID",
+            "nullable": true
           },
           "actor_user_id": {
             "description": "The ID of the user that the session acts on behalf of",
@@ -5315,7 +5353,7 @@
           "expires_in": {
             "description": "Token expiry time in seconds. If not set, the token won't expire.",
             "type": "integer",
-            "format": "uint64",
+            "format": "uint32",
             "minimum": 0.0,
             "nullable": true
           }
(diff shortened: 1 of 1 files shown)


crates/storage/src/personal/access_token.rs
```diff
@@ -50,6 +50,22 @@ pub trait PersonalAccessTokenRepository: Send + Sync {
...
+    /// # Parameters
+    ///
+    /// * `session_id`: The ID of the session to lookup
+    ///
+    /// # Errors
+    ///
+    /// Returns [`Self::Error`] if the underlying repository fails
+    async fn find_active_for_session(
+        &mut self,
+        session_id: Ulid,
```
sandhose: (note for later to check where it's being used but) when possible I like to do something like `session: &Session` instead for a ressemblance of type-safety in lookups. Of course sometimes it's not convenient if you don't have the 'parent' object loaded yet

crates/storage/src/personal/session.rs
```diff
@@ -150,6 +154,8 @@ pub struct PersonalSessionFilter<'a> {
     scope: Option<&'a Scope>,
     last_active_before: Option<DateTime<Utc>>,
     last_active_after: Option<DateTime<Utc>>,
+    expires_before: Option<DateTime<Utc>>,
```
sandhose, line 157: Could we also have an `expires=true|false` filter to get tokens with (no) expiry?

crates/storage-pg/src/oauth2/client.rs
```diff
@@ -811,6 +811,49 @@ impl OAuth2ClientRepository for PgOAuth2ClientRepository<'_> {
             .await?;
         }
 
+        // Delete any personal access tokens & sessions owned
```
sandhose, line 814: Neat.

crates/handlers/src/admin/model.rs
```diff
@@ -771,3 +778,179 @@ impl UpstreamOAuthProvider {
...
+    id: Ulid,
+
+    /// When the session was created
+    created_at: DateTime<Utc>,
+
+    /// When the session was revoked, if applicable
+    revoked_at: Option<DateTime<Utc>>,
+
+    /// The ID of the user who owns this session (if user-owned)
+    #[schemars(with = "super::schema::Ulid")]
```
sandhose: should be `#[schemars(with = "Option<super::schema::Ulid>")]` I think

crates/handlers/src/admin/v1/personal_sessions/regenerate.rs
```diff
@@ -0,0 +1,253 @@
...
+        .find_active_for_session(session_id)
+        .await?;
+    let Some(old_token) = old_token_opt else {
+        // This shouldn't happen
+        error!("session is supposedly valid but had no access token");
+        return Err(RouteError::SessionNotValid);
+    };
+
+    // Calculate the expiry time of the new token, defaulting
+    // to the token's previous lifetime
```
sandhose: Unsure if that's expected behaviour?

crates/handlers/src/admin/v1/personal_sessions/regenerate.rs
```diff
@@ -0,0 +1,253 @@
...
+}
+
+/// # JSON payload for the `POST /api/admin/v1/personal-sessions/{id}/regenerate` endpoint
+#[derive(Deserialize, JsonSchema)]
+#[serde(rename = "RegeneratePersonalSessionRequest")]
+pub struct Request {
+    /// Token expiry time in seconds.
+    /// If not set, the token will default to the same lifetime as when
+    /// originally issued.
+    expires_in: Option<u64>,
```
sandhose: `u32` is enough for 136 years, presumably this is enough :D 
Also avoids an issue where the generated TS client uses a bigint for u64 :(

crates/handlers/src/admin/v1/personal_sessions/list.rs
```diff
@@ -0,0 +1,540 @@
...
+    #[serde(rename = "filter[status]")]
+    status: Option<PersonalSessionStatus>,
+
+    /// Filter by access token expiry date
+    #[serde(rename = "filter[expires_before]")]
+    expires_before: Option<DateTime<Utc>>,
+
+    /// Filter by access token expiry date
+    #[serde(rename = "filter[expires_after]")]
+    expires_after: Option<DateTime<Utc>>,
```
sandhose, line 81: could we also get a filter for:

 - a particular scope (useful to give you 'synapse admin tokens')
 - `expired=true|false`, especially to find tokens with no expiration time

crates/handlers/src/admin/model.rs
reivilibre: hm hm, I think I got this from `CompatSession`, which also doesn't have the `Option` in the `with` type. Any idea if this is on purpose?

crates/handlers/src/admin/v1/personal_sessions/regenerate.rs
reivilibre: Fair enough, maybe this should be a client-side concern, since I guess nothing stops the client calculating this

crates/storage/src/personal/access_token.rs
reivilibre: good idea, done in 78b010d3f6e461143cbc89ed5a65b38f52f88e66

crates/storage/src/personal/session.rs
reivilibre: → 52c04c173f46070cd2084769a520e438d6f0709d

crates/handlers/src/admin/model.rs
reivilibre: → 6102a4b672ff75940e218dd346f001ce69df9a0f

crates/handlers/src/admin/v1/personal_sessions/regenerate.rs
reivilibre: → ba9fc3513cb6084c653a49c2e5d9e23c61ab8ed3

crates/handlers/src/admin/v1/personal_sessions/regenerate.rs
reivilibre: → ba9fc3513cb6084c653a49c2e5d9e23c61ab8ed3

crates/handlers/src/admin/v1/personal_sessions/list.rs
reivilibre: scope filter is in d516b3dee368c2d3a5a8b187bdfc777c2fe7f1a4

crates/handlers/src/admin/model.rs
sandhose: > Any idea if this is on purpose?

definitely not on purpose

sandhose (approved):
Alright, take your tick and run!

PR #5171, opened 2025-10-20
Support personal access tokens for introspection

reivilibre:
Follows: https://github.com/element-hq/matrix-authentication-service/pull/5106

This PR makes it so that you can present a personal access token (`mpt_` prefix) at introspection and have it accepted.

This means personal access tokens can be presented to Synapse and used on the client-server API.

### DIFF
--- a/crates/oauth2-types/src/requests.rs
+++ b/crates/oauth2-types/src/requests.rs
@@ -807,6 +807,7 @@ pub struct IntrospectionResponse {
     pub jti: Option<String>,
 
     /// MAS extension: explicit device ID
+    /// Only used for compatibility access and refresh tokens.
     pub device_id: Option<String>,
 }
 


crates/handlers/src/activity_tracker/mod.rs
```diff
@@ -115,7 +117,7 @@ impl ActivityTracker {
     pub async fn record_personal_access_token_session(
         &self,
         clock: &dyn Clock,
-        session: &Session,
+        session: &PersonalSession,
```
sandhose, line 120: 🙈

crates/handlers/src/oauth2/introspection.rs
```diff
@@ -627,8 +643,94 @@ pub(crate) async fn post(
...
+                    }
+
+                    None
+                }
+                PersonalSessionOwner::OAuth2Client(owner_client_id) => {
+                    let owner_client = repo
+                        .oauth2_client()
+                        .lookup(owner_client_id)
+                        .await?
+                        .ok_or(RouteError::CantLoadOAuth2Client(owner_client_id))?;
```
sandhose, line 695: I mean in theory we have a FK preventing this but sure, doesn't hurt :)

reivilibre: indeed, though we do this already for some things, it felt easier to go with the flow somewhat

PR #5181, opened 2025-10-22
Revoke personal sessions when users are deactivated

reivilibre:

Revoke both personal sessions that are owned by, and acting as, the deactivated user.

Owned by because: it doesn't make sense for a deactivated user to be able to control themselves or other users, so them having active personal sessions is just confusing.

Acting as because: current precedent is that deactivated users are not controllable, even by admins.
To uphold this, the admin API is also fixed to stop allowing the creation of personal sessions for deactivated users.

### DIFF
--- a/crates/storage/src/personal/session.rs
+++ b/crates/storage/src/personal/session.rs
@@ -89,8 +89,6 @@ pub trait PersonalSessionRepository: Send + Sync {
 
     /// Revoke all the [`PersonalSession`]s matching the given filter.
     ///
-    /// This will also revoke the relevant personal access tokens.
-    ///
     /// Returns the number of sessions affected
     ///
     /// # Parameters


crates/handlers/src/admin/v1/personal_sessions/add.rs
```diff
@@ -46,6 +49,7 @@ impl IntoResponse for RouteError {
         let status = match self {
             Self::Internal(_) => StatusCode::INTERNAL_SERVER_ERROR,
             Self::UserNotFound => StatusCode::NOT_FOUND,
+            Self::UserDeactivated => StatusCode::GONE,
```
sandhose, line 53: sure why not :D

crates/handlers/src/admin/v1/personal_sessions/add.rs
```diff
@@ -114,6 +118,10 @@ pub async fn handler(
         .await?
         .ok_or(RouteError::UserNotFound)?;
 
+    if actor_user.deactivated_at.is_some() {
```
sandhose: maybe check just `is_valid` (assuming it does the right thing) so that we also check for locked status?

crates/storage-pg/src/personal/session.rs
```diff
@@ -357,6 +357,68 @@ impl PersonalSessionRepository for PgPersonalSessionRepository<'_> {
...
+        clock: &dyn Clock,
+        filter: PersonalSessionFilter<'_>,
+    ) -> Result<usize, Self::Error> {
+        let revoked_at = clock.now();
+
+        let (sql, arguments) = Query::update()
+            .table(PersonalSessions::Table)
+            .value(PersonalSessions::RevokedAt, revoked_at)
+            .and_where(
+                Expr::col((PersonalSessions::Table, PersonalSessions::PersonalSessionId))
```
sandhose, line 379: That's a chunky subquery. I assume you need this because of the JOIN; could you  stick in a comment explaining that? Also we could be doing that in a CTE, but not convinced the code would be clearer

crates/storage/src/personal/session.rs
```diff
@@ -87,6 +87,26 @@ pub trait PersonalSessionRepository: Send + Sync {
         personal_session: PersonalSession,
     ) -> Result<PersonalSession, Self::Error>;
 
+    /// Revoke all the [`PersonalSession`]s matching the given filter.
+    ///
+    /// This will also revoke the relevant personal access tokens.
```
sandhose: Does it really?

crates/tasks/src/user.rs
```diff
@@ -80,6 +81,36 @@ impl RunnableJob for DeactivateUserJob {
             .map_err(JobError::retry)?;
         info!(affected = n, "Killed all compatibility sessions for user");
 
+        let n = repo
+            .personal_session()
+            .revoke_bulk(
+                clock,
+                PersonalSessionFilter::new()
+                    .for_actor_user(&user)
+                    .active_only(),
```
sandhose, line 90: love those filters 🙇

crates/storage/src/personal/session.rs
reivilibre: ah nope, I intended to, but it wasn't as straightforward as I expected. I think we'd have to buffer the list of session IDs to be revoked in memory first and issue 2 deletes subsequently, for a total of 3 queries.

It probably wouldn't be terrible, not sure what you prefer?

sandhose: we have the same problem in regular access tokens I think, I'm fine with just revoking the sessions. I assume that we'll also want to remove tokens from revoked sessions over time, like we do for access tokens maybe?

crates/handlers/src/admin/v1/personal_sessions/add.rs
reivilibre: now using `is_valid_actor`

crates/storage-pg/src/personal/session.rs
reivilibre: added some comments

yeah, it's a bit chunky, the sea-query doesn't help with that verbosity either, but hopefully the new comments make the intent clear

PR #5182, opened 2025-10-22
When adding or revoking personal sessions, schedule device synchronisation

reivilibre:

### DIFF
--- a/crates/handlers/src/admin/v1/personal_sessions/add.rs
+++ b/crates/handlers/src/admin/v1/personal_sessions/add.rs
@@ -11,7 +11,7 @@ use axum::{Json, extract::State, response::IntoResponse};
 use chrono::Duration;
 use hyper::StatusCode;
 use mas_axum_utils::record_error;
-use mas_data_model::{BoxRng, Device, TokenType, personal::session::PersonalSessionOwner};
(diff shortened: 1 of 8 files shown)


crates/handlers/src/admin/v1/personal_sessions/add.rs
```diff
@@ -144,6 +145,14 @@ pub async fn handler(
         )
         .await?;
 
+    if session.has_device() {
+        // If the session has a device, then we are now
+        // creating a device and should schedule a device sync.
+        repo.queue_job()
+            .schedule_job(&mut rng, &clock, SyncDevicesJob::new(&actor_user))
+            .await?;
+    }
```
sandhose, line 167: Other session-creation logic do it synchronously to avoid having to deal race conditions on the consumer side

I think we should do the same here. It's fine to simplify here (as long as there is a comment about that I guess) and do it within the transaction, as I don't expect this to be called so often, so it shouldn't really take off all the db connections from the pool

reivilibre: done in 95bc20e4408ad67b965d752f42246e9c1da4b985

PR #5183, opened 2025-10-22
Accept personal access tokens on the Admin API

reivilibre:

Allow the Admin API to be used by providing a personal access token whose session has the correct scope.

### DIFF
--- a/crates/handlers/src/admin/call_context.rs
+++ b/crates/handlers/src/admin/call_context.rs
@@ -255,26 +255,30 @@ where
                 .lookup(user_id)
                 .await?
                 .ok_or_else(|| Rejection::LoadUser(user_id))?;
+
+            match session {
+                CallerSession::OAuth2Session(_) => {
+                    // For OAuth2 sessions: check that the user is valid enough
+                    // to be a user.
+                    if !user.is_valid() {
+                        return Err(Rejection::UserLocked);
+                    }
+                }
+                CallerSession::PersonalSession(_) => {
+                    // For personal sessions: check that the actor is valid enough
+                    // to be an actor.
+                    if !user.is_valid_actor() {
+                        return Err(Rejection::UserLocked);
+                    }
+                }
+            }
+
             Some(user)
         } else {
+            // Double check we're not using a PersonalSession
+            assert!(matches!(session, CallerSession::OAuth2Session(_)));
             None
         };
-
-        if let CallerSession::PersonalSession(_) = &session {
-            // For personal sessions: check that the actor is valid enough
-            // to be an actor.
-            // unwrap: personal sessions always have an actor user
-            if !user.as_ref().unwrap().is_valid_actor() {
-                return Err(Rejection::UserLocked);
-            }
-        } else {
-            // If there is a user for this session, check that it is not locked
-            if let Some(user) = &user
-                && !user.is_valid()
-            {
-                return Err(Rejection::UserLocked);
-            }
-        }
 
         // For now, we only check that the session has the admin scope
         // Later we might want to check other route-specific scopes


crates/handlers/src/admin/call_context.rs
```diff
@@ -41,6 +42,10 @@ pub enum Rejection {
     #[error("Invalid repository operation")]
     Repository(#[from] RepositoryError),
 
+    /// The access token was not of the correct type for the Admin API
+    #[error("Invalid type of access token")]
+    InvalidAccessTokenType,
```
sandhose: This could carry the underlying error?
```suggestion
    InvalidAccessTokenType(#[source] TokenFormatError),
```

crates/handlers/src/admin/call_context.rs
```diff
@@ -154,28 +160,76 @@ where
             })?;
 
         let token = token.token();
+        let token_type = TokenType::check(token).or(Err(Rejection::InvalidAccessTokenType))?;
```
sandhose: this should probably map the 

```suggestion
        let token_type = TokenType::check(token).map_err(Rejection::InvalidAccessTokenType)?;
```

crates/handlers/src/admin/call_context.rs
```diff
@@ -154,28 +160,76 @@ where
...
+                // Look for the associated session in the database
+                let session = repo
+                    .personal_session()
+                    .lookup(token.session_id)
+                    .await?
+                    .ok_or_else(|| Rejection::LoadSession(token.session_id))?;
+
+                if !session.is_valid() {
+                    return Err(Rejection::SessionRevoked);
+                }
```
sandhose, line 216: maybe we should check if the actor & owner are both valid? Especially important to handle when they are locked

crates/handlers/src/admin/call_context.rs
```diff
@@ -154,28 +160,76 @@ where
...
-            .ok_or_else(|| Rejection::LoadSession(token.session_id))?;
-
-        // Record the activity on the session
-        activity_tracker
-            .record_oauth2_session(&clock, &session)
-            .await;
+                CallerSession::PersonalSession(session)
+            }
+            _other => {
+                return Err(Rejection::InvalidAccessTokenType);
```
sandhose: right, hence the error type not having the underlying error. `#[source]` can be optional I think?

crates/handlers/src/admin/call_context.rs
sandhose: oh nevermind we check for the user under there. Not sure about the owner though?

reivilibre: Have added a check for the owner & relaxed the check on the actor (in line with our agreement to allow locked, not deactivated, users to be 'puppetted')

crates/handlers/src/admin/call_context.rs
```diff
@@ -41,6 +45,14 @@ pub enum Rejection {
     #[error("Invalid repository operation")]
     Repository(#[from] RepositoryError),
 
+    /// The access token was not of the correct type for the Admin API
+    #[error("Invalid type of access token")]
+    InvalidAccessTokenType(
+        #[source]
+        #[from]
+        Option<TokenFormatError>,
```
sandhose: I think `#[from]` implies `#[source]`

crates/handlers/src/admin/call_context.rs
```diff
@@ -186,24 +264,25 @@ where
...
-            return Err(Rejection::SessionRevoked);
-        }
-
-        if !token.is_valid(clock.now()) {
-            return Err(Rejection::TokenExpired);
+        if let CallerSession::PersonalSession(_) = &session {
+            // For personal sessions: check that the actor is valid enough
+            // to be an actor.
+            // unwrap: personal sessions always have an actor user
+            if !user.as_ref().unwrap().is_valid_actor() {
```
sandhose: maybe more readable this way?

```rust
        // Load the user if there is one
        let user = if let Some(user_id) = session.user_id() {
            let user = repo
                .user()
                .lookup(user_id)
                .await?
                .ok_or_else(|| Rejection::LoadUser(user_id))?;
            
            match session {
                CallerSession::OAuth2Session(_) => {
                    // For OAuth2 sessions: check that the user is valid enough
                    // to be a user.
                    if !user.is_valid() {
                        return Err(Rejection::UserLocked);
                    }
                }
                CallerSession::PersonalSession(_) => {
                    // For personal sessions: check that the actor is valid enough
                    // to be an actor.
                    if !user.is_valid_actor() {
                        return Err(Rejection::UserLocked);
                    }
                }
            }
            
            Some(user)
        } else {
            // Double check we're not using a PersonalSession
            assert!(matches!(session, CallerSession::OAuth2Session(_)));
            None
        };
```

PR #5201, opened 2025-10-24
templates: Be strict about undefined variables

reivilibre:
Currently the template engine is lax about undefined variables, meaning typos become silent rendering issues.

By switching to the strict undefined variable behaviour, we can make these issues noisier and easier to notice.

This also makes the `templates check` command more useful.

### DIFF
--- a/crates/cli/src/commands/server.rs
+++ b/crates/cli/src/commands/server.rs
@@ -160,8 +160,14 @@ impl Options {
         )?;
 
         // Load and compile the templates
-        let templates =
-            templates_from_config(&config.templates, &site_config, &url_builder).await?;
+        let templates = templates_from_config(
+            &config.templates,
+            &site_config,
+            &url_builder,
+            // Don't use strict mode in production yet
+            false,
+        )
+        .await?;
         shutdown.register_reloadable(&templates);
 
         let http_client = mas_http::reqwest_client();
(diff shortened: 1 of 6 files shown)


templates/components/footer.html
```diff
@@ -7,27 +7,27 @@
...
         </a>
       {%- endif -%}
 
-      {%- if branding.policy_uri and branding.tos_uri -%}
+      {%- if branding.policy_uri is defined and branding.tos_uri is defined -%}
         <div class="separator" aria-hidden="true">•</div>
       {%- endif -%}
 
-      {%- if branding.tos_uri -%}
+      {%- if branding.tos_uri is defined -%}
```
sandhose: This scares me a little bit, we don't have 100% coverage of templates with the samples, so we might have missed some spots. What about using `SemiStrict` instead, which allows for `if undefined`?

reivilibre: Changed my mind on this: how about using `is none` instead? So we never make fields undefined, just none/nullable. I think this would offer even more typo resistance than `is_defined`

sandhose: Agreed on using none instead of undefined, but the point that this PR might break other stuff still stands

reivilibre: Yeah OK, this is going to be gnarly, I've downgraded in https://github.com/element-hq/matrix-authentication-service/commit/20a13f49c7d245d00fe2a44cadd0b0a99855afd4

PR #5202, opened 2025-10-24
templates check command: Allow saving rendered samples

reivilibre:
Gives template samples a stable identifier that we can use to track them across different sample render invocations.

Then allow saving rendered samples to a directory using the `--out-dir` flag.

Note: rendered pages contain unstable data like subresource integrity and asset hashes.
A future PR will add a feature to strip these out so the renders can be compared across different versions of MAS.

---

Commit-by-commit review.

### DIFF
--- a/crates/cli/src/commands/templates.rs
+++ b/crates/cli/src/commands/templates.rs
@@ -72,8 +72,13 @@ impl Options {
                     &account_config,
                     &captcha_config,
                 )?;
-                let templates =
-                    templates_from_config(&template_config, &site_config, &url_builder).await?;
+                let templates = templates_from_config(
+                    &template_config,
+                    &site_config,
+                    &url_builder, // Use strict mode in template checks
+                    true,
+                )
+                .await?;
                 let all_renders = templates.check_render(clock.now(), &mut rng)?;
 
                 if let Some(out_dir) = out_dir {
--- a/crates/templates/src/lib.rs
+++ b/crates/templates/src/lib.rs
@@ -20,7 +20,7 @@ use camino::{Utf8Path, Utf8PathBuf};
 use mas_i18n::Translator;
 use mas_router::UrlBuilder;
 use mas_spa::ViteManifest;
-use minijinja::Value;
+use minijinja::{UndefinedBehavior, Value};
 use rand::Rng;
 use serde::Serialize;
 use thiserror::Error;
@@ -75,6 +75,9 @@ pub struct Templates {
     vite_manifest_path: Utf8PathBuf,
     translations_path: Utf8PathBuf,
     path: Utf8PathBuf,
+    /// Whether template rendering is in strict mode (for testing,
+    /// until this can be rolled out in production.)
+    strict: bool,
 }
 
 /// There was an issue while loading the templates
@@ -155,6 +158,7 @@ impl Templates {
         translations_path: Utf8PathBuf,
         branding: SiteBranding,
         features: SiteFeatures,
+        strict: bool,
     ) -> Result<Self, TemplateLoadingError> {
         let (translator, environment) = Self::load_(
             &path,
@@ -163,6 +167,7 @@ impl Templates {
             &translations_path,
             branding.clone(),
             features,
+            strict,
         )
         .await?;
         Ok(Self {
@@ -174,6 +179,7 @@ impl Templates {
             translations_path,
             branding,
             features,
+            strict,
         })
     }
 
@@ -184,6 +190,7 @@ impl Templates {
         translations_path: &Utf8Path,
         branding: SiteBranding,
         features: SiteFeatures,
+        strict: bool,
     ) -> Result<(Arc<Translator>, Arc<minijinja::Environment<'static>>), TemplateLoadingError> {
         let path = path.to_owned();
         let span = tracing::Span::current();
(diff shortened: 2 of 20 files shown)


crates/templates/src/lib.rs
```diff
@@ -383,7 +386,7 @@ register_templates! {
     pub fn render_recovery_disabled(WithLanguage<EmptyContext>) { "pages/recovery/disabled.html" }
 
     /// Render the form used by the `form_post` response mode
-    pub fn render_form_post<T: Serialize>(WithLanguage<FormPostContext<T>>) { "form_post.html" }
+    pub fn render_form_post<#[sample(EmptyContext)] T: Serialize>(WithLanguage<FormPostContext<T>>) { "form_post.html" }
```
sandhose, line 407: not 100% convinced by this syntax but also 🤷

crates/cli/src/commands/templates.rs
```diff
@@ -67,7 +74,39 @@ impl Options {
...
+                let all_renders = templates.check_render(clock.now(), &mut rng)?;
+
+                if let Some(out_dir) = out_dir {
+                    // Save renders to disk.
+                    if out_dir.exists() {
+                        let mut read_dir =
+                            tokio::fs::read_dir(&out_dir).await.with_context(|| {
+                                format!("could not read {out_dir} to check it's empty")
+                            })?;
+                        while let Some(x) = read_dir.next_entry().await? {
```
sandhose: `s/while/if/`? plus even `if ….is_some()`?

sandhose: nevermind, that was fixed in the next commit :)

crates/templates/src/context.rs
```diff
@@ -105,21 +106,72 @@ pub trait TemplateContext: Serialize {
...
         Self: Sized;
 }
 
+#[derive(Clone, Debug, PartialEq, Eq, PartialOrd, Ord)]
+pub struct SampleIdentifier {
+    /// A stable locale identifier.
+    pub locale: Option<String>,
+
+    /// A stable identifier for the session that was used in this sample.
+    pub session_index: Option<usize>,
```
sandhose: little bit confused on what a session is yet?

crates/templates/src/context.rs
```diff
@@ -105,21 +106,72 @@ pub trait TemplateContext: Serialize {
...
+        now: chrono::DateTime<Utc>,
+        rng: &mut impl Rng,
+        locales: &[DataLocale],
+    ) -> BTreeMap<SampleIdentifier, Self>
     where
         Self: Sized;
 }
 
+#[derive(Clone, Debug, PartialEq, Eq, PartialOrd, Ord)]
+pub struct SampleIdentifier {
```
sandhose, line 119: Hmm, this feels like it could be a `Vec<(Key, Value)>` instead? (with `Key` probably being `&'static str` and `Value` being a `String`?)

and then the various stages push in that stack, e.g. at the end you get `[("lang", "en"), ("user", "alice"), ("index", "1")]`, and when you reconstruct the file name you get `lang:en-user:alice-index:1`?

On why a Vec and not a btreemap: it allows duplicate key, and probably that the various levels of encapsulation will have some semantic meaning

crates/cli/src/commands/templates.rs
```diff
@@ -92,19 +92,49 @@ impl Options {
...
+                            let mut s = String::new();
+                            if let Some(session_index) = sample_identifier.session_index {
+                                write!(s, "-session{session_index}")?;
+                            }
+                            write!(s, "-sample{}", sample_identifier.index)?;
+                            s
+                        };
+
+                        let locale_dir = if let Some(locale) = &sample_identifier.locale {
+                            out_dir.join(locale)
```
sandhose: not convinced that we should do one directory per locale

crates/templates/src/lib.rs
reivilibre: I mean, it's valid macro syntax in normal rust, so it upholds the 'we try to make valid Rust syntax' thing earlier ;D

crates/cli/src/commands/templates.rs
reivilibre: I thought it'd make it easier for a German deployment to only care about the German outputs, for example. Maybe there are better ways of doing that.

crates/templates/src/context.rs
reivilibre: Have changed to this approach. Example filename is `pages_upstream_oauth2_suggest_link-index:0-browser-session:0-locale:fr.html`. Are we happy with that?

Though I am not sure `:` is allowed on Windows, maybe it'd be better to find something else.

crates/templates/src/context.rs
reivilibre: Now we call this "browser-session" as part of the rework below

crates/cli/src/commands/templates.rs
sandhose: feels like the kind of post-processing a consumer could do themselves?

crates/templates/src/context.rs
sandhose: you might be right, `=` would also work I suppose

crates/cli/src/commands/templates.rs
```diff
@@ -67,7 +74,49 @@ impl Options {
...
+                    }
+
+                    for ((template, sample_identifier), template_render) in &all_renders {
+                        let (template_filename_base, template_ext) =
+                            template.rsplit_once('.').unwrap_or((template, "txt"));
+                        let template_filename_base = template_filename_base.replace('/', "_");
+
+                        // Make a string like:
+                        // - `-sample1`
+                        // - `-session2-sample1`
```
sandhose: This should be updated

crates/templates/src/context.rs
reivilibre: Yeah, let's go with this, just on the off chance it actually bites someone.

PR #5214, opened 2025-10-30
templates check: Add `--stabilise` flag to make renders reproducible

reivilibre:


crates/spa/src/vite.rs
```diff
@@ -47,6 +47,48 @@ pub struct Manifest {
...
+    #[allow(clippy::missing_panics_doc)]
+    pub fn sample() -> Self {
+        let mut inner = HashMap::new();
+
+        for name in &[
+            "src/shared.css",
+            "src/templates.css",
+            "src/main.tsx",
+            "src/swagger.ts",
+        ] {
```
sandhose: I don't particularly like that this list is hard-coded like that, especially since you might have a fork of the templates & assets that include extra assets handled by the vite build pipeline?

Instead, if feasible I would have a 'fake' version of include_assets that doesn't use the manifest at all?

crates/cli/src/commands/templates.rs
```diff
@@ -59,9 +66,17 @@ impl Options {
                 let captcha_config = CaptchaConfig::extract_or_default(figment)
                     .map_err(anyhow::Error::from_boxed)?;
 
-                let clock = SystemClock::default();
-                // XXX: we should disallow SeedableRng::from_entropy
-                let mut rng = rand_chacha::ChaChaRng::from_entropy();
+                let now = if stabilise {
+                    DateTime::from_timestamp_secs(0).unwrap()
```
sandhose: Could we use a more believable timestamp so that the IDs look like they are not all zeros?

crates/templates/src/context/captcha.rs
```diff
@@ -58,9 +59,9 @@ impl<T> WithCaptcha<T> {
 }
 
 impl<T: TemplateContext> TemplateContext for WithCaptcha<T> {
-    fn sample(
+    fn sample<R: Rng + Clone>(
```
sandhose: hmmmm a cloneable Rng is a bit suspicious at this stage? 🤔

crates/templates/src/macros.rs
```diff
@@ -78,15 +78,18 @@ macro_rules! register_templates {
             /// # Errors
             ///
             /// Returns an error if any template fails to render with any of the sample.
-            pub(crate) fn all(templates: &Templates, now: chrono::DateTime<chrono::Utc>, rng: &mut impl rand::Rng) -> anyhow::Result<::std::collections::BTreeMap<(&'static str, SampleIdentifier), String>> {
+            pub(crate) fn all<R: Rng + Clone>(templates: &Templates, now: chrono::DateTime<chrono::Utc>, rng: &R) -> anyhow::Result<::std::collections::BTreeMap<(&'static str, SampleIdentifier), String>> {
```
sandhose, line 81: Instead of requiring a cloneable RNG here, we could seed a ChaChaRng out of the rng passed as parameter and clone it within the function. That would be a lot less suspicious :p

sandhose (changes requested):

crates/templates/src/context/captcha.rs
reivilibre: It's only samples :D

crates/templates/src/macros.rs
reivilibre: I mean, either one is suspicious, but at least this one is a bit less prolific so sure.

crates/spa/src/vite.rs
sandhose: That was done and I like how it was done

Issue #4339, opened 2025-04-03
Ability to limit number of devices per account.


ara4n:
**Is your feature request related to a problem? Please describe.**

A common failure mode in Matrix is that a user ends up creating hundreds of devices on their account, causing problems with slow encryption, performance problems when they get deleted (https://github.com/matrix-org/synapse/issues/2704 and https://github.com/matrix-org/synapse/issues/14651 aka https://github.com/element-hq/synapse/issues/14651 and https://github.com/element-hq/synapse/issues/2704) etc.

**Describe the solution you'd like**

Ability to specify a per-user limit on the number of logins they're allowed per account.  Optionally one could prompt to log out stale devices if they hit the limit as they attempt to use new ones. "Sorry, you're only allowed 32 logins per account; please select some old ones to delete".

**Describe alternatives you've considered**

Leaving it as is, which basically amounts to a DoS vector.

**Additional context**

This actually came in from an Element customer, wanting to hard-limit 1 device per user, although it's something that's come up a lot in the past - e.g. https://github.com/matrix-org/synapse/issues/8263 / https://github.com/element-hq/synapse/issues/8263

MadLittleMods:
Closing as all of pieces have been merged and will be part of the next MAS release (estimated `v1.18.0`) 🚀 

Here is a break-down of what's happened in this space:

Matrix Live demo: https://www.youtube.com/watch?v=0Yr67OT61Vc&t=1042s

### Configuration

High level overview:

We have a `soft_limit` and a `hard_limit`. If you reach the `soft_limit` and you're in an interactive context, we’ll ask you to sign out some devices before continuing.

When that’s not possible (non-interactive contexts like a bot logging in), we will let you log in, but only up to the `hard_limit`, at which point we will either reject the login or automatically evict devices if `dangerous_hard_limit_eviction` is configured.

`dangerous_hard_limit_eviction` gets its name because removing sessions is a potentially damaging operation. Any end-to-end encrypted history on the device will be lost and can only be recovered if the user has another verified active device or has a recovery key setup.

[...]

PR #5215, opened 2025-10-31
Only allow C-S device scopes when the C-S API scope has been requested

reivilibre:
It'd be weird for a client to request a device on the client-server API but yet not request any client-server API scopes to use it with.

By adding this restriction, we can then create a partial index on the `oauth2_sessions` table to quickly identify sessions that have C-S API scopes and use this as a proxy metric for how many sessions *may* have device scopes.
This in turn makes it feasible to efficiently limit the number of 'devices' a user has, or more precisely: the number of sessions with client-server API access.

We can't do the same for device scopes themselves because, other than nastiness like parsing the JSON stringification of the scope list, it's not feasible to identify device scopes within a Postgres index predicate.

Part of: #4339

### DIFF
--- a/docs/development/contributing.md
+++ b/docs/development/contributing.md
@@ -96,7 +96,7 @@ If you haven't already, install [Cargo-Nextest](https://nexte.st/docs/installati
 - Run the tests to the backend by running `cargo nextest run --workspace`. This requires a connection to a PostgreSQL database, set via the `DATABASE_URL` environment variable.
 - Run the tests to the frontend by running `npm run test` in the `frontend` directory.
 - To run the tests for the included policy, change to the `policies` directory and run one of:
-  - `make test` (needs OpenPolicyAgent installed)
+  - `make test` (needs Open Policy Agent installed)
   - `make PODMAN=1 test` (runs inside a container; needs Podman installed)
   - `make DOCKER=1 test` (runs inside a container; needs Docker installed)
 


docs/development/contributing.md
```diff
@@ -86,13 +87,18 @@ While you're developing and before submitting a patch, you'll want to test your
...
+- Run `make fmt` and `make lint` in the `policies` directory to format and lint the included policy.
 
 ### Run the tests
 
 If you haven't already, install [Cargo-Nextest](https://nexte.st/docs/installation/pre-built-binaries/).
 
 - Run the tests to the backend by running `cargo nextest run --workspace`. This requires a connection to a PostgreSQL database, set via the `DATABASE_URL` environment variable.
 - Run the tests to the frontend by running `npm run test` in the `frontend` directory.
+- To run the tests for the included policy, change to the `policies` directory and run one of:
+  - `make test` (needs OpenPolicyAgent installed)
```
sandhose: nit
```suggestion
  - `make test` (needs Open Policy Agent installed)
```

policies/Makefile
```diff
@@ -24,9 +24,9 @@ ifeq ($(DOCKER), 1)
 	REGAL := docker run -i -v $(shell pwd):/policies:ro -w /policies --rm $(REGAL_DOCKER_IMAGE)
 else ifeq ($(PODMAN), 1)
 	# When running rootless, the volume directory may need to be given global write permissions on the host
-	OPA := podman run -i -v $(shell pwd):/policies:ro:Z -w /policies --rm $(OPA_DOCKER_IMAGE)
+	OPA := podman run -i -v $(shell pwd):/policies:ro,Z -w /policies --rm $(OPA_DOCKER_IMAGE)
```
sandhose, line 27: did that change in podman or did this never work?

sandhose (approved):

policies/Makefile
reivilibre: I don't know, but it only has one opts field now

PR #5221, opened 2025-11-06
Add experimental and preliminary policy-driven session limiting when logging in OAuth 2 sessions.

reivilibre:
Part of: #4339

Intended for commit-by-commit review.

Introduces a soft and hard session limit. At login time, we count the user's existing sessions (OAuth 2, Compat and self-owned Personal) and refuse creation of a new one if the limit is reached. This is configurable by policy.

The soft limit will be used for interactive logins where the user could be shown a web UI to remove devices, whereas the hard limit will be used for `m.login.password` logins where the user can't be shown such a UI.

Limitations:

- Only applies at OAuth 2 login time.
- Does not yet show a nice UI to rectify the solution, just a shouty policy violation screen.
- Triggers a false positive when the login includes device replacement. (The limit shouldn't apply if you are replacing a device.)

These limitations will be handled in future PRs, but for now this is part of the experimental config section.

### DIFF
--- a/docs/config.schema.json
+++ b/docs/config.schema.json
@@ -1,0 +1,2883 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema#",
+  "title": "RootConfig",
+  "description": "Application configuration root",
+  "type": "object",
+  "properties": {
+    "clients": {
+      "description": "List of OAuth 2.0/OIDC clients config",
+      "type": "array",
+      "items": {
+        "$ref": "#/definitions/ClientConfig"
+      }
+    },
+    "http": {
+      "description": "Configuration of the HTTP server",
+      "default": {
+        "listeners": [
+          {
+            "name": "web",
+            "resources": [
+              {
+                "name": "discovery"
+              },
+              {
+                "name": "human"
+              },
+              {
+                "name": "oauth"
+              },
+              {
+                "name": "compat"
+              },
+              {
+                "name": "graphql"
+              },
+              {
+                "name": "assets"
+              }
+            ],
+            "binds": [
+              {
+                "address": "[::]:8080"
+              }
+            ],
+            "proxy_protocol": false
+          },
+          {
+            "name": "internal",
+            "resources": [
+              {
+                "name": "health"
+              }
+            ],
+            "binds": [
+              {
+                "host": "localhost",
+                "port": 8081
+              }
+            ],
+            "proxy_protocol": false
+          }
+        ],
+        "trusted_proxies": [
+          "192.168.0.0/16",
+          "172.16.0.0/12",
+          "10.0.0.0/10",
+          "127.0.0.1/8",
+          "fd00::/8",
+          "::1/128"
+        ],
+        "public_base": "http://[::]:8080/",
+        "issuer": "http://[::]:8080/"
+      },
+      "allOf": [
+        {
+          "$ref": "#/definitions/HttpConfig"
+        }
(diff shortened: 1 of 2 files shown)


crates/config/src/sections/experimental.rs
```diff
@@ -106,3 +114,10 @@ impl ExperimentalConfig {
 impl ConfigurationSection for ExperimentalConfig {
     const PATH: Option<&'static str> = Some("experimental");
 }
+
+/// Configuration options for the inactive session expiration feature
+#[derive(Clone, Debug, Deserialize, JsonSchema, Serialize)]
+pub struct SessionLimitConfig {
+    pub soft_limit: u64,
+    pub hard_limit: u64,
```
sandhose: TBD, how does it behave when a limit is set to zero? Would it be worth using `NonZeroU64`?

crates/policy/src/lib.rs
```diff
@@ -85,18 +86,25 @@ impl Entrypoints {
     }
 }
 
-#[derive(Debug)]
+#[derive(Serialize, Debug)]
```
sandhose: alternatively this could have a serializeable static part plus rest as two separate properties, which feels a smidge less 'smart', but also fine with this

crates/handlers/src/oauth2/authorization/consent.rs
```diff
@@ -136,10 +136,15 @@ pub(crate) async fn get(
 
     let (csrf_token, cookie_jar) = cookie_jar.csrf_token(&clock, &mut rng);
 
+    let session_counts = count_user_sessions_for_limiting(&mut repo, &session.user)
+        .await
+        .map_err(|e| RouteError::Internal(e.into()))?;
```
sandhose: surprised RouteError doesn't implement From that error?

sandhose: oh, `anyhow`, that's why. Fine then

crates/policy/src/lib.rs
reivilibre: ok you tricked me into writing `DataBase`

crates/handlers/src/oauth2/authorization/consent.rs
reivilibre: to be fair, maybe I should have double-thought it and passed through the database error

crates/config/src/sections/experimental.rs
reivilibre: 604ed545b03e5af49ec230f4a357699ce5af003b

crates/policy/src/lib.rs
reivilibre: fde8dc6218e2cd3d17d83e797a22deaa8bf4a847

crates/handlers/src/oauth2/authorization/consent.rs
reivilibre: e7b538c63326043bea0dde116ba30e30838ad488

crates/config/src/sections/experimental.rs
```diff
@@ -106,3 +116,10 @@ impl ExperimentalConfig {
 impl ConfigurationSection for ExperimentalConfig {
     const PATH: Option<&'static str> = Some("experimental");
 }
+
+/// Configuration options for the inactive session expiration feature
```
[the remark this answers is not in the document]
reivilibre: busted

PR #5240, opened 2025-11-14
Modify Nginx compatibility-layer location block to forward not just client stuff to matrix-synapse

marrobHD:
The Nginx configuration guide for the compatibility layer does not seem complete.

It just forwards `/_matrix` & `/_synapse/client` to synapse. With the current state for example `_synapse/mas/upsert_device` would not reach synapse.

This MR modifies the location to fix this.

### DIFF
--- a/docs/setup/reverse-proxy.md
+++ b/docs/setup/reverse-proxy.md
@@ -94,7 +94,7 @@ server {
 
     # Forward to Synapse
     # as per https://element-hq.github.io/synapse/latest/reverse_proxy.html#nginx
-    location ~ ^(/_matrix|/_synapse) {
+    location ~ ^(/_matrix|/_synapse/client|/_synapse/mas) {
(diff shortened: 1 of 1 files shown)


docs/setup/reverse-proxy.md
```diff
@@ -94,7 +94,7 @@ server {
 
     # Forward to Synapse
     # as per https://element-hq.github.io/synapse/latest/reverse_proxy.html#nginx
-    location ~ ^(/_matrix|/_synapse/client) {
+    location ~ ^(/_matrix|/_synapse) {
```
reivilibre: I think we still are reluctant to tell people to forward `/_synapse/admin` so I think I will change this to as follows:
```suggestion
    location ~ ^(/_matrix|/_synapse/client|/_synapse/mas) {
```

PR #5281, opened 2025-11-21
Unify registrations for local passwords and upstream OAuth registrations

### LINKED ISSUE #4980: Require registration tokens when registering via OAuth provider
I'm running a private server and I want to limit the ability to register accounts on my server to my friends and relatives.
I have enabled `registration_token_required: true` and this effectively limited registrations with email and password.
But I've setup Google OAuth to allow users to sign-up/sign-in with Google account to simplify registration process.
The problem is that the registration token is not required when signing up with Google, allowing strangers to register an account.

Enforce registration tokens for all kinds of registrations: both email/password and OAuth.

sandhose:
This can be reviewed commit by commit.

The intention of this is to unify how we register user, by using the `UserRegistration` abstraction for upstream OAuth registrations. I intended to do that when I introduced that in #3784 but didn't got around to do it.

The nice thing with this is that it gives us registration tokens support for upstream OAuth providers

Fixes #4980

### DIFF
--- a/crates/storage-pg/src/user/registration.rs
+++ b/crates/storage-pg/src/user/registration.rs
@@ -902,7 +902,7 @@ mod tests {
     }
 
     #[sqlx::test(migrator = "crate::MIGRATOR")]
-    async fn test_set_upstream_oauth_link(pool: PgPool) {
+    async fn test_set_upstream_oauth_session(pool: PgPool) {
         let mut rng = ChaChaRng::seed_from_u64(42);
         let clock = MockClock::default();
 


crates/storage/src/user/email.rs
```diff
@@ -306,12 +306,34 @@ pub trait UserEmailRepository: Send + Sync {
...
     /// Returns an error if the underlying repository fails
-    async fn complete_authentication(
+    async fn complete_authentication_with_code(
         &mut self,
         clock: &dyn Clock,
         authentication: UserEmailAuthentication,
         code: &UserEmailAuthenticationCode,
     ) -> Result<UserEmailAuthentication, Self::Error>;
+
+    /// Complete a [`UserEmailAuthentication`] by using the given upstream oauth
```
reivilibre, line 316: is this saying that we accept e.g. Apple's claim of a user's email address instead of them actually verifying their email directly with us?

Maybe this doc would be a bit better if it said the circumstances in which you can do this, but maybe something I'll see in later commits.

crates/storage-pg/migrations/20251121145458_user_registration_upstream_oauth_session.sql
```diff
@@ -0,0 +1,13 @@
...
+-- Please see LICENSE in the repository root for full details.
+
+-- Track what upstream OAuth session to associate during user registration
+ALTER TABLE user_registrations
+    ADD COLUMN upstream_oauth_authorization_session_id UUID
+      REFERENCES upstream_oauth_authorization_sessions (upstream_oauth_authorization_session_id)
+      ON DELETE SET NULL;
+
+CREATE INDEX user_registrations_upstream_oauth_session_id_idx
+    ON user_registrations (upstream_oauth_authorization_session_id);
```
reivilibre: is `user_registrations` small enough that we're happy to do this non-CONCURRENTLY?

crates/storage-pg/src/user/registration.rs
```diff
@@ -851,4 +900,120 @@ mod tests {
             .await;
         assert!(res.is_err());
     }
+
+    #[sqlx::test(migrator = "crate::MIGRATOR")]
+    async fn test_set_upstream_oauth_link(pool: PgPool) {
```
reivilibre: maybe `_link` is a slight misnomer / not consistent with the other methods

crates/storage/src/user/email.rs
sandhose: We already assumed that. In the past, we've let the admin choose in the configuration whether to trust it or not, but this lead to more confusion than anything

crates/storage-pg/migrations/20251121145458_user_registration_upstream_oauth_session.sql
sandhose: It might, but that's a good point! Fixed in 4c3d2bae8

crates/storage-pg/src/user/registration.rs
sandhose: That's a leftover for when it was the link that was attached to registrations, not the session! Fixed in 61ee8dae8

PR #5286, opened 2025-11-25
Explain the purpose of signing keys in the config documentation.

reivilibre:
We had some questions about what these were used for; I gave answers along these lines. I have now reworded a little for the docs.
@sandhose Please check you agree with them :D

A relevant issue that I haven't addressed is https://github.com/element-hq/matrix-authentication-service/issues/4806, which asks about the significance of the order of `keys`.
I didn't want to commit to any specific behaviour except the ranking based on claim matching, because the fact we do a stable sort and pick the last key in the list seems like an implementation detail and anything relying on it seems brittle.

### DIFF
--- a/docs/reference/configuration.md
+++ b/docs/reference/configuration.md
@@ -238,9 +238,12 @@ The following key formats are supported:
 - PKCS#8 PEM or DER-encoded RSA or ECDSA private key, encrypted or not
 - SEC1 PEM or DER-encoded ECDSA private key
 
-The signing keys are used for signing ID Tokens (as returned in the [Token Endpoint]
-at `/oauth2/token`) and for signing the response of the [UserInfo Endpoint] at
-`/oauth2/userinfo` if the client requests a signed response.
+The signing keys are used for:
+- signing ID Tokens (as returned in the [Token Endpoint] at `/oauth2/token`);
+- signing the response of the [UserInfo Endpoint] at `/oauth2/userinfo` if the
+  client requests a signed response;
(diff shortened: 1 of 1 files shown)


docs/reference/configuration.md
```diff
@@ -222,7 +222,7 @@ The secret is not updated when the content of the file changes.
 > Changing the encryption secret afterwards will lead to a loss of all encrypted
 > information in the database.
 
-### Singing Keys
+### Signing Keys
```
sandhose, line 225: 🎵 🐦

docs/reference/configuration.md
```diff
@@ -238,9 +238,21 @@ The following key formats are supported:
 - PKCS#8 PEM or DER-encoded RSA or ECDSA private key, encrypted or not
 - SEC1 PEM or DER-encoded ECDSA private key
 
+The signing keys are used for signing ID Tokens (as returned in the [Token Endpoint]
```
sandhose: They are also used if an upstream is configured to use the `private_key_jwt` client auth method; in which case, MAS uses one of those private keys to sign a JWT that the upstream OAuth provider can verify for auth. But it's a bit of a niche use case, not necessarily worth mentioning

sandhose (approved):

docs/reference/configuration.md
reivilibre: I feel like I'd rather mention this, even if we somehow point out it's a niche use case

sandhose (approved):

PR #5287, opened 2025-11-25
Add experimental and preliminary policy-driven session limiting when logging in compatibility sessions.

reivilibre:
Part of: #4339
Follows: #5221

This PR introduces a new policy for compatibility logins and starts off by adding a session limit.

Compatibility login is a bit troublesome in that the actual device creation is uninteractive, so we can't show an interactive screen to clean up devices at the time of actually creating the device!
Only for `m.login.sso`-style logins can we do an optimistic 'please clean up some devices' screen (although this PR on its own just displays a generic policy violation screen),
but this can only happen *before* the actual creation of the session.

For this reason we will make use of a soft and hard limit. The soft limit will display a policy violation screen in interactive contexts but not be enforced in uninteractive contexts.
Meanwhile the hard limit *will* be enforced in the uninteractive contexts, eventually (in the future) through automatic eviction of a session.

### DIFF
--- a/templates/pages/compat_login_policy_violation.html
+++ b/templates/pages/compat_login_policy_violation.html
@@ -1,6 +1,5 @@
 {#
-Copyright 2024, 2025 New Vector Ltd.
-Copyright 2022-2024 The Matrix.org Foundation C.I.C.
+Copyright 2025 New Vector Ltd.
 
 SPDX-License-Identifier: AGPL-3.0-only OR LicenseRef-Element-Commercial
 Please see LICENSE files in the repository root for full details.
--- a/translations/en.json
+++ b/translations/en.json
@@ -22,7 +22,7 @@
     },
     "sign_out": "Sign out",
     "@sign_out": {
-      "context": "pages/account/logged_out.html:22:28-48, pages/compat_login_policy_violation.html:29:28-48, pages/consent.html:65:28-48, pages/device_consent.html:136:30-50, pages/index.html:28:28-48, pages/policy_violation.html:38:28-48, pages/sso.html:45:28-48, pages/upstream_oauth2/link_mismatch.html:24:24-44, pages/upstream_oauth2/suggest_link.html:32:26-46"
+      "context": "pages/account/logged_out.html:22:28-48, pages/compat_login_policy_violation.html:28:28-48, pages/consent.html:65:28-48, pages/device_consent.html:136:30-50, pages/index.html:28:28-48, pages/policy_violation.html:38:28-48, pages/sso.html:45:28-48, pages/upstream_oauth2/link_mismatch.html:24:24-44, pages/upstream_oauth2/suggest_link.html:32:26-46"
     },
     "skip": "Skip",
     "@skip": {
@@ -496,17 +496,17 @@
     "policy_violation": {
       "description": "This might be because of the client which authored the request, the currently logged in user, or the request itself.",
       "@description": {
-        "context": "pages/compat_login_policy_violation.html:19:25-62, pages/policy_violation.html:19:25-62",
+        "context": "pages/compat_login_policy_violation.html:18:25-62, pages/policy_violation.html:19:25-62",
         "description": "Displayed when an authorization request is denied by the policy"
       },
       "heading": "The authorization request was denied by the policy enforced by this service",
       "@heading": {
-        "context": "pages/compat_login_policy_violation.html:18:27-60, pages/policy_violation.html:18:27-60",
+        "context": "pages/compat_login_policy_violation.html:17:27-60, pages/policy_violation.html:18:27-60",
         "description": "Displayed when an authorization request is denied by the policy"
       },
       "logged_as": "Logged as <span class=\"font-semibold\">%(username)s</span>",
       "@logged_as": {
-        "context": "pages/compat_login_policy_violation.html:26:11-86, pages/policy_violation.html:35:11-86"
+        "context": "pages/compat_login_policy_violation.html:25:11-86, pages/policy_violation.html:35:11-86"
       }
     },
     "recovery": {


crates/policy/src/model.rs
```diff
@@ -187,6 +187,42 @@ pub struct AuthorizationGrantInput<'a> {
...
+    /// Whether the user is currently in an interactive context.
+    /// For `m.login.password`: false
+    /// For `m.login.sso`:
+    /// - true when asking for consent,
+    /// - false when actually performing the login (at which point we create the
+    ///   compat session, but it's too late to show a web page)
+    pub is_interactive: bool,
+
+    // TODO I don't know if we should keep this anymore, not used by current policy
+    pub login_type: CompatLoginType,
```
reivilibre: thoughts?

templates/pages/compat_login_policy_violation.html
```diff
@@ -0,0 +1,34 @@
...
+  <main class="flex flex-col gap-10">
+    <div class="flex gap-1 justify-center items-center">
+      <p class="cpd-text-secondary cpd-text-body-md-regular">
+        {{ _("mas.policy_violation.logged_as", username=current_session.user.username) }}
+      </p>
+
+      {{ logout.button(text=_("action.sign_out"), csrf_token=csrf_token, post_logout_action=action, as_link=True) }}
+    </div>
+
+    <a class="cpd-button destructive" data-kind="primary" data-size="lg" href="/TODO">{{ _("action.cancel") }}</a>
```
reivilibre: ...is there even a cancel action we *can* display?

crates/handlers/src/compat/login_sso_complete.rs
```diff
@@ -107,6 +116,38 @@ pub async fn get(
...
 
+    let session_counts = count_user_sessions_for_limiting(&mut repo, &session.user).await?;
+
+    let res = policy
+        .evaluate_compat_login(mas_policy::CompatLoginInput {
+            user: &session.user,
+            is_interactive: true,
+            login_type: CompatLoginType::WebSso,
+            // TODO should we predict a replacement?
+            session_replaced: false,
```
reivilibre: heh, so there's at the very least if we try and 'predict' a replacement, there would be a TOCTOU hazard here.

But now I think about it, I don't think we even know if there's a device ID by this point, do we?

crates/handlers/src/compat/login_sso_complete.rs
```diff
@@ -200,6 +246,40 @@ pub async fn post(
...
+    let session_counts = count_user_sessions_for_limiting(&mut repo, &session.user).await?;
+
+    let res = policy
+        .evaluate_compat_login(mas_policy::CompatLoginInput {
+            user: &session.user,
+            is_interactive: true,
+            login_type: CompatLoginType::WebSso,
+            session_counts,
+            // TODO should we predict a replacement?
+            session_replaced: false,
```
reivilibre: (same as the other)

crates/handlers/src/compat/login_sso_complete.rs
reivilibre: (I don't think we know the device ID here anyway, so we can't predict anything anyway)

templates/pages/compat_login_policy_violation.html
reivilibre: It appears Synapse's SSO pages don't show any action on an error. I also don't see any way to handle errors in the spec, I guess it's assumed the user will close the browser on their own.

crates/policy/src/model.rs
```diff
@@ -187,6 +187,32 @@ pub struct AuthorizationGrantInput<'a> {
...
+pub struct CompatLoginInput<'a> {
+    #[schemars(with = "std::collections::HashMap<String, serde_json::Value>")]
+    pub user: &'a User,
+
+    /// How many sessions the user has.
+    pub session_counts: SessionCounts,
+
+    // TODO is this actually what we care about? Don't we care a bit more about whether we're in an
+    // interactive context or a non-interactive context? (SSO type has both phases :()
+    pub login_type: CompatLoginType,
```
sandhose: It would be nice if the context also included the SSO redirect URI on `m.login.sso`? Surely we'll want to limit those in some contexts

policies/compat_login/compat_login.rego
```diff
@@ -0,0 +1,61 @@
...
+violation contains {
+	"code": "too-many-sessions",
+	"msg": "user has too many active sessions (soft limit)",
+} if {
+	# Only apply if session limits are enabled in the config
+	data.session_limit != null
+
+	# This is a web-based interactive login
+	# TODO not strictly correct...
+	input.login_type == "m.login.sso"
```
sandhose: you could easily do a helper like

```rego
is_interactive if {
    input.login_type == "m.login.sso"
}
```

crates/handlers/src/compat/login.rs
```diff
@@ -548,6 +580,27 @@ async fn token_login(
...
+    let res = policy
+        .evaluate_compat_login(mas_policy::CompatLoginInput {
+            user: &browser_session.user,
+            login_type: CompatLoginType::WebSso,
+            session_counts,
+            requester,
+        })
+        .await?;
+    if !res.valid() {
+        if res.violations.len() == 1 {
```
sandhose, line 601: what's the reasoning behind having a different message only if there is a single "too-many-sessions" violation vs. if it is one of them in the list?

In any case, I think this would benefit from a comment of what we're doing and why

crates/handlers/src/compat/login_sso_complete.rs
```diff
@@ -107,6 +117,35 @@ pub async fn get(
...
+            login_type: CompatLoginType::WebSso,
+            session_counts,
+            requester: mas_policy::Requester {
+                ip_address: activity_tracker.ip(),
+                user_agent,
+            },
+        })
+        .await?;
+    if !res.valid() {
+        let ctx = CompatLoginPolicyViolationContext::for_violations(
```
sandhose: I wonder if we should just pass the violation result? I assume this is a dependency problem, mas-templates would need to depend on mas-policy? Not a strong opinion if it's too annoying

crates/policy/src/model.rs
```diff
@@ -200,8 +200,15 @@ pub struct CompatLoginInput<'a> {
...
 
-    // TODO is this actually what we care about? Don't we care a bit more about whether we're in an
-    // interactive context or a non-interactive context? (SSO type has both phases :()
+    /// Whether the user is currently in an interactive context.
+    /// For `m.login.password`: false
+    /// For `m.login.sso`:
+    /// - true when asking for consent,
+    /// - false when actually performing the login (at which point we create the
+    ///   compat session, but it's too late to show a web page)
+    pub is_interactive: bool,
```
sandhose: what if the distinction was instead `m.login.token` on the login type since that's the actual login type that is going through at this stage?

reivilibre: Did consider, no super strong opinion on that, but if you veer towards it let's do that.

crates/handlers/src/compat/login.rs
reivilibre: Definitely should comment, but I think the underlying thought was: If you have too many sessions but also your client 'Tnemele' is banned, isn't showing you the UI for 'pick a device to delete' missing the point?

crates/policy/src/model.rs
reivilibre: changed things around and expose that in 9c7c15774450be6becab5a1b4506277b70ad202d now

policies/compat_login/compat_login.rego
reivilibre: makes sense, done as much in 9c7c15774450be6becab5a1b4506277b70ad202d

crates/policy/src/model.rs
reivilibre: in 9c7c15774450be6becab5a1b4506277b70ad202d

crates/handlers/src/compat/login_sso_complete.rs
reivilibre: exposed direct violations to template in d21922f10f8f5ab62af736b0b3353cd5502b528a, didn't turn out as bad as I expected

crates/handlers/src/compat/login.rs
reivilibre: added this comment to 8f523e3959014071d1d0636ff587d6b8d85763c8

templates/pages/compat_login_policy_violation.html
```diff
@@ -0,0 +1,32 @@
+{#
+Copyright 2024, 2025 New Vector Ltd.
+Copyright 2022-2024 The Matrix.org Foundation C.I.C.
```
sandhose:
```suggestion
Copyright 2025 New Vector Ltd.
```

I think that's what we're supposed to do for new files

PR #5293, opened 2025-11-27
Allow linking upstream accounts to matching users without confirmation

### LINKED ISSUE #5284: 'suggest' action on localpart isn't really a suggestion when the suggested localpart is already in use
**Bug Description**
With MAS in between a synapse and a keycloak instance, when using

```
claims_imports:
        localpart:
          action: suggest
          template: "{{ user.email | split('@') | first }}"
```

MAS doesn't let the user choose a different localpart when the localpart suggested by the upstream oauth provider is already in use. Instead, it just fails registration.

**Steps To Reproduce**
1. have an upstream oauth2 provider
2. configure localpart claim import as seen above
3. register a user with email someone@domain1 from upstream (don't change the suggested localpart of 'someone' in the form)
4. try to register a user with email someone@domain2 from upstream
5. get "Upstream account provider returned "someone" as username, which is not linked to that upstream account. Your homeserver does not allow linking an upstream account to an existing account."

**Expected behavior**
I would expect the same registration flow as in step 3 to occur in step 4, only with a 'this username is already taken' or something like that for the username text field.

**Screenshots**
not really helpful, but can provide screenshots when neccessary.

 - OS: nixos
 - Browser firefox
 - Version 145.0.1

**Additional context**
MAS and synapse are running from the latest docker image tagged 'latest'

sandhose:
This reworks the link flow to handle many edge cases better.

One major functionality change is that when we had a new upstream account with no user linked, but the localpart matching an existing user, if `on_conflict` was set to `add`, we prompt the user to link the existing account.
This prompt is now skipped and the user is linked automatically.

Another important thing it does, is that it now handles cases when the username is only 'suggested' better, as it will ignore it if the user exists, or if the suggested username doesn't pass the policy check.

Fixes #5284

### DIFF
--- a/crates/handlers/src/upstream_oauth2/link.rs
+++ b/crates/handlers/src/upstream_oauth2/link.rs
@@ -454,7 +454,7 @@ pub(crate) async fn get(
             }
 
             // We do a bunch of checks for the localpart. Instead of using nested ifs all
-            // the way, we use a labelled block, and use `break` for 'exitting' early when
+            // the way, we use a labelled block, and use `break` for 'exiting' early when
             // needed
             let localpart = 'localpart: {
                 if provider.claims_imports.localpart.ignore() {
@@ -854,7 +854,7 @@ pub(crate) async fn post(
             if let Some(ref display_name) = display_name {
                 ctx = ctx.with_display_name(
                     display_name.clone(),
-                    provider.claims_imports.email.is_forced_or_required(),
+                    provider.claims_imports.displayname.is_forced_or_required(),
                 );
             }
 


crates/handlers/src/upstream_oauth2/link.rs
```diff
@@ -834,14 +851,12 @@ pub(crate) async fn post(
                 None
             };
 
-            let ctx = if let Some(ref display_name) = display_name {
-                ctx.with_display_name(
+            if let Some(ref display_name) = display_name {
+                ctx = ctx.with_display_name(
                     display_name.clone(),
                     provider.claims_imports.email.is_forced_or_required(),
```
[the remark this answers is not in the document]
sandhose: That's a fun bug.

crates/handlers/src/upstream_oauth2/link.rs
```diff
@@ -437,140 +438,254 @@ pub(crate) async fn get(
...
+                    }
+
+                    if existing_user.locked_at.is_some() {
+                        // The account is locked, show the 'account locked' fallback
+                        let ctx = AccountInactiveContext::new(existing_user)
+                            .with_csrf(csrf_token.form_value())
+                            .with_language(locale);
+                        let fallback = templates.render_account_locked(&ctx)?;
+                        return Ok((cookie_jar, Html(fallback).into_response()));
+                    }
```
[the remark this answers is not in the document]
sandhose: We roll back the transaction anyway, that's fine

crates/handlers/src/upstream_oauth2/link.rs
```diff
@@ -437,140 +438,254 @@ pub(crate) async fn get(
...
+                ctx = ctx.with_email(
+                    email.clone(),
+                    provider.claims_imports.email.is_forced_or_required(),
+                );
+            }
+
+            // We do a bunch of checks for the localpart. Instead of using nested ifs all
+            // the way, we use a labelled block, and use `break` for 'exiting' early when
+            // needed
+            let localpart = 'localpart: {
```
reivilibre, line 459: this is a handy trick I should probably make more use of

crates/handlers/src/upstream_oauth2/link.rs
```diff
@@ -437,140 +438,254 @@ pub(crate) async fn get(
...
+                    // TODO: translate
+                    let ctx = ErrorContext::new()
+                        .with_code("Policy error")
+                        .with_description(format!(
+                            r"Upstream account provider returned {localpart:?} as username,
+                            which does not pass the policy check: {res}"
+                        ))
+                        .with_language(&locale);
+
+                    return Ok((
```
reivilibre, line 524: should there be a status code here (& other places too)?

sandhose: Since it's mostly a user visible page, I don't think it really matters? I guess it would show up in logs/metrics 🤔

PR #5295, opened 2025-11-28
Add more options to deal with localpart conflicts on upstream OAuth 2.0 logins

### LINKED ISSUE #2089: Make it possible to auto-import existing users when the localpart matches in upstream OAuth 2.0 logins

sandhose:
Fixes #2089
Follow up from #4193
On top of #5293 

This adds more granular options for how to deal when an upstream login matches an existing user.

 - `set` will only add the link *if and only if* there is no other existing link on the user for that provider. This is likely the safest option in most cases
 - `replace` will replace any existing link on the matching user for that provider

### DIFF
--- a/crates/handlers/src/upstream_oauth2/link.rs
+++ b/crates/handlers/src/upstream_oauth2/link.rs
@@ -588,7 +588,7 @@ pub(crate) async fn get(
                                 .await?;
                         }
 
-                        // We matched anexisting user and the conflict resolution is to replace any
+                        // We matched an existing user and the conflict resolution is to replace any
                         // link on the existing user with this one
                         UpstreamOAuthProviderOnConflict::Replace => {
                             // Find existing links for this provider and user
@@ -635,7 +635,7 @@ pub(crate) async fn get(
                                 .await?;
                         }
 
-                        // We matched an existing user and the conflict resolution is link to the
+                        // We matched an existing user and the conflict resolution is to link to the
                         // existing user *only if* there is no existing link on that user
                         UpstreamOAuthProviderOnConflict::Set => {
                             // Find existing links for this provider and user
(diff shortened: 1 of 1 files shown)


crates/cli/src/sync.rs
```diff
@@ -45,6 +45,12 @@ fn map_import_on_conflict(
         mas_config::UpstreamOAuth2OnConflict::Add => {
             mas_data_model::UpstreamOAuthProviderOnConflict::Add
         }
+        mas_config::UpstreamOAuth2OnConflict::Replace => {
+            mas_data_model::UpstreamOAuthProviderOnConflict::Replace
+        }
+        mas_config::UpstreamOAuth2OnConflict::Set => {
```
reivilibre, line 51: Not the hugest fan of calling this `Set` because it doesn't give off any vibe as to what it means. Maybe `SetNew`, kinda like `OpenOptions::create_new`?
Or `SetIfNotExists` kinda like SQL `CREATE IF NOT EXISTS` (even though now that I think about it, that's not great grammar :D)?

sandhose: Errr I don't know; at the same time, `set` is probably the option which makes most sense in most cases, so I don't want it to look long and complicated

sandhose: I think that as long as it's documented, it's fine… I can't really think of another single word to better express it?

PR #5297, opened 2025-11-28
Add 'IF NOT EXISTS' to all 'CREATE INDEX CONCURRENTLY' statements to avoid deadlocks

benbz:
ESS CI is sometimes seeing issues where MAS (with multiple replicas) refuses to initially startup

The Postgres logs show
```
2025-11-28 14:42:46.989 UTC [127] ERROR:  deadlock detected
2025-11-28 14:42:46.989 UTC [127] DETAIL:  Process 127 waits for ShareLock on virtual transaction 19/5; blocked by process 128.
	Process 128 waits for ExclusiveLock on advisory lock [16385,438033630,3765090638,1]; blocked by process 127.
	Process 127: -- no-transaction
	-- Copyright 2025 New Vector Ltd.
	--
	-- SPDX-License-Identifier: AGPL-3.0-only
	-- Please see LICENSE in the repository root for full details.
	
	CREATE INDEX CONCURRENTLY
	  compat_access_tokens_session_fk
	  ON compat_access_tokens (compat_session_id);
	
	Process 128: SELECT pg_advisory_lock($1)
```

DB migration jobs then fail with `relation "compat_access_tokens_session_fk" already exists`.

On discussion this is the first migration that isn't in a transaction / uses `CREATE INDEX CONCURRENTLY`. It is assumed that removing `CONCURRENTLY` isn't desirable as will cause slowness on large MAS DBs upgrading, so use `IF NOT EXISTS` to subsequent migration attempts to progress. Unclear whether this will stop the deadlock in the first place but this shouldn't make the problem any worse

### DIFF
--- a/crates/storage-pg/migrations/20250410000000_idx_compat_access_tokens_session_fk.sql
+++ b/crates/storage-pg/migrations/20250410000000_idx_compat_access_tokens_session_fk.sql
@@ -4,6 +4,6 @@
 -- SPDX-License-Identifier: AGPL-3.0-only
 -- Please see LICENSE in the repository root for full details.
 
-CREATE INDEX CONCURRENTLY
+CREATE INDEX CONCURRENTLY IF NOT EXISTS
   compat_access_tokens_session_fk
   ON compat_access_tokens (compat_session_id);
--- a/crates/storage-pg/migrations/20250410000001_idx_compat_refresh_tokens_session_fk.sql
+++ b/crates/storage-pg/migrations/20250410000001_idx_compat_refresh_tokens_session_fk.sql
@@ -4,6 +4,6 @@
 -- SPDX-License-Identifier: AGPL-3.0-only
 -- Please see LICENSE in the repository root for full details.
 
-CREATE INDEX CONCURRENTLY
+CREATE INDEX CONCURRENTLY IF NOT EXISTS
   compat_refresh_tokens_session_fk
   ON compat_refresh_tokens (compat_session_id);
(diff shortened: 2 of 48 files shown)


reivilibre:
Some extra context that should be captured here rather than buried in our chat:

This `CREATE INDEX CONCURRENTLY` statement causes a deadlock when there are multiple MAS processes trying to run the database migrations because:
 - the migration running this statement, which holds the advisory lock for applying migrations, must wait for all existing transactions that could potentially modify or use the index to terminate
- the other MAS process is blocking to acquire an advisory lock, which means it isn't shutting down and won't unblock the `CREATE INDEX CONCURRENTLY`

This appears to be the same cause as an issue in another migration-related project: https://github.com/golang-migrate/migrate/issues/960#issuecomment-1948570316

The `IF NOT EXISTS` actually doesn't resolve this problem above, but it is useful for another reason: when we run these migrations, we opt-out of transactional semantics with `-- no-transaction` (from memory, I don't think `CREATE INDEX CONCURRENTLY` can run in a transaction?).
So that means that the migration doesn't happen atomically with the operation of persisting the fact that we completed the migration, meaning this migration might run a second time despite the index already being created.

(I'm not clear why the supposedly deadlocked operation gets this far, to be honest.)

reivilibre:
Opened an issue at https://github.com/element-hq/matrix-authentication-service/issues/5298 to track the idea that we should poll for advisory lock, rather than block, to truly avoid the deadlock (rather than just workaround/allow it to proceed stochastically).

PR #5300, opened 2025-12-01
Reapply "Add 'IF NOT EXISTS' to all 'CREATE INDEX CONCURRENTLY' statements to avoid deadlocks (#5297)"

### LINKED ISSUE #5301: `CREATE INDEX CONCURRENTLY` migrations should use `IF NOT EXISTS`
This is because these migrations don't run in a transaction and so could be double-applied.

PR open but blocked: https://github.com/element-hq/matrix-authentication-service/pull/5300

Blocked by the ability to support changing migrations after they've been applied. Currently sqlx will generate migration errors when detecting that scenario.

reivilibre:
Reverts #5299 to reapply #5297

Fixes #5301 

This reverts commit 6abe9ac5e26dbc99f62aeef3087821eea067bc2c, reapplying changes made to 1de9148f533d368ec440959ca8807e76ab06485c.


sandhose:
I've opened #5329 to allow migrations to be modified

sandhose (approved):

PR #5312, opened 2025-12-03
Support for stable MSC4191 account management actions

hughns:
Ref: MSC4191

- Support the stable names and the unstable names
- Advertise both the stable names and the unstable names

This means that MAS supports either names. We can remove the unstable names once the is enough client adoption of the stable names.

n.b. this does not change the oauth2-types crate as not used by MAS.

### DIFF
--- a/crates/handlers/src/oauth2/discovery.rs
+++ b/crates/handlers/src/oauth2/discovery.rs
@@ -187,7 +187,8 @@ pub(crate) async fn get(
             "org.matrix.device_view".to_owned(),
             "org.matrix.device_delete".to_owned(),
             "org.matrix.cross_signing_reset".to_owned(),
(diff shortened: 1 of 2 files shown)


crates/handlers/src/oauth2/discovery.rs
```diff
@@ -183,10 +183,15 @@ pub(crate) async fn get(
         // see frontend/src/routes/__root.tsx
         account_management_actions_supported: vec![
             "org.matrix.profile".to_owned(),
+            "org.matrix.devices_list".to_owned(),
+            "org.matrix.device_view".to_owned(),
+            "org.matrix.device_delete".to_owned(),
+            "org.matrix.cross_signing_reset".to_owned(),
```
reivilibre, line 189: notably, missing support for `org.matrix.account_deactivate`

I guess this is currently intentional as we don't support it yet?

Issue #5298, opened 2025-12-01
Database Migrations should poll for advisory lock, not block, to avoid deadlock on `CREATE INDEX CONCURRENTLY`


reivilibre:
Spawned from the investigation into https://github.com/element-hq/matrix-authentication-service/pull/5297, with an explanatory comment at https://github.com/element-hq/matrix-authentication-service/pull/5297#issuecomment-3596168969:

> Some extra context that should be captured here rather than buried in our chat:
> 
> This `CREATE INDEX CONCURRENTLY` statement causes a deadlock when there are multiple MAS processes trying to run the database migrations because:
>  - the migration running this statement, which holds the advisory lock for applying migrations, must wait for all existing transactions that could potentially modify or use the index to terminate
> - the other MAS process is blocking to acquire an advisory lock, which means it isn't shutting down and won't unblock the `CREATE INDEX CONCURRENTLY`
> 
> This appears to be the same cause as an issue in another migration-related project: https://github.com/golang-migrate/migrate/issues/960#issuecomment-1948570316
> 
> The `IF NOT EXISTS` actually doesn't resolve this problem above, but it is useful for another reason: when we run these migrations, we opt-out of transactional semantics with `-- no-transaction` (from memory, I don't think `CREATE INDEX CONCURRENTLY` can run in a transaction?).
> So that means that the migration doesn't happen atomically with the operation of persisting the fact that we completed the migration, meaning this migration might run a second time despite the index already being created.
> 
> (I'm not clear why the supposedly deadlocked operation gets this far, to be honest.)

The outcome should be the same as for https://github.com/golang-migrate/migrate/issues/960#issuecomment-1948570316: for getting the advisory lock to run migrations, we should switch to `pg_try_advisory_lock` and poll in a loop (without holding a transaction open), instead of using `pg_advisory_lock` which blocks.

(Also not clear if this means we will have problems running migrations when we have `LISTEN` connections open.)

PR #5329, opened 2025-12-04
Fix running multiple migration process in parallel

sandhose:
Fixes https://github.com/element-hq/matrix-authentication-service/issues/5298

This rewrites the database migration code to:

 - avoid deadlocks when running multiple migration processes at the same
   time with a `CREATE INDEX CONCURRENTLY` statement
 - allow us to remove some migrations from the code base and mark them as
   intentionally removed
 - allow us to modify some migrations and declare alternate checksums
   for previous versions of the migration

We also get a few bonuses from this:

 - we can now log whenever a migration gets applied
 - we now have a span for each migration being applied to help investigate timings
 - removed the spurious `relation "_sqlx_migrations" already exists, skipping` notice on startup

### DIFF
--- a/docs/development/database.md
+++ b/docs/development/database.md
@@ -54,7 +54,7 @@ Note that migrations are embedded in the final binary and can be run from the se
 ### Removing migrations
 
 For various reasons, we may want to delete migrations.
-In case we do, we *must* declare that migration version as it is fine to be missing.
+In case we do, we *must* declare that migration version as allowed to be missing.
 This is because on startup, MAS will validate that all the applied migrations are known, and warn if some are missing.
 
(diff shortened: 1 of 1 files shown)


docs/development/database.md
```diff
@@ -50,3 +50,29 @@ cargo sqlx migrate add [description] # Add new migration files
 ```
 
 Note that migrations are embedded in the final binary and can be run from the service CLI tool.
+
+### Removing migrations
+
+For various reasons, we may want to delete migrations.
+In case we do, we *must* declare that migration version as it is fine to be missing.
```
reivilibre:
```suggestion
In case we do, we *must* declare that migration version as allowed to be missing.
```

PR #5339, opened 2025-12-09
Remove `rustls-pemfile` dependency

### LINKED ISSUE #5337: CI is failing on cargo-deny for rustls-pemfile crate
The `rustls-pemfile` crate is newly unmaintained.
Rather than having this block the current release cycle, an ignore was added to allow this for now.

This should be addressed, and the ignore removed, before the next release cycle.

devonh:
Fixes #5337 

Removes the `rustls-pemfile` dependency and replaces it's usage with `rustls-pki-types`.

### DIFF
--- a/deny.toml
+++ b/deny.toml
@@ -19,9 +19,6 @@ ignore = [
     # RSA key extraction "Marvin Attack". This is only relevant when using
     # PKCS#1 v1.5 encryption, which we don't
     "RUSTSEC-2023-0071",
-    # This is a newly unmaintained package that we can allow temporarily.
-    # Remove ASAP once https://github.com/element-hq/matrix-authentication-service/issues/5337 is fixed.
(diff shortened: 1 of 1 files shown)


anoadragon453 (approved):
Looks like we're just replacing methods to read the same keys. And the [original advisory](https://github.com/element-hq/matrix-authentication-service/issues/5337#issuecomment-3636203737) recommended we use this alternative dependency (which itself was used by `rustls-pemfile`.

This all looks sound to me. I requested a review from @element-hq/security out of paranoia in case there's some unknown unknown while dealing with certificates. For context, this code sets up TLS certificates for MAS workers ([used here](https://github.com/element-hq/matrix-authentication-service/blob/617d063251f2845145a6322e7bfe5f57d4e5011c/crates/cli/src/commands/server.rs#L275)).

sandhose (approved):

PR #5371, opened 2025-12-18
Fix configuration names secret_path vs secret_file

schildbach:
It's the other way round: Synapse requires `secret_path`, MAS requires `secret_file`.

(I think it should be changed to `secret_file` for both, because it's a file not a directory or tree.)

### DIFF
--- a/docs/setup/homeserver.md
+++ b/docs/setup/homeserver.md
@@ -20,7 +20,7 @@ matrix:
   endpoint: "http://localhost:8008"
   secret: "AVeryRandomSecretPleaseUseSomethingSecure"
   # Alternatively, using a file:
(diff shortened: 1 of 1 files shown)


sandhose:
Hey, thanks for this contribution!

> I won't sign CLAs.

We cannot accept contributions without the CLA being signed. There are more details in [this blog post](https://element.io/blog/element-to-adopt-agplv3/) on why this is, and to quote one specific part that is I think relevant:

> to be clear, the sole reason for a CLA is to allow Element to dual-license the software as per https://gnu.org/philosophy/selling-exceptions.html - not to give Element the ability to relicense to a non-OSI license in future. After all, Element already had that ability with the Apache license, and has not used it.

PR #5376, opened 2025-12-26
Fix CAPTCHA hostname verification when providers return parent domains

### LINKED ISSUE #3549: hCaptcha response strips subdomain
**Describe the bug**
When hCaptcha is implemented, the response strips your specified subdomain, resulting in failure.

```bash
# Matrix Authentication Service Logs
2024-11-20T12:59:23.821722Z ERROR
http.server.request {
    otel.kind = "server",
    otel.name = "POST /register",
    network.protocol.name = "http",
    network.protocol.version = "1.1",
    http.request.method = "POST",
    url.path = "/register",
    url.scheme = "http",
    http.route = "/register",
    url.query = "id=01JD4TDY0DZ1ZHQBBN9BJ7B8FH&kind=continue_compat_sso_login",
    user_agent.original = "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/131.0.0.0 Safari/537.36"
}:

handlers.views.register.post:
captcha.verify {
    captcha.service = "HCaptcha",
    captcha.hostname = "ninbura.com",
    captcha.challenge_ts = "2024-11-20T12:59:19.000000Z"
}:

mas_handlers::captcha:
crates/handlers/src/captcha.rs:157:
error = The hostname in the CAPTCHA response ("ninbura.com") does not match the site hostname ("mauth.ninbura.com")
```

**To Reproduce**
1. Configure Matrix Authentication Service to run on a subdomain
2. Configure Site in hCaptcha GUI w/ subdomain in the Domains section (ie mauth.ninbura.com).
3. Attempt to create an account via Matrix Authentication Service
4. See error in Matrix Authentication Service logs and GUI.

**Expected behavior**
hCaptcha response should return full domain including the subdomain.

**Screenshots**
hcaptcha-failure
hcaptcha-site-settings

 - macOS Sequoia v15.1 | Windows 11 Pro N for Workstations v26100.2314
 - Arc (Chromium) | Zen (Gecko)

**Additional context**
I'm running Matrix & Matrix Authentication Server on Ubuntu 24.04 via Docker.

Note that `mauth.ninbura.com` is the *only* domain that I have listed in the hCaptcha Site settings. I've tried adding `ninbura.com` & `matrix.ninbura.com`, as well as creating entirely new Site entries, and refreshing my secrets.

```bash
curl -v -I \
  -H "Host: mauth.ninbura.com" \
  -H "X-Test: ReverseProxyTest" \
  "https://mauth.ninbura.com/register?id=01JD5ZAP1RA12HQX6V5B2F3F7S&kind=continue_compat_sso_login"
```

This will return the headers, which all correctly point to `mauth.ninbura.com` and not `ninbura.com`. The only place I can find `ninbura.com` displayed is in the certificate. I used certbot to create a new cert exclusive to `mauth.ninbura.com` and that did not fix the issue. Happy to share my nginx configuration if needed, but I believe this curl should reveal that it's good.

`ninbura.com` is not hardcoded anywhere in my Matrix repository (Matrix and MAS configurations) or relevant nginx configurations, only `matrix.ninbura.com` and `mauth.ninbura.com`. I have no clue where it could be getting `ninbura.com` outside of something that explicitly Matrix Authentication Service is passing outside of my control. Though, I'd be happy to hear that it's user error and something that I can resolve.

[Here's a link to my full configuration](https://github.com/ninbura-dotcom/matrix/tree/main) for both my Matrix server & Matrix Authentication Service. `synapse.yml` = `homserver.yaml` & `synapse-mas.yml` = `config.yaml`. I use my the `Dockerfile.substitute` image + `entrypoint.sh` script to substitute secrets in each configuration, and copy them into the proper locations/volumes for runtime.

ashleyburgers:
Fixes #3549

This pull request improves the CAPTCHA hostname verification logic to be more robust and accurate, especially when dealing with parent domains returned by some CAPTCHA providers. The main changes introduce a new hostname matching function, update the verification logic to use it, and add tests to ensure correct behavior.

**Hostname verification improvements:**

* Added a new `normalize_hostname` function to ensure hostnames are compared in a case-insensitive manner and without trailing dots.
* Introduced the `hostname_matches_site` function, which allows the CAPTCHA response hostname to match either exactly or as a parent domain of the site hostname. This accommodates providers that return only the registrable domain.
* Updated the CAPTCHA verification logic in the `Form` implementation to use the new `hostname_matches_site` function instead of a strict equality check.

**Testing improvements:**

* Added a new test module to verify the correctness of the hostname matching logic, including cases for exact matches, parent domains, similar suffixes, case insensitivity, and trailing dots.

### DIFF
--- a/crates/handlers/src/captcha.rs
+++ b/crates/handlers/src/captcha.rs
@@ -48,6 +48,29 @@ pub enum Error {
     RequestFailed(#[from] reqwest::Error),
 }
 
+fn normalize_hostname(hostname: &str) -> String {
+    hostname.trim_end_matches('.').to_ascii_lowercase()
+}
+
+/// Checks whether the CAPTCHA response hostname matches the site hostname.
+///
+/// Some CAPTCHA providers (notably hCaptcha) may return only the registrable
+/// domain in the verification response even if the site is hosted on a
+/// subdomain. In that case, the response hostname is a parent domain of the
+/// site hostname and should be accepted.
+fn hostname_matches_site(site_hostname: &str, response_hostname: &str) -> bool {
+    let site = normalize_hostname(site_hostname);
+    let response = normalize_hostname(response_hostname);
+
+    if site == response {
+        return true;
+    }
+
+    // Accept the case where the response contains the parent domain of the site,
+    // e.g. site: "mauth.example.com"; response: "example.com".
+    site.strip_suffix(&format!(".{response}")).is_some()
+}
+
 #[allow(clippy::struct_field_names)]
 #[derive(Debug, Deserialize, Default)]
 #[serde(rename_all = "kebab-case")]
@@ -245,7 +268,7 @@ impl Form {
         span.record("captcha.hostname", &hostname);
         span.record("captcha.challenge_ts", &challenge_ts);
 
-        if hostname != site_hostname {
+        if !hostname_matches_site(site_hostname, &hostname) {
             return Err(Error::HostnameMismatch {
                 expected: site_hostname.to_owned(),
                 got: hostname,
@@ -253,5 +276,32 @@ impl Form {
         }
 
         Ok(())
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::hostname_matches_site;
+
+    #[test]
+    fn hostname_matches_exact() {
+        assert!(hostname_matches_site("mauth.example.com", "mauth.example.com"));
+    }
+
+    #[test]
+    fn hostname_matches_parent_domain() {
+        assert!(hostname_matches_site("mauth.example.com", "example.com"));
+    }
+
+    #[test]
+    fn hostname_does_not_match_similar_suffix() {
+        assert!(!hostname_matches_site("badexample.com", "example.com"));
+        assert!(!hostname_matches_site("mauth.example.com", "example.com"));
+    }
+
+    #[test]
+    fn hostname_matching_is_case_insensitive_and_allows_trailing_dot() {
+        assert!(hostname_matches_site("MAuth.Example.Com.", "example.com"));
+        assert!(hostname_matches_site("mauth.example.com", "EXAMPLE.COM."));
     }
 }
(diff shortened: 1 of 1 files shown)


crates/handlers/src/captcha.rs
```diff
@@ -48,6 +48,29 @@ pub enum Error {
     RequestFailed(#[from] reqwest::Error),
 }
 
+fn normalize_hostname(hostname: &str) -> String {
+    hostname.trim_end_matches('.').to_ascii_lowercase()
+}
+
+/// Checks whether the CAPTCHA response hostname matches the site hostname.
+///
+/// Some CAPTCHA providers (notably hCaptcha) may return only the registrable
```
reivilibre, line 57: do you maybe have any documentation reference from hCaptcha about this behaviour?

crates/handlers/src/captcha.rs
```diff
@@ -245,7 +268,7 @@ impl Form {
         span.record("captcha.hostname", &hostname);
         span.record("captcha.challenge_ts", &challenge_ts);
 
-        if hostname != site_hostname {
+        if !hostname_matches_site(site_hostname, &hostname) {
```
reivilibre, line 271: So in fact, for hCaptcha, you're apparently not meant to  validate the `hostname` field at all.

> Please also note that the hostname field is derived from the user's browser, and should not be used for authentication of any kind; it is primarily useful as a statistical metric. Additionally, in the event that your site experiences unusually high challenge traffic, the hostname field may be returned as "not-provided" rather than the usual value; all other fields will return their normal values.
> — https://docs.hcaptcha.com/#verify-the-user-response-server-side

That doesn't really explain why you only get the apex domain (my colleague who implemented this was using a subdomain and got the subdomain as the response FWIW)

But we should still verify the hostname for reCAPTCHA and Turnstile.
- https://developers.google.com/recaptcha/docs/domain_validation
- https://developers.cloudflare.com/turnstile/get-started/server-side-validation/#implementation

Please would you mind changing your PR to just entirely disable hostname validation but only for hCaptcha?
(& please include the quoted part as a comment for future reference)

reivilibre (changes requested):
Sorry for the long delay & that you hit this trouble. A victim of the myth of drop-in compatible. :)

PR #5406, opened 2026-01-08
Remove imported unsupported threepids when deactivating a user

sandhose:
With syn2mas, we chose to import unsupported threepids (mainly MSISDNs) in a table, but forgot to remove them when a user deactivate their account, which we should do.

Plan is to ship this in the next release, then clean up the table in the release after that.
The table is small enough even on matrix.org for that cleanup to happen in a migration:

```sql
DELETE FROM user_unsupported_third_party_ids
USING users
WHERE users.deactivated_at IS NOT NULL
  AND users.user_id = user_unsupported_third_party_ids.user_id;
```

### DIFF
--- a/crates/storage/src/user/mod.rs
+++ b/crates/storage/src/user/mod.rs
@@ -299,7 +299,7 @@ pub trait UserRepository: Send + Sync {
     /// Delete all the unsupported third-party IDs of a [`User`].
     ///
     /// Those were imported by syn2mas and kept in case we wanted to support
-    /// them later. They still need to be cleaned up when a user deactivate
+    /// them later. They still need to be cleaned up when a user deactivates
     /// their account.
     ///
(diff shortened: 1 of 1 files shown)


crates/storage/src/user/mod.rs
```diff
@@ -295,6 +296,24 @@ pub trait UserRepository: Send + Sync {
     /// Returns [`Self::Error`] if the underlying repository fails
     async fn reactivate(&mut self, user: User) -> Result<User, Self::Error>;
 
+    /// Delete all the unsupported third-party IDs of a [`User`].
+    ///
+    /// Those were imported by syn2mas and kept in case we wanted to support
+    /// them later. They still need to be cleaned up when a user deactivate
```
reivilibre:
```suggestion
    /// them later. They still need to be cleaned up when a user deactivates
```

pedantic

PR #5415, opened 2026-01-14
Cleanup old user registrations from the database

sandhose:
This removes user registrations from the database after 24h. This is fine to do as:

 - the user_emails referencing user registrations are ON DELETE CASCADE
 - the user_registrations are effectively usable for 1h only

### DIFF
--- a/crates/tasks/src/database.rs
+++ b/crates/tasks/src/database.rs
@@ -13,9 +13,10 @@ use async_trait::async_trait;
 use mas_storage::queue::{
     CleanupConsumedOAuthRefreshTokensJob, CleanupExpiredOAuthAccessTokensJob,
     CleanupRevokedOAuthAccessTokensJob, CleanupRevokedOAuthRefreshTokensJob,
-    PruneStalePolicyDataJob,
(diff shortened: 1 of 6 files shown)


crates/tasks/src/database.rs
```diff
@@ -214,6 +215,53 @@ impl RunnableJob for CleanupConsumedOAuthRefreshTokensJob {
     }
 }
 
+#[async_trait]
+impl RunnableJob for CleanupUserRegistrationsJob {
+    #[tracing::instrument(name = "job.cleanup_user_registrations", skip_all)]
+    async fn run(&self, state: &State, context: JobContext) -> Result<(), JobError> {
+        // Remove user registrations after 24h. They are in practice only valid for 1h
+        let until = state.clock.now() - chrono::Duration::hours(24);
```
reivilibre, line 223: I think this is fine but just to double check: is there any reason we might want these longer for debugging purposes?
I suspect they don't have much interesting stuff in there that's not in the logs so probably not

sandhose: I've asked T&S from an abuse perspective, will see what they think. In the meantime, let's keep it like that and adjust if needed

PR #5427, opened 2026-01-19
Cleanup old completed jobs from the database

sandhose:
Add scheduled cleanup job that removes old completed and failed queue jobs after 30 days. Jobs are kept for debugging purposes.

Includes migration to change the next_attempt_id FK constraint from NO ACTION to SET NULL, allowing cleanup of retry chains without breaking foreign key constraints.

One caveat is that cleanup is based on their creation time, *not* when they got completed/failed. This means that if the job takes a long time (as in, several days) to get scheduled, it might get cleared as soon as it runs. This is fine for now, we may want to revisit this if we start scheduling jobs far in the future

### DIFF
--- a/crates/storage-pg/migrations/20260116000003_queue_jobs_next_attempt_set_null.sql
+++ b/crates/storage-pg/migrations/20260116000003_queue_jobs_next_attempt_set_null.sql
@@ -10,4 +10,5 @@ ALTER TABLE queue_jobs
   ADD CONSTRAINT queue_jobs_next_attempt_id_fkey
     FOREIGN KEY (next_attempt_id)
     REFERENCES queue_jobs (queue_job_id)
-    ON DELETE SET NULL;
+    ON DELETE SET NULL
+    NOT VALID;


crates/storage-pg/migrations/20260116000004_queue_jobs_next_attempt_set_null_validate.sql
```diff
@@ -0,0 +1,10 @@
+-- no-transaction
+-- Copyright 2026 Element Creations Ltd.
+--
+-- SPDX-License-Identifier: AGPL-3.0-only OR LicenseRef-Element-Commercial
+-- Please see LICENSE files in the repository root for full details.
+
+-- Validate the FK constraint that was added in the previous migration
+-- This is done in a separate migration to avoid holding locks for too long
+ALTER TABLE queue_jobs
+  VALIDATE CONSTRAINT queue_jobs_next_attempt_id_fkey;
```
reivilibre, line 10: the constraint you added is already validated (didn't use deferred / not valid / whatever it's called)

sandhose: Right, that's why we do code reviews :D 

Fixed in dbf3c351f

PR #5447, opened 2026-01-23
Developer documentation about the various cleanup jobs

sandhose:
This adds a bunch of developper-oriented documentation about cleanup jobs.

It covers the general architecture, the list of jobs we have, why they exist as well as general instructions on how to add a new job.
It also has details on how they interact witb OIDC Backchannel Logout, as the user sessions cleanup is not really obvious because of that.

### DIFF
--- a/docs/development/cleanup-jobs.md
+++ b/docs/development/cleanup-jobs.md
@@ -1,6 +1,6 @@
 # Cleanup Jobs
 
-In MAS, most of the data only soft-deleted, through setting a `deleted_at`, `finished_at`, `consumed_at` timestamp on the row, instead of actually deleting the row.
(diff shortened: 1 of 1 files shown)


docs/development/cleanup-jobs.md
```diff
@@ -0,0 +1,266 @@
+# Cleanup Jobs
+
+In MAS, most of the data only soft-deleted, through setting a `deleted_at`, `finished_at`, `consumed_at` timestamp on the row, instead of actually deleting the row.
```
reivilibre:
```suggestion
In MAS, most of the data are initially only soft-deleted, by setting a `deleted_at`, `finished_at`, `consumed_at` timestamp on the row, instead of actually deleting the row.
```

PR #5448, opened 2026-01-23
Clear out last active IP on each sessions after 30 days

sandhose:
This adds three new scheduled cleanup jobs that clear the last_active_ip
field from sessions that have been inactive for more than 30 days:

- CleanupInactiveOAuth2SessionIpsJob
- CleanupInactiveCompatSessionIpsJob
- CleanupInactiveUserSessionIpsJob

This helps with data minimization by not retaining IP addresses longer


crates/storage-pg/src/compat/session.rs
```diff
@@ -758,4 +758,48 @@ impl CompatSessionRepository for PgCompatSessionRepository<'_> {
...
+        limit: usize,
+    ) -> Result<usize, Self::Error> {
+        let res = sqlx::query_scalar!(
+            r#"
+                WITH to_update AS (
+                    SELECT compat_session_id
+                    FROM compat_sessions
+                    WHERE last_active_ip IS NOT NULL
+                      AND last_active_at IS NOT NULL
+                      AND last_active_at < $1
```
reivilibre: since we're doing this in limited batches, should we not also paginate from `$2 < last_active_at`, to skip over dead tuples?
Applies to all 3

sandhose: Have done that in  dbdb297

PR #5449, opened 2026-01-25
Add `password_file` config variant for `email` and `database`

networkException:


crates/config/src/sections/email.rs
```diff
@@ -136,8 +145,14 @@ impl EmailConfig {
 
     /// Password for use to authenticate when connecting to the SMTP server
     #[must_use]
-    pub fn password(&self) -> Option<&str> {
-        self.password.as_deref()
+    pub fn password(&self) -> Option<String> {
+        if let Some(password_file) = &self.password_file {
+            return std::fs::read_to_string(password_file)
+                .ok()
```
sandhose: This should be async and bubble up read error correctly, like the database side does

sandhose (changes requested):

crates/config/src/sections/email.rs
networkException: Database `password_file` isn't async, should I also convert that? I'm not sure it's worth the overhead

sandhose: Yes please, we should avoid doing sync I/O once the Tokio runtime started as much as possible, even if it's just during startup

networkException:
I pushed to resolve conflicts in the config schema. The CI failure seems to be unrelated

sandhose:
> The CI failure seems to be unrelated

That is correct, fixed that in https://github.com/element-hq/matrix-authentication-service/pull/5876

sandhose (approved):

PR #5451, opened 2026-01-26
Add `syn2mas` flag to ignore missing auth providers

jaywink:
Currently `syn2mas` will always error in the Synapse checks phase if it finds auth providers in the `user_external_ids` database table, that are not configured in Synapse config. While normally this the right thing to do, we may have situations where we know what we're doing, and want to ignore invalid looking data in the external identifiers table. If the flag is given, ignore errors and output them as warnings instead.

### DIFF
--- a/crates/cli/src/commands/syn2mas.rs
+++ b/crates/cli/src/commands/syn2mas.rs
@@ -61,6 +61,14 @@ pub(super) struct Options {
     /// configure all values through those environment variables.
     #[clap(long = "synapse-database-uri", global = true)]
     synapse_database_uri: Option<PgConnectOptions>,
+
+    /// Make missing auth providers in Synapse config warnings instead of
+    /// errors. If this flag is set, and we find `auth_provider` values in
+    /// the Synapse `user_external_ids` table, that are not configured in
+    /// the Synapse OIDC configuration, instead of erroring we will just
+    /// output warnings.
+    #[clap(long = "ignore-missing-auth-providers", global = true)]
+    ignore_missing_auth_providers: bool,
 }
 
 #[derive(Parser, Debug)]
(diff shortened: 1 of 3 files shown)


crates/syn2mas/src/migration.rs
```diff
@@ -141,6 +141,7 @@ struct MigrationState {
 /// - An underlying database access error, either to MAS or to Synapse.
 /// - Invalid data in the Synapse database.
 #[expect(clippy::implicit_hasher)]
+#[allow(clippy::too_many_arguments)]
```
jaywink, line 144: I felt that the arbitrary "max 7 arguments", while that feels  like a completely sane thing to have, probably doesn't justify refactoring things to avoid adding a new argument here? Open to suggestions though.

crates/cli/src/commands/syn2mas.rs
```diff
@@ -61,6 +61,14 @@ pub(super) struct Options {
...
     #[clap(long = "synapse-database-uri", global = true)]
     synapse_database_uri: Option<PgConnectOptions>,
+
+    /// Make missing auth providers in Synapse config warnings instead of
+    /// errors. If this flag is set, and we find `auth_provider` values in
+    /// the Synapse `user_external_ids` table, that are not configured in
+    /// the Synapse OIDC configuration, instead of erroring we will just
+    /// output warnings.
+    #[clap(long = "ignore-missing-auth-providers", global = true)]
+    ignore_missing_auth_providers: bool,
```
sandhose, line 71: Since we usually know exactly which provider we want to ignore (I think?), could we instead make this a list instead of a boolean?

Something like

```
    #[arg(long = "skip-auth-provider", global = true, action = clap::ArgAction::Append)]
    skip_auth_providers: Vec<String>,
```

That allows things to get repeated, e.g. `--skip-auth-provider foo --skip-auth-provider bar`

sandhose: Discussed out of band, this is not going to really fit our use case, so going with the global flag

crates/syn2mas/src/migration.rs
sandhose: It is kinda arbitrary, fine like that :)

PR #5456, opened 2026-01-27
Trim value when reading `client_secret_file`

networkException:
Follow up to #4882

### DIFF
--- a/crates/config/src/sections/mod.rs
+++ b/crates/config/src/sections/mod.rs
@@ -345,7 +345,7 @@ impl ClientSecret {
     /// Returns an error when the client secret could not be read from file.
     pub async fn value(&self) -> anyhow::Result<String> {
         Ok(match self {
-            ClientSecret::File(path) => tokio::fs::read_to_string(path).await?,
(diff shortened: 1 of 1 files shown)


crates/config/src/sections/mod.rs
```diff
@@ -345,7 +345,7 @@ impl ClientSecret {
     /// Returns an error when the client secret could not be read from file.
     pub async fn value(&self) -> anyhow::Result<String> {
         Ok(match self {
-            ClientSecret::File(path) => tokio::fs::read_to_string(path).await?,
+            ClientSecret::File(path) => tokio::fs::read_to_string(path).await?.trim().to_string(),
```
reivilibre:
```suggestion
            ClientSecret::File(path) => tokio::fs::read_to_string(path).await?.trim().to_owned(),
```

(This is just a personal preference)

reivilibre (approved):
Thanks!

crates/config/src/sections/mod.rs
networkException: I'm never quite sure which to use

PR #5474, opened 2026-02-04
Split multi-arch Docker build into parallel jobs

sandhose:
- Modify Dockerfile to build single architecture based on TARGETARCH instead of cross-compiling both targets in one run
- Replace single build-image job with matrix job (amd64, arm64)
- Add finalize-image job that creates multi-arch manifests using `docker buildx imagetools create` and signs the final images
- Each architecture gets its own build cache
- Centralise the `docker/metadata-action` invocations in a single `compute-image-meta` job and pass labels to the per-arch builds via an uploaded bake-file artifact

This enables parallel builds of each architecture, reducing total build time by running both simultaneously rather than sequentially.

Small drive-by fixes:
- Override `org.opencontainers.image.licenses` with `AGPL-3.0-only OR LicenseRef-Element-Commercial`. GitHub's Licensee detection only surfaces the legacy `AGPL-3.0` SPDX identifier and can't represent the project's dual licensing
- Apply OCI annotations to the image index (manifest list) via `docker buildx imagetools create --annotation index:...`, which the current image lacks. Should make tools like renovate detect image metadata better

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -464,6 +464,7 @@ jobs:
         with:
           pattern: digests-*
           path: /tmp/digests
+          # Collect digests from both amd64 and arm64 builds
           merge-multiple: true
 
       - name: Setup Cosign
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -28,7 +28,9 @@ env:
   SCCACHE_GHA_ENABLED: "true"
   RUSTC_WRAPPER: "sccache"
   BUILDCACHE: ghcr.io/element-hq/matrix-authentication-service/buildcache
-  DOCKER_METADATA_ANNOTATIONS_LEVELS: manifest,index
+  # We only ever apply annotations to the image index (manifest list) during
+  # `finalize-image`; per-arch manifests don't get any.
+  DOCKER_METADATA_ANNOTATIONS_LEVELS: index
 
 jobs:
   compute-version:
@@ -216,8 +218,10 @@ jobs:
       contents: read
 
     outputs:
-      regular-json: ${{ steps.meta.outputs.json }}
-      debug-json: ${{ steps.meta-debug.outputs.json }}
+      regular-tags: ${{ steps.meta.outputs.tags }}
+      regular-annotations: ${{ steps.meta.outputs.annotations }}
+      debug-tags: ${{ steps.meta-debug.outputs.tags }}
+      debug-annotations: ${{ steps.meta-debug.outputs.annotations }}
 
     steps:
       - name: Docker meta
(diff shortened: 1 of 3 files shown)


.github/workflows/build.yaml
```diff
@@ -247,6 +239,15 @@ jobs:
             type=semver,pattern={{major}}.{{minor}}
             type=semver,pattern={{major}}
             type=sha
+          # GitHub's license detection (via Licensee) can only report a single
```
reivilibre, line 248: where does GitHub's licence detection fit into this?

.github/workflows/build.yaml
```diff
@@ -337,35 +397,185 @@ jobs:
...
-        # metadata can exceed the shell ARG_MAX limit when expanded.
-        id: metadata
+            *.platform=linux/${{ matrix.arch }}
+            *.output=type=image,"name=${{ env.IMAGES }}",push-by-digest=true,name-canonical=true,push=true
+            *.cache-from=type=registry,ref=${{ env.BUILDCACHE }}:buildcache-${{ matrix.arch }}
+            *.cache-to=type=registry,ref=${{ env.BUILDCACHE }}:buildcache-${{ matrix.arch }},mode=max
+
+      # We use github-script rather than shelling out to jq because the bake
+      # metadata can exceed the shell ARG_MAX limit when inherited as an env
+      # var by an exec'd jq.
```
reivilibre, line 416: if that's the only reason, why not use a here-string to pass the var in to jq over stdin?

But maybe a proper scripting language is fine.

.github/workflows/build.yaml
```diff
@@ -337,35 +397,185 @@ jobs:
...
+    outputs:
+      metadata: ${{ steps.output.outputs.metadata }}
+
+    permissions:
+      contents: read
+      packages: write
+      id-token: write
+
+    steps:
+      - name: Download digests
```
reivilibre, line 462: probably want to signpost that the `merge-multiple` here  is how you get the merging of manifests between the other jobs

.github/workflows/build.yaml
```diff
@@ -337,35 +397,185 @@ jobs:
...
+            ARGS+=(-t "$tag")
+          done < <(jq -r '.tags[]' <<< "$META_JSON")
+          while IFS= read -r annotation; do
+            ARGS+=(--annotation "$annotation")
+          done < <(jq -r '
+            .annotations
+            | map(select(startswith("index:")))
+            | map(select(endswith("=") | not))
+            | .[]
+          ' <<< "$META_JSON")
```
reivilibre: this is a bit monstruous however

reivilibre (approved):
Probably fine if it works though

.github/workflows/build.yaml
sandhose: because the default docker metadata thingy will fill in fields like the description and the license info out of the GitHub API

.github/workflows/build.yaml
sandhose: because zizmor doesn't like this, as it's possibly an arbitrary code injection vulnerability. Inputs should ideally only be provided through files or env variables

.github/workflows/build.yaml
sandhose: I've made that less horrible in [63deb0b](https://github.com/element-hq/matrix-authentication-service/pull/5474/commits/63deb0b1fdba9cfeee9e19455b06cef175810388)

PR #5516, opened 2026-02-23
Support MSC4198 login_hint in account management uri

tonkku107:
Equivalent implementation to #3343 but for the account management uri.

Code had to be moved around a bit to support both

### DIFF
--- a/crates/handlers/src/views/app.rs
+++ b/crates/handlers/src/views/app.rs
@@ -62,11 +62,9 @@ pub async fn get(
     let Some(session) = maybe_session else {
         let mut url = mas_router::Login::and_then(PostAuthAction::manage_account(action));
 
-        url = if let Some(login_hint) = unstable_login_hint {
-            url.with_login_hint(login_hint)
-        } else {
-            url
-        };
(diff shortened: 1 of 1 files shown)


crates/handlers/src/views/app.rs
```diff
@@ -54,13 +60,15 @@ pub async fn get(
...
-            )),
-        )
-            .into_response());
+        let mut url = mas_router::Login::and_then(PostAuthAction::manage_account(action));
+
+        url = if let Some(login_hint) = unstable_login_hint {
+            url.with_login_hint(login_hint)
+        } else {
+            url
+        };
```
sandhose: Either shadow the variable or have it mutable

```suggestion
        let url = mas_router::Login::and_then(PostAuthAction::manage_account(action));

        let url = if let Some(login_hint) = unstable_login_hint {
            url.with_login_hint(login_hint)
        } else {
            url
        };
```

or

```suggestion
        let mut url = mas_router::Login::and_then(PostAuthAction::manage_account(action));

        if let Some(login_hint) = unstable_login_hint {
            url = url.with_login_hint(login_hint);
        }
```

PR #5538, opened 2026-03-04
Use system-configured proxies for outgoing HTTP requests

weareoutman:
Enable the system-proxy feature, so that when the MAS server make a request to an OAuth provider, it can respect the system proxy config. This is essential for restricted environments.


sandhose (approved):
I was surprised we didn't already enable that, but because we disable the default features of reqwest it was disabled.

PR #5550, opened 2026-03-13
Synchronise the user `locked` flag to Synapse

reivilibre:
Companion PR: https://github.com/element-hq/synapse/pull/19554

### DIFF
--- a/crates/matrix/src/mock.rs
+++ b/crates/matrix/src/mock.rs
@@ -10,9 +10,10 @@ use anyhow::Context;
 use async_trait::async_trait;
 use tokio::sync::RwLock;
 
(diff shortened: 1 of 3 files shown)


crates/matrix-synapse/src/modern.rs
sandhose, line 1: Would be nice if we also had a `synapse_legacy` implementation, maybe some configurations are still using that, but also we should drop that one.

crates/matrix/src/mock.rs
sandhose, line 1: I actually would appreciate if the mock connection saved the lock state on users so that we can actually test this behaviour

PR #5553, opened 2026-03-16
Convert ViolationCode into ViolationVariant to allow adding fields on each variant

reivilibre:
We'll need the policy to be able to send additional fields for some types of validations.

Concretely, for the [device](https://github.com/element-hq/matrix-authentication-service/issues/4339) [limiting](https://github.com/element-hq/backend-internal/issues/199) stuff, we'll need the policy to be able to tell MAS how many sessions should be removed before the user can log in.
This way the UI will be able to guide the user through it.

Intended to look something like:
```rust
ViolationVariant::TooManySessions {
    /// How many devices to remove
    need_to_remove: u32
}
```

### DIFF
--- a/crates/policy/src/model.rs
+++ b/crates/policy/src/model.rs
@@ -82,6 +82,10 @@ pub struct Violation {
     pub redirect_uri: Option<String>,
     pub field: Option<String>,
 
+    // We flatten as policies expect `code` as another top-level field.
+    //
+    // This also means all of the extra fields from the variant will be splatted at this
+    // level which is fine (arbitrary).
     #[serde(flatten)]
     pub variant: Option<ViolationVariant>,
 }


crates/policy/src/model.rs
```diff
@@ -80,7 +81,9 @@ pub struct Violation {
     pub msg: String,
     pub redirect_uri: Option<String>,
     pub field: Option<String>,
-    pub code: Option<Code>,
+
+    #[serde(flatten)]
+    pub variant: Option<ViolationVariant>,
```
MadLittleMods, line 90: Why flatten? Perhaps this was to maintain the `code` as a top-level field. Seems normal to nest whatever extra under `variant` as this doesn't seem like it matters.

Perhaps just a preference thing and need the original author/MAS taste-maker to weigh in

MadLittleMods: I guess the placement of `code` may actually matter as this interacts with actual policies like 

https://github.com/element-hq/matrix-authentication-service/blob/9927382400588aaf93ee501b8f3d9493e32fb01a/policies/authorization_grant/authorization_grant.rego#L157-L160

https://github.com/element-hq/matrix-authentication-service/blob/9927382400588aaf93ee501b8f3d9493e32fb01a/policies/compat_login/compat_login.rego#L28-L31

MadLittleMods: Added a comment to explain this detail (https://github.com/element-hq/matrix-authentication-service/pull/5553/commits/7b02748af42287f78bbf3117f579078345439d88). Comment could probably be better.

crates/policy/src/model.rs
```diff
@@ -80,7 +81,13 @@ pub struct Violation {
     pub msg: String,
     pub redirect_uri: Option<String>,
```
sandhose, line 82: We could look into migrating some of the fields (like `redirect_uri`) to proper violation variants, but can be done in another PR

MadLittleMods:
Thanks for the review @sandhose and unburying this PR @reivilibre 🦈

PR #5580, opened 2026-03-24
Add warning/note when removing devices

MadLittleMods:
Add warning/note when removing devices

Actual MAS showing off the remove device dialog

There are a few differences from the design but this is just using the standard/existing components in MAS so I assume it's fine :fast_forward: 

### Design

Design: https://www.figma.com/design/HUysJAhv6cK6p1Pc81Fxaa/Matrix-Authentication-Service--MAS-?node-id=6008-13220&t=QkbCHZBAandZ4k5I-4

Figma design: Remove device dialog

### Dev notes

Per the [discussion](https://www.figma.com/design/HUysJAhv6cK6p1Pc81Fxaa?node-id=6008-10017#1639965314) in the Figma design, in the future, we probably want to copy the flows from the [recovery key usability work in Element](https://www.figma.com/design/hDUT4q0pynhkbmB5s56GN8/ER-162--Recovery-Key-Usability---Management?node-id=0-1&p=f&t=n9oKZDkXNWJJB7ts-0) to expand our warning to give more context and actual advice.

Essentially breaking it up to 3 different flows:

 - Recovery setup or another device available.
 - No other devices but recovery set up.
 - No other devices and no recovery.
 
For example, we can offer to setup recovery if not already setup or check your recovery keys.

### DIFF
--- a/frontend/locales/pt-BR.json
+++ b/frontend/locales/pt-BR.json
@@ -125,7 +125,7 @@
       "heading": "The email address {{email}} is already in use."
     },
     "end_session_button": {
-      "confirmation_modal_title": "Are you sure you want to remove this device?",
+      "confirmation_modal_title": "Are you sure you want to end this session?",
       "text": "Remove device"
     },
     "error": {
--- a/frontend/locales/sk.json
+++ b/frontend/locales/sk.json
@@ -125,7 +125,7 @@
       "heading": "The email address {{email}} is already in use."
     },
     "end_session_button": {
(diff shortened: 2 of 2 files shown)


frontend/locales/pt-BR.json
sandhose, line 1: Those files are automatically managed by Localazy, they should never be modified manually

frontend/locales/en.json
sandhose, line 1: Except of course the 'source' language

PR #5587, opened 2026-03-27
fix: trim values read from client secret files

uku3lig:
Avoids needing to forcefully save the file without an ending newline, which can cause very confusing errors due to the client secret simply not matching because of additional whitespace.


reivilibre:
Fixed concurrently by #5456, thanks though!

uku3lig:
Oh, just realized that this other pr existed, I didn't see it when making mine :-) Thanks for taking a look and merging theirs!

PR #5603, opened 2026-04-03
Refactor login form to include password input conditionally

gordnhoo:
This removes the username login box when only providers are chosen


sandhose (approved):
Thanks for doing this, sorry for the delay reviewing this

PR #5605, opened 2026-04-03
Per-provider registration token requirement

### LINKED ISSUE #5383: Creating user with SSO also asks registation token
**Describe the bug**
We have enabled both password login with registation token and SSO, current MAS version asks registation token when using SSO which it should not.

**To Reproduce**
1. Have MAS config as listed below
2. Try to create user (login) with SSO
3. In the last step MAS asks for registration token

**Expected behavior**
When using SSO, no registation token is expected.

**Screenshots**
Fine up to this point:

Image

After clicking the "Create user" -button, this happens when should not:

Image

Using MAS 1.8.0

**Additional context**
MAS config snippet for relevant part:

```
passwords:
  enabled: true
  schemes:
  - version: 1
    algorithm: bcrypt
  - version: 2
    algorithm: argon2id
  minimum_complexity: 3
matrix:
  kind: synapse
  homeserver: hacklab.fi
  secret: <redacted>
  endpoint: "http://localhost:8008"
upstream_oauth2:
  providers:
    - id: "<redacted>"
      issuer: "https://sso.hacklab.fi/realms/hacklabfi"
      human_name: "Hacklab Finland SSO"
      brand_name: "Hacklab.fi"
      token_endpoint_auth_method: client_secret_basic
      client_id: "matrix-authentication-service"
      client_secret: "<redacted>"
      scope: "openid profile email"
      claims_imports:
        localpart:
          action: ignore
          template: "{{ user.preferred_username }}"
        displayname:
          action: suggest
          template: "{{ user.name }}"
        email:
          action: suggest
          template: "{{ user.email }}"
          set_email_verification: always
account:
  registration_token_required: true
  password_registration_enabled: true
```

defaultdino:
closes #5383


crates/config/src/sections/account.rs
```diff
@@ -86,8 +86,12 @@ pub struct AccountConfig {
...
     /// password registration. This has no effect if password registration
     /// is disabled.
-    #[serde(default = "default_false", skip_serializing_if = "is_default_false")]
-    pub registration_token_required: bool,
+    #[serde(
+        default = "default_false",
+        skip_serializing_if = "is_default_false",
+        alias = "registration_token_required"
+    )]
+    pub password_registration_token_required: bool,
```
sandhose: I'm slightly concerned with changing (again) the behavior if `registration_token_required` is set during an upgrade.

If it's not too annoying, I would keep a separate `registration_token_required` config knob but mark it as deprecated, and emit a warning on startup. In a later version, we would then make this config option error out and refuse to start.

During the transition, it would propagate down to `SiteConfig` and act as an override when we determine whether we need a registration token or not

crates/handlers/src/admin/v1/site_config.rs
```diff
@@ -26,7 +26,7 @@ pub struct SiteConfig {
     pub password_registration_email_required: bool,
 
     /// Whether registration tokens are required for password registrations.
-    pub registration_token_required: bool,
+    pub password_registration_token_required: bool,
```
sandhose, line 28: This is a breaking change to a public interface, we should also apply some transition period or something. This is especially relevant for Element Admin, as it shows the registration token UI only if they are enabled, but maybe we should change that there.

The old attribute can be marked as deprecated so that it shows up like that in the API docs

crates/handlers/src/views/register/steps/finish.rs
```diff
@@ -120,8 +120,37 @@ pub(crate) async fn get(
...
+    let token_required = if registration
+        .upstream_oauth_authorization_session_id
+        .is_some()
+    {
+        // load provider to let OAuth registrations check its
+        // per provider `registration_token_required` flag
+        let session_id = registration
+            .upstream_oauth_authorization_session_id
+            .context("Missing upstream OAuth session ID")
+            .map_err(InternalError::from_anyhow)?;
```
sandhose:
```suggestion
    let token_required = if let Some(session_id) = registration.upstream_oauth_authorization_session_id {
        // load provider to let OAuth registrations check its
        // per provider `registration_token_required` flag
```

crates/storage-pg/migrations/20260403144138_upstream_oauth2_providers_registration_token_required.sql
```diff
@@ -0,0 +1,4 @@
+-- Adds a `registration_token_required` column to the UpstreamOauthProvider table
```
sandhose, line 6: Could you add a copyright header similar to the other migration files?

sandhose (commented):
Overall looks good, just slightly concerned with breaking changes in behavior in the config and in the admin API

defaultdino:
@sandhose Thanks, I will look into this in the coming week

defaultdino:
I believe this is now ready for another look

sandhose (approved):
Sorry for the delay on the review, thanks for this!

PR #5606, opened 2026-04-03
Defer session creation to token exchange time

### LINKED ISSUE #5276: OAuth sessions are created before the authorization grant gets actually completed
For some reason, we're creating the OAuth session (but not the access token nor the device on the HS side) when we fulfill the grant, but not when the token exchanges.

This leads to a weird UX where the session starts appearing in your list even though you haven't completed it

defaultdino:
closes #5276 

the `OAuth2` session is no longer created at consent time. Instead,

- Consent only records the browser session that approved the grant
- Token exchange creates the OAuth2 session, like the device code flow already does

this means sessions won't appear in the user's list until the client actually completes the token exchange.


crates/handlers/src/oauth2/token.rs
```diff
@@ -520,16 +504,35 @@ async fn authorization_code_grant(
...
-    let Some(user_session_id) = session.user_session_id else {
-        tracing::warn!("No user session associated with this OAuth2 session");
-        return Err(RouteError::InvalidGrant(authz_grant.id));
-    };
-
+    // look up the browser session that approved this grant
     let browser_session = repo
         .browser_session()
-        .lookup(user_session_id)
+        .lookup(browser_session_id)
```
sandhose, line 517: We need to check here that the session is still active, because of how it interacts with features like OIDC Backchannel Logout.

Backchannel Logout might be configured to end both browser sessions *and* client sessions. If we don't check this, we might end up with client sessions not logged out by a backchannel logout if we receive it between the redirect and the code exchange.

crates/handlers/src/oauth2/authorization/consent.rs
```diff
@@ -350,7 +340,7 @@ pub(crate) async fn post(
     repo.save().await?;
 
     activity_tracker
-        .record_oauth2_session(&clock, &session)
+        .record_browser_session(&clock, &browser_session)
```
sandhose: This is already done earlier in this handler

crates/storage-pg/migrations/20260403000001_authz_grant_user_session_id.sql
```diff
@@ -0,0 +1,20 @@
...
+UPDATE "oauth2_authorization_grants" AS g
+SET "user_session_id" = s."user_session_id"
+FROM "oauth2_sessions" AS s
+WHERE g."oauth2_session_id" = s."oauth2_session_id"
+  AND g."fulfilled_at" IS NOT NULL;
+
+UPDATE "oauth2_authorization_grants"
+SET "oauth2_session_id" = NULL
+WHERE "fulfilled_at" IS NOT NULL
+  AND "exchanged_at" IS NULL;
```
sandhose: This is a bit of a problem, as this will lock the table until the update completes. It also makes it impossible to safely rollout this new version, as older version of MAS might be running with the updated schema.

It would be better to *not* have this update, and take into account the potential inconsistent state in the app code

crates/storage-pg/migrations/20260403000001_authz_grant_user_session_id.sql
```diff
@@ -0,0 +1,20 @@
+-- Copyright 2026 Element Creations Ltd.
+--
+-- SPDX-License-Identifier: AGPL-3.0-only OR LicenseRef-Element-Commercial
+-- Please see LICENSE files in the repository root for full details.
+
+ALTER TABLE "oauth2_authorization_grants"
```
sandhose, line 6: This is missing an index, so deleting an `user_session` will trigger a full table scan of `oauth2_authorization_grants`

crates/storage-pg/src/oauth2/authorization_grant.rs
```diff
@@ -77,23 +78,29 @@ impl TryFrom<GrantLookup> for AuthorizationGrant {
             value.fulfilled_at,
             value.exchanged_at,
             value.cancelled_at,
+            value.user_session_id,
             value.oauth2_session_id,
         ) {
-            (None, None, None, None) => AuthorizationGrantStage::Pending,
-            (Some(fulfilled_at), None, None, Some(session_id)) => {
+            (None, None, None, None, None) => AuthorizationGrantStage::Pending,
```
sandhose, line 84: I think this is where instead of backfilling, if we're seeing a grant fulfilled but without a user session ID, we should consider it as 'cancelled' or invalid, and have the user re-do the flow

sandhose (changes requested):
This is very very similar to what I had in #5515 

If you prefer I can try to revive that PR, or you can fix the few things I flagged in this review, lmk!

defaultdino:
Thanks! I'll have a look at the things you flagged in the PR and fix them.

defaultdino:
I think I've addressed the things you commented on, just let me know if I missed anything

crates/handlers/src/oauth2/token.rs
```diff
@@ -527,16 +511,36 @@ async fn authorization_code_grant(
...
         .await?
-        .ok_or(RouteError::NoSuchBrowserSession(user_session_id))?;
+        .filter(BrowserSession::active) // this treats inactive sessions as 'no such browser session' errors
+        .ok_or(RouteError::NoSuchBrowserSession(browser_session_id))?;
```
sandhose: Problem is that `RouteError::NoSuchBrowserSession` a 500 server error, which is then sent to Sentry, etc.; but this branch is very much something a regular user can trigger with a relatively simple race, so it should be treated as a regular 'invalid grant' error

crates/storage-pg/migrations/20260403000001_authz_grant_user_session_id.sql
```diff
@@ -0,0 +1,11 @@
...
+  ADD COLUMN "user_session_id" UUID
+    CONSTRAINT "oauth2_authorization_grants_user_session_id_fkey"
+    REFERENCES "user_sessions" ("user_session_id");
+
+CREATE INDEX "oauth2_authorization_grants_user_session_id_idx" ON "oauth2_authorization_grants" (user_session_id);
```
sandhose: This index should be created `CONCURRENTLY` in a separate migration with `-- no-transaction` set.

This avoids taking an exclusive lock on the table during the migration. Because this will run outside of a transaction (`CREATE INDEX CONCURRENTLY` can't be rolled back), we'll need to also make it `IF NOT EXISTS`, as it might be run multiple times if the service crashes midway


```sql
-- no-transaction
-- …usual copyright & comment
CREATE INDEX CONCURRENTLY IF NOT EXISTS
    "oauth2_authorization_grants_user_session_id_idx"
    ON "oauth2_authorization_grants" (user_session_id);
```

…one day I'll document all those quirks 🙈

crates/storage-pg/src/oauth2/authorization_grant.rs
```diff
@@ -77,23 +78,38 @@ impl TryFrom<GrantLookup> for AuthorizationGrant {
...
+                    browser_session_id: browser_session_id.into(),
                     fulfilled_at,
                 }
             }
-            (Some(fulfilled_at), Some(exchanged_at), None, Some(session_id)) => {
-                AuthorizationGrantStage::Exchanged {
-                    session_id: session_id.into(),
-                    fulfilled_at,
-                    exchanged_at,
+            (Some(fulfilled_at), None, None, None, None) => {
```
sandhose: This branch would not get hit by the old version of the service, as the old fulfill would fill in the `oauth2_session_id`. 


```suggestion
            (Some(fulfilled_at), None, None, None, Some(_)) => {
```

or if we want to match mode widely (but might hide some other bugs?):


```suggestion
            (Some(fulfilled_at), _, _, None, _) => {
```

sandhose (changes requested):
Almost there, thanks for pursuing this

crates/storage-pg/migrations/20260403000001_authz_grant_user_session_id.sql
```diff
@@ -0,0 +1,11 @@
+-- Copyright 2026 Element Creations Ltd.
+--
+-- SPDX-License-Identifier: AGPL-3.0-only OR LicenseRef-Element-Commercial
+-- Please see LICENSE files in the repository root for full details.
+
+ALTER TABLE "oauth2_authorization_grants"
+  ADD COLUMN "user_session_id" UUID
+    CONSTRAINT "oauth2_authorization_grants_user_session_id_fkey"
+    REFERENCES "user_sessions" ("user_session_id");
```
sandhose: That FK 'action' should probably `ON DELETE SET NULL` so that it doesn't block the regular session cleanup we got on `user_sessions`. It will fall under the 'finished_at but no user_session_id branch' which considers the grant as cancelled

defaultdino:
Thanks for the quick correspondence! I did some more changes, have a look when you can

crates/data-model/src/oauth2/authorization_grant.rs
```diff
@@ -62,11 +62,12 @@ pub enum AuthorizationGrantStage {
     #[default]
     Pending,
     Fulfilled {
-        session_id: Ulid,
+        browser_session_id: Ulid,
         fulfilled_at: DateTime<Utc>,
     },
     Exchanged {
         session_id: Ulid,
+        browser_session_id: Ulid,
```
sandhose: Now that the foreign key on `user_session_id` is set to `ON DELETE SET NULL`, we might get in a state where the auth grant is exchanged, but the `user_session_id` is NULL.

This is problematic, because when we load the grant from the database, if we've got `fulfilled_at` and `exchanged_at` set, but `user_session_id` NULL, you'll get a 'database inconsistency' error, even though it's a perfectly valid state

Now, because we don't seem to read the `browser_session_id` anywhere in that state, I think it would be better to just not put this field in the 'Exchanged' state. Alternatively, we can make it optional

crates/storage-pg/src/oauth2/authorization_grant.rs
```diff
@@ -77,23 +78,41 @@ impl TryFrom<GrantLookup> for AuthorizationGrant {
...
+                    "Grant was fulfilled without a user_session_id, it will be treated as cancelled"
+                );
+                AuthorizationGrantStage::Cancelled {
+                    cancelled_at: fulfilled_at,
                 }
             }
-            (None, None, Some(cancelled_at), None) => {
+            (Some(fulfilled_at), None, None, None, None) => AuthorizationGrantStage::Cancelled {
+                cancelled_at: fulfilled_at,
+            },
```
sandhose: this branch should either be merged with the one up there or removed. It *technically* can be reached if the server is offline for 30 days, and the finished user session gets cleaned up *before* the oauth grant… but that's never really gonna happen

sandhose (changes requested):

sandhose (approved):
Thanks for this!

Issue #2491, opened 2024-09-10
Errors when loading the configuration file should be easier to read


matrixbot:
*This issue was originally created by [**@sandhose**](https://github.com/sandhose) at <https://github.com/matrix-org/matrix-authentication-service/issues/2491>.*

Because we use `figment` to handle configuration, we sometimes have really weird error messages. See #2477 for example.

A few things we should do:

 - avoid `#[serde(flatten)]`, as it breaks error reporting
 - [tag](https://docs.rs/figment/latest/figment/value/magic/struct.Tagged.html) important places of the config so that we can 
 - do configuration post-processing instead
 - actually test some misconfiguration scenarios

PR #5607, opened 2026-04-03
Session hard limit automatic eviction in non-interactive contexts

MadLittleMods:
Session hard limit automatic eviction in non-interactive contexts.

#### Legacy compatibility login

 1. Start Synapse, `poetry run synapse_homeserver --config-path homeserver.yaml`
 1. Configure MAS with a `session_limit` -> `hard_limit`
    ```yaml
    experimental:
      session_limit:
        soft_limit: 1
        hard_limit: 2
    ```
 1. Start MAS, `cargo run -- server -c config.yaml` ([contributing guide](https://element-hq.github.io/matrix-authentication-service/development/contributing.html#5-build-and-run-mas))
 1. Register a user if you don't have one already (`cargo run -- manage register-user`)
 1. Send a couple login requests until you get an error
   ```shell
   curl -X POST http://localhost:8080/_matrix/client/v3/login \
    -H "Content-Type: application/json" \
    -d '{
      "identifier": {
        "type": "m.id.user",
        "user": "some-test-user"
      },
      "initial_device_display_name": "Jungle Phone",
      "password": "xxx",
      "type": "m.login.password"
    }'
  ```
 1. Expect `403 Forbidden`: `{"errcode":"M_FORBIDDEN","error":"You have reached your hard device limit. Please visit your account page to sign some out."}`

Test automatic eviction:
 
 1. Configure MAS with with a `session_limit` -> `hard_limit_eviction`
    ```yaml
    experimental:
      session_limit:
        soft_limit: 1
        hard_limit: 2
        hard_limit_eviction: true
    ```
 1. Restart MAS
 1. Send another login request
 1. Notice how the login works
 1. If you inspect http://localhost:8080/account/sessions, the oldest sessions were removed.

### Todo

 - [x] Figure out ordering, https://github.com/element-hq/matrix-authentication-service/pull/5607#discussion_r3041621109
 - [x] TODO Grab `need_to_remove` from violation
 - [x] Wait for merge/direction on https://github.com/element-hq/matrix-authentication-service/pull/5611

### Dev notes

`POST http://localhost:8080/_matrix/client/v3/login`
```
{
  "identifier": {
    "type": "m.id.user",
    "user": "some-test-user"
  },
  "initial_device_display_name": "Jungle Phone",
  "password": "xxx",
  "type": "m.login.password"
}
```

---

Compatibility sessions vs OAuth 2.0 sessions:

 - https://element-hq.github.io/matrix-authentication-service/topics/authorization.html#compatibility-sessions
 - https://element-hq.github.io/matrix-authentication-service/topics/authorization.html#oauth-20-sessions

---

```
# BaseData for the policies
`crates/policy/src/lib.rs`

# Legacy compatibility login
# `m.login.password` is non-interactive
# `m.login.sso` is interactive
policies/compat_login/compat_login.rego
policies/schema/compat_login_input.json
crates/handlers/src/compat/login.rs

# OAuth 2.0 authorization grant is an interactive login
policies/authorization_grant/authorization_grant.rego
```

```
count_user_sessions_for_limiting

end_compat_session
```

---

Previous attempts in Synapse:

 - https://github.com/matrix-org/synapse/pull/14595
 - https://github.com/matrix-org/synapse/pull/15183

---

To delete a device, it's as simple as `repo.compat_session().finish(clock, compat_session.to_owned()).await?;` which just updates the `finished_at` column in the database for that session.

Then a `SyncDevices` job communicates with Synapse to update its view of the world. This is where the real work happens

#### Testing

See [`docs/development/contributing.md#run-the-tests`](https://github.com/element-hq/matrix-authentication-service/blob/0f4ce9eccb9ed1d7df1a8bc8f737843b112e38ed/docs/development/contributing.md#run-the-tests)

```
DATABASE_URL='postgresql://postgres:postgres@localhost/mas' cargo nextest run --workspace

DATABASE_URL='postgresql://postgres:postgres@localhost/mas' cargo nextest run -p mas-handlers compat::login::tests

DATABASE_URL='postgresql://postgres:postgres@localhost/mas' cargo nextest run -p mas-handlers tests::test_get_login
```

```
cargo test --no-fail-fast --workspace

cargo test -p mas-handlers tests::test_get_login -- --nocapture
```

#### Prior work

 - https://github.com/element-hq/matrix-authentication-service/pull/5221
 - https://github.com/element-hq/matrix-authentication-service/pull/5287
 - https://github.com/element-hq/matrix-authentication-service/pull/5215
 
 
 #### Config error messages
 
See https://github.com/element-hq/matrix-authentication-service/pull/5607#discussion_r3034406794


crates/config/src/sections/experimental.rs
```diff
@@ -116,11 +120,79 @@ impl ExperimentalConfig {
...
+            error.profile = Some(figment::Profile::Default);
+            error.path = vec![Self::PATH.unwrap().to_owned(), field.to_owned()];
+            error
+        };
+
+        // See [`hard_limit_eviction`] docstring
+        if self.hard_limit_eviction && self.hard_limit.get() < 2 {
+            return Err(error_on_field(figment::error::Error::from(
+                "Session `hard_limit` must be at-least 2 when automatic `hard_limit_eviction` is set.",
+            ), "hard_limit").into());
```
MadLittleMods: The error output sucks. We don't see the config path at all.

```
$ cargo run -- config check --config config.yaml
Error: Session hard limit must be at-least 2 when automatic `hard_limit_eviction` is set.
```

`config.yaml`
```yaml
experimental:
  session_limit:
    soft_limit: 1
    hard_limit: 1
    hard_limit_eviction: true
```

Related: https://github.com/element-hq/matrix-authentication-service/issues/2491

https://docs.rs/figment/latest/figment/index.html#metadata mentions fancy looking errors like the following but I assume they're just saying you can add your own formatter and use the metadata 🤷 

```
error: invalid type: found string "hi", expected u16
 --> key `debug.port` in TOML file App.toml
```

crates/policy/src/model.rs
```diff
@@ -202,7 +202,7 @@ pub struct CompatLoginInput<'a> {
     pub user: &'a User,
 
     /// How many sessions the user has.
-    pub session_counts: SessionCounts,
+    pub session_counts: &'a SessionCounts,
```
MadLittleMods: Made this a reference so we don't have to clone it when we want to use it in two spots (see `user_password_login(...)`)

crates/handlers/src/compat/login.rs
```diff
@@ -726,24 +730,84 @@ async fn user_password_login(
...
+
+                    // Find the least recently used compat sessions
+                    //
+                    // In the future, it may be nice to avoid sessions with
+                    // cryptographic state (what does that mean exactly? keys uploaded
+                    // for device?).
+                    let compat = repo
+                        .compat_session()
+                        .list(
+                            // TODO: Order by `last_active_at`
```
MadLittleMods: It looks like `list(...)` always has a fixed lexicographical ordering based on the ID's.

Before I make a giant refactor, how should we do the ordering?

We could include the ordering `column: IntoColumnRef` as part of the [`Pagination`](https://github.com/element-hq/matrix-authentication-service/blob/0f4ce9eccb9ed1d7df1a8bc8f737843b112e38ed/crates/storage/src/pagination.rs#L17-L31) struct which probably makes sense but is a big refactor that I should split out to its own PR.

Overall, this kind of sorting would be good for the UI as well.

### Dev notes

`CompatSessionRepository` trait:

https://github.com/element-hq/matrix-authentication-service/blob/0f4ce9eccb9ed1d7df1a8bc8f737843b112e38ed/crates/storage/src/compat/session.rs#L290-L306

`impl CompatSessionRepository for PgCompatSessionRepository`:

https://github.com/element-hq/matrix-authentication-service/blob/0f4ce9eccb9ed1d7df1a8bc8f737843b112e38ed/crates/storage-pg/src/compat/session.rs#L530-L533

Ends up in the query as:

https://github.com/element-hq/matrix-authentication-service/blob/0f4ce9eccb9ed1d7df1a8bc8f737843b112e38ed/crates/storage-pg/src/pagination.rs#L26-L31

https://github.com/element-hq/matrix-authentication-service/blob/0f4ce9eccb9ed1d7df1a8bc8f737843b112e38ed/crates/storage-pg/src/pagination.rs#L48-L61

MadLittleMods: Looks like we have a good index for this: `"compat_sessions_last_active_at_idx" btree (last_active_at)`

```shell
$ psql --username=postgres mas
mas=# \d+ compat_sessions
                                                   Table "public.compat_sessions"
      Column       |           Type           | Collation | Nullable | Default | Storage  | Compression | Stats target | Description
-------------------+--------------------------+-----------+----------+---------+----------+-------------+--------------+-------------
 compat_session_id | uuid                     |           | not null |         | plain    |             |              |
 user_id           | uuid                     |           | not null |         | plain    |             |              |
 device_id         | text                     |           |          |         | extended |             |              |
 created_at        | timestamp with time zone |           | not null |         | plain    |             |              |
 finished_at       | timestamp with time zone |           |          |         | plain    |             |              |
 is_synapse_admin  | boolean                  |           | not null | false   | plain    |             |              |
 last_active_at    | timestamp with time zone |           |          |         | plain    |             |              |
 last_active_ip    | inet                     |           |          |         | main     |             |              |
 user_session_id   | uuid                     |           |          |         | plain    |             |              |
 user_agent        | text                     |           |          |         | extended |             |              |
 human_name        | text                     |           |          |         | extended |             |              |
Indexes:
    "compat_sessions_pkey" PRIMARY KEY, btree (compat_session_id)
    "compat_sessions_finished_at_idx" btree (finished_at) WHERE finished_at IS NOT NULL
    "compat_sessions_inactive_ips_idx" btree (last_active_at) WHERE last_active_ip IS NOT NULL AND last_active_at IS NOT NULL
    "compat_sessions_last_active_at_idx" btree (last_active_at)
    "compat_sessions_user_fk" btree (user_id) INCLUDE (last_active_at)
    "compat_sessions_user_session_fk" btree (user_session_id)
Foreign-key constraints:
    "compat_sessions_user_id_fkey" FOREIGN KEY (user_id) REFERENCES users(user_id)
    "compat_sessions_user_session_id_fkey" FOREIGN KEY (user_session_id) REFERENCES user_sessions(user_session_id)
Referenced by:
    TABLE "compat_access_tokens" CONSTRAINT "compat_access_tokens_compat_session_id_fkey" FOREIGN KEY (compat_session_id) REFERENCES compat_sessions(compat_session_id)
    TABLE "compat_refresh_tokens" CONSTRAINT "compat_refresh_tokens_compat_session_id_fkey" FOREIGN KEY (compat_session_id) REFERENCES compat_sessions(compat_session_id)
    TABLE "compat_sso_logins" CONSTRAINT "compat_sso_logins_compat_session_id_fkey" FOREIGN KEY (compat_session_id) REFERENCES compat_sessions(compat_session_id) ON DELETE SET NULL
Not-null constraints:
    "compat_sessions_compat_session_id_not_null" NOT NULL "compat_session_id"
    "compat_sessions_user_id_not_null" NOT NULL "user_id"
    "compat_sessions_created_at_not_null" NOT NULL "created_at"
    "compat_sessions_is_synapse_admin_not_null" NOT NULL "is_synapse_admin"
Access method: heap
```

crates/policy/src/model.rs
MadLittleMods: No longer relevant (removed change)

crates/cli/src/commands/server.rs
```diff
@@ -127,9 +127,7 @@ impl Options {
 
         // Load and compile the WASM policies (and fallback to the default embedded one)
         info!("Loading and compiling the policy module");
-        let policy_factory =
-            policy_factory_from_config(&config.policy, &config.matrix, &config.experimental)
-                .await?;
```
MadLittleMods, line 132: The `session_limit` refactor to move it from the `BaseData` to one of the inputs has been split off to https://github.com/element-hq/matrix-authentication-service/pull/5611

crates/handlers/src/compat/login.rs
MadLittleMods: cc @sandhose as a tastemaker

crates/cli/src/commands/debug.rs
```diff
@@ -9,8 +9,7 @@ use std::process::ExitCode;
 use clap::Parser;
```
MadLittleMods, line 9: Putting this up for review in order to get opinions on https://github.com/element-hq/matrix-authentication-service/pull/5607#discussion_r3041621109 and glaring problems (or preferred patterns).

crates/handlers/src/compat/login.rs
```diff
@@ -1512,4 +1591,240 @@ mod tests {
...
+            .await
+            .unwrap();
+        assert_eq!(
+            session_counts.total, 2,
+            "Must not have more sessions ({}) than allowed by the `hard_limit` ({}). \
+            Expected one of the old sessions to be dropped to make room for the new login",
+            session_counts.total, session_limit_config.hard_limit,
+        );
+
+        // Also ensure that the newest sessions remain (we dropped the oldest)
```
MadLittleMods, line 2115: This is expected to fail until https://github.com/element-hq/matrix-authentication-service/pull/5607#discussion_r3041621109 is resolved.

crates/data-model/src/site_config.rs
```diff
@@ -39,10 +39,12 @@ pub struct SessionExpirationConfig {
     pub compat_session_inactivity_ttl: Option<Duration>,
 }
 
+/// See [`mas_config::sections::ExperimentalSessionLimitConfig`]
```
MadLittleMods: :x: https://github.com/element-hq/matrix-authentication-service/actions/runs/24103719954/job/70321493421?pr=5607
```
error: unresolved link to `mas_config::sections::ExperimentalSessionLimitConfig`
  --> crates/data-model/src/site_config.rs:42:11
   |
42 | /// See [`mas_config::sections::ExperimentalSessionLimitConfig`]
   |           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ no item named `sections` in module `mas_config`
   |
   = note: requested on the command line with `-D rustdoc::broken-intra-doc-links`
```

This still fails even though I've added `mas-config` to `crates/data-model/Cargo.toml`.

For reference, the reference works fine to jump in VSCode 🤔 

If I import `use mas_config;` at the top, it complains about it being unused/redundant 🤷

crates/handlers/src/compat/login.rs
```diff
@@ -1512,4 +1591,242 @@ mod tests {
...
+        // Sanity check our comparison (ensure we're not comparing an empty set)
+        assert_eq!(
+            most_recent_login_device_ids.len(),
+            2,
+            "Expected 2 logins for the next comparison"
+        );
+
+        // The remaining sessions should be the most recent sessions
+        #[allow(clippy::uninlined_format_args)]
+        {
```
MadLittleMods: I don't really want to inline the format args. They are long and annoying.

In order to associate this clippy allow with this single usage, I've added a block here.

Strong preference? Perhaps, we ignore this clippy lint at a project level

crates/config/src/sections/experimental.rs
```diff
@@ -116,11 +120,92 @@ impl ExperimentalConfig {
...
+    /// and you want to avoid breaking their operation while maintaining some
+    /// level of sanity with the number of devices that people can have.
+    ///
+    /// [`hard_limit`]: Self::hard_limit
+    /// [`hard_limit_eviction`]: Self::hard_limit_eviction
+    #[serde(default = "default_false")]
+    pub hard_limit_eviction: bool,
+}
+
+impl ConfigurationSection for SessionLimitConfig {
```
sandhose: It's not really obvious but the purpose of `ConfigurationSection` is only for top-level configs, so that we can do partial loading of config sections easily. So the validation should happen in the `impl ConfigurationSection for ExperimentalConfig`

crates/config/src/sections/experimental.rs
sandhose: It might because the path of the error is effectively `session_limit` instead of `experimental.session_limit`, so figment cannot figure out the exact source of the error

crates/data-model/src/site_config.rs
sandhose: I think it's because `mas_config::sections` isn't pub, but there is a re-export of all its children. `rust-analyzer` will happily resolve private items, but rustdoc only documents public items, hence the 'broken link'


```suggestion
/// See [`mas_config::ExperimentalSessionLimitConfig`]
```

crates/handlers/src/compat/login.rs
```diff
@@ -490,12 +492,104 @@ pub(crate) async fn post(
...
+                // cryptographic state (what does that mean exactly? keys uploaded
+                // for device?).
+                //
+                // FIXME: We could potentially use
+                // `repo.compat_session().finish_bulk(...)` if it had the ability to
+                // limit and order.
+                let compat_session_page = repo
+                    .compat_session()
+                    .list(
+                        // TODO: Order by `last_active_at`
```
sandhose: Figuring out the right way to do custom ordering is *hard*. I had an attempt the `quenting/admin-api/order-by` branch but it's… complicated. This is definitely something we want, especially in the admin API, but query building for this with proper pagination cursors is complicated :)

crates/handlers/src/compat/login.rs
sandhose: That is a very good question, and as you already figured out, it is quite a big refactor to figure that one out.

I did an attempt a few months ago, in an effort to add custom order criterias to the admin API, to then use it in Element Admin. That attempt lives in the `quenting/admin-api/order-by` branch, but I don't exactly remember how far did I go with this.

The problem is that you end up with wanting a *generic* `Pagination` structure with an ordering criteria specific to the object type. You kinda want the API to look like


```rust
.list(
  filter,
  Pagination::first(10)
    .order_by(UserOrdering::Name)
    .after(UserCursor::Name { username, id }),
)
```

So you need an 'Ordering' trait with an associated cursor that can be parsed/serialized (if we then want that to come up in the API):

https://github.com/element-hq/matrix-authentication-service/blob/badaf6667ccb9cea168d35120886b140e5114ced/crates/storage/src/pagination.rs#L22-L39

You need the pagination to be generic over the ordering:

https://github.com/element-hq/matrix-authentication-service/blob/badaf6667ccb9cea168d35120886b140e5114ced/crates/storage/src/pagination.rs#L58-L71

You need the resulting `Page` to be generic over the cursor used:

https://github.com/element-hq/matrix-authentication-service/blob/badaf6667ccb9cea168d35120886b140e5114ced/crates/storage/src/pagination.rs#L241-L252

And each item on the page should actually have the cursor of the item in it (or the page should have a start/end cursor):

https://github.com/element-hq/matrix-authentication-service/blob/badaf6667ccb9cea168d35120886b140e5114ced/crates/storage/src/pagination.rs#L234-L239

Note that because we're not ordering by the primary key (a unique constraint) anymore, the cursors *must* include the value of your ordering criteria AND the properly unique primary key, and you must filter by this, e.g.


```sql
WHERE (last_activity, id) > ($1, $2)
ORDER BY last_activity ASC, id ASC
```

or something like that, which makes query building quite annoying.

Nothing that isn't achievable, but yes, lots of work to get right!

So, indeed, out of scope of this PR, so what I would cheekily do in the meantime is do the eviction in two steps, with first a pool of 'inactive' sessions, then considering all the sessions. So you would first look at sessions that have a last active at `timestamp > now - 90 days` (which you can do with a filter), ordered by creation time (which you get naturally out of the primary key), then if you haven't evicted enough sessions, then you start looking at all sessions, regardless of inactivity.

sandhose: Indeed, that index exists for two reason:

 - filtering inactive sessions in the frontend
 - automatically remove inactive session after some time (an experimental feature)

crates/handlers/src/compat/login.rs
```diff
@@ -490,12 +492,104 @@ pub(crate) async fn post(
...
+    user: &User,
+    violations: Vec<Violation>,
+) -> Result<(), RouteError> {
+    match (violations.len(), violations.first()) {
+        // If the only violation is having reached the session limit, we might be
+        // able to resolve the situation.
+        //
+        // We don't trigger this if there was some other violation anyway, since
+        // that means that removing a session wouldn't actually unblock the login.
+        (1, Some(violation)) if violation.variant == Some(ViolationVariant::TooManySessions) => {
```
sandhose: I think you can do slice patterns, e.g.

```rust
match &violations[..] {
  [Violation { variant: Some(ViolationVariant::TooManySessions), .. }] => { }
  [] => { }
  violations => { }
}
```

crates/handlers/src/compat/login.rs
```diff
@@ -490,12 +492,104 @@ pub(crate) async fn post(
...
+                    .expect("We should have a `session_limit` config if we are seeing a `TooManySessions` violation. \
+                    This is most likely a programming error.");
+
+            // TODO: This should come from `ViolationVariant::TooManySessions`
+            let need_to_remove: u32 = 1;
+            let need_to_remove_usize = usize::try_from(need_to_remove).map_err(|err| {
+                RouteError::Internal(
+                    anyhow::anyhow!("Unable to convert `need_to_remove` to usize: {err}").into(),
+                )
+            })?;
```
sandhose: This effectively never happen, because the reason `usize` only implements `TryFrom<u32>` and not `From<u32>` (but it implements `From<u16>`) is that technically Rust supports architectures where the pointer size is 16 bits?

I honestly usually do `usize::try_from(…).unwrap_or(usize::MAX)` as it's a good enough approximation ('truncating' to the max size of usize), just in case.

crates/handlers/src/compat/login.rs
```diff
@@ -490,12 +492,104 @@ pub(crate) async fn post(
...
+                    }
+                    num_sessions_removed
+                };
+
+                // For now, we only automatically clean up compatibility sessions.
+                // If there are still too many sessions, we just throw an error with
+                // an explanation.
+                if num_sessions_removed < need_to_remove {
+                    return Err(RouteError::PolicyHardSessionLimitReached);
+                }
```
sandhose: This all feels very defensive :D 

`Pagination::first(need_to_remove_usize)` creates an `compat_session_page.edges.len() <= need_to_remove_usize` invariant already.

Then we check if `compat_session_page.edges.len() < need_to_remove_usize`, which means that effectively `compat_session_page.edges.len() == need_to_remove_usize`

Then we iterate over edges, counting the number of iterations, which is effectively equivalent to `num_sessions_removed == compat_session_page.edges.len()`, which means by this point we *always* have `num_sessions_removed == need_to_remove`

crates/handlers/src/compat/login.rs
```diff
@@ -490,12 +492,104 @@ pub(crate) async fn post(
...
+        // We don't trigger this if there was some other violation anyway, since
+        // that means that removing a session wouldn't actually unblock the login.
+        (1, Some(violation)) if violation.variant == Some(ViolationVariant::TooManySessions) => {
+            let session_limit_config = session_limit_config
+                    .expect("We should have a `session_limit` config if we are seeing a `TooManySessions` violation. \
+                    This is most likely a programming error.");
+
+            // TODO: This should come from `ViolationVariant::TooManySessions`
+            let need_to_remove: u32 = 1;
+            let need_to_remove_usize = usize::try_from(need_to_remove).map_err(|err| {
```
sandhose: style preference, but [variable shadowing](https://doc.rust-lang.org/rust-by-example/variable_bindings/scope.html) is a really cool feature of rust, which means we don't *need* to do things like that :)

```suggestion
            let need_to_remove: u32 = 1;
            let need_to_remove = usize::try_from(need_to_remove).map_err(|err| {
```

crates/handlers/src/compat/login.rs
```diff
@@ -490,12 +492,104 @@ pub(crate) async fn post(
...
+            } else {
+                // Tell the user about the limit
+                return Err(RouteError::PolicyHardSessionLimitReached);
+            }
+        }
+        // Nothing is wrong
+        (0, None) => return Ok(()),
+        // Just throw an error for any other violation
+        _ => {
+            return Err(RouteError::PolicyRejected);
```
sandhose, line 620: surprised we don't expose the violation details there

crates/data-model/src/site_config.rs
MadLittleMods: Good shout, thank you! That seems to be exactly it and happy to have some explanation for why :tada:

crates/config/src/sections/experimental.rs
MadLittleMods: We already have `impl ConfigurationSection for ExperimentalConfig` that calls this one. It feels better with the validation closer to `SessionLimitConfig` :thinking: 

The current pattern here is copied from 

https://github.com/element-hq/matrix-authentication-service/blob/cc0903eb789a52f76c7a8f124307bf4de2f2e6ff/crates/config/src/sections/mod.rs#L134-L158

https://github.com/element-hq/matrix-authentication-service/blob/cc0903eb789a52f76c7a8f124307bf4de2f2e6ff/crates/config/src/sections/mod.rs#L295-L307

But I guess that also aligns with "only for top-level configs". I can update to do that, it just feels a bit icky to separate the config validation so far apart.

My guess is that we don't want to do this because it just wasn't designed for this :shrug: and we should think about a more automatically composable solution in the future.

(just poke again)

crates/config/src/sections/experimental.rs
MadLittleMods: Even hard-coding `error.path` to `["experimental", "session_limit", "hard_limit"]` doesn't help :thinking: 

Does it work for other errors? (do you expect it to work?)

crates/handlers/src/compat/login.rs
MadLittleMods: Resolving in favor of the other ordering thread -> https://github.com/element-hq/matrix-authentication-service/pull/5607#discussion_r3041621109

crates/handlers/src/compat/login.rs
MadLittleMods: Ouuu, I haven't seen this before :+1:

crates/handlers/src/compat/login.rs
MadLittleMods: I'd rather keep the explicit error case handling (better context when things go wrong). And in this case, we'd remove too few sessions if we ran into an overflow and used `usize::MAX`.

The [other place](https://github.com/element-hq/matrix-authentication-service/blob/db8ee567c003d48ed7599e997b782a41893c0d26/crates/storage-pg/src/queue/job.rs#L446) I see `usize::try_from(...).unwrap_or(usize::MAX)` won't cause any logic problems. I see other examples for other types that I haven't analyzed.

The assumption of a minimum 32-bit system is probably a fine reality but taints the code.

crates/handlers/src/compat/login.rs
MadLittleMods: hmm, this does appear weird. This is a holdover from the previous code introduced in https://github.com/element-hq/matrix-authentication-service/pull/5287 which also introduced `policies/compat_login/compat_login.rego` in the first place.

I think for the sake of this PR, I'm going to opt to push this problem forward (should be handled in a future PR).

I've added a `FIXME` comment to call out the weirdness :fast_forward:

crates/handlers/src/compat/login.rs
```diff
@@ -1512,4 +1653,403 @@ mod tests {
...
+        for edge in compat_session_page.edges {
+            let (compat_session, _) = edge.node;
+            let last_active_at = compat_session
+                .last_active_at
+                .expect("We expect compat sessions to have `last_active_at` set for this test");
+            assert!(
+                last_active_at < (state.clock.now().sub(Duration::days(90))),
+                "Expected compat sessions to have a `last_active_at` older than the 90 day 'inactive' threshold"
+            );
+        }
```
MadLittleMods: How do I wait for `last_active_at` to be set on the compat sessions after logging in?

It seems like it's asynchronously in a background job as part of [`record_batch_activity`](https://github.com/element-hq/matrix-authentication-service/blob/db8ee567c003d48ed7599e997b782a41893c0d26/crates/storage-pg/src/compat/session.rs#L579) 🤔

crates/policy/src/lib.rs
```diff
@@ -98,21 +98,18 @@ pub struct Data {
...
+    pub server_name: String,
 
     /// Limits on the number of application sessions that each user can have
-    session_limit: Option<SessionLimitConfig>,
+    pub session_limit: Option<SessionLimitConfig>,
 }
 
 impl Data {
     #[must_use]
-    pub fn new(server_name: String, session_limit: Option<SessionLimitConfig>) -> Self {
```
MadLittleMods, line 111: Because I had to update the signature of the `policy_factory(...)` function to pass `session_limit`, I decided to make the usage a little more straight-forward by just using `BaseData` which also gets us nice labels from the struct instead of adding yet another argument.

crates/handlers/src/compat/login.rs
MadLittleMods: Things have been refactored ⏩

crates/handlers/src/compat/login.rs
MadLittleMods: Updated to use the suggestion explained at the end. Great way forward!

I also made things pull out at-least 100 sessions so we can sort it ourselves outside of the database. 100 should be no big deal. I didn't want to pull out everything as people could have thousands of sessions.

crates/handlers/src/compat/login.rs
MadLittleMods: Updated :+1: 

Depends on the scenario

Cargo.toml
```diff
@@ -24,7 +24,9 @@ unsafe_code = "deny"
 # We use groups as good defaults, but with a lower priority so that we can override them
 all = { level = "deny", priority = -1 }
 pedantic = { level = "warn", priority = -1 }
-
+# Allowed because it's nice to have temporary semantic names, see
+# https://github.com/rust-lang/rust-clippy/issues/12512#issuecomment-3316736180
+let_and_return = { level = "allow", priority = 1 }
```
MadLittleMods: Added because I want to do stuff like the following where the `sorted_compat_sessions` label is useful.

```rust
let sorted_compat_sessions = {
    ...
};

sorted_compat_sessions
```


Previous times we've introduced:

 - https://github.com/element-hq/synapse-rust-apps/pull/459#discussion_r2801860417
 - https://github.com/element-hq/synapse-rust-apps/pull/397#discussion_r2710424049

crates/config/src/sections/experimental.rs
sandhose:
```
> mas-cli config check
Error: client_secret is required for client_secret_basic for key "default.clients.0.client_auth_method" in config.yaml YAML file
```

with

```yaml
clients:
  - client_id: "01JZAEP46HN8H4B4GMH4NCDMPD"
    client_auth_method: client_secret_basic
```

so unsure why it's not working there 🤔

crates/handlers/src/compat/login.rs
sandhose: I don't mind. You can also trick it by naming the args explicitly e.g.

```rust
assert!(
    "{remaining:?} {recent:?}",
    remaining = remaining_active_compat_session_device_ids,
    recent = most_recent_login_device_ids,
)
```

crates/config/src/sections/experimental.rs
sandhose: > and we should think about a more automatically composable solution in the future.

Agree with that. `figment` was good for some things, but it's limited. I'm interested from the work from [amos](https://fasterthanli.me) on [facet](https://github.com/facet-rs/facet) and [figue](https://github.com/bearcove/figue). He's got some [pretty impressive error reporting](https://github.com/bearcove/figue/blob/main/crates/figue/tests/integration/snapshots/main__integration__layered__layered_dump_shows_all_sources.snap) with source tracking going on there!

crates/handlers/src/compat/login.rs
sandhose: I wish we could tell rust/clippy 'we're only targetting 32-bit+ systems' and have an easy u32 -> usize conversion. This is only necessary because we have [`clippy::cast_possible_truncation`](https://rust-lang.github.io/rust-clippy/master/index.html#cast_possible_truncation) enabled as part of the `pedantic` ruleset. Maybe we could think about dropping that one

crates/handlers/src/compat/login.rs
sandhose: `TestState.activity_tracker.flush()` should do the trick?

crates/policy/src/lib.rs
sandhose: Makes sense 👍

Cargo.toml
sandhose: sure, no strong feeling on this 👍

sandhose (commented):
(don't forget to put @element-hq/mas-maintainers back as reviewer after this round since I'll be off for a bit)

crates/handlers/src/compat/login.rs
MadLittleMods: Looks like a reasonable pattern for this case. Updated :+1:

MadLittleMods: hmmm, `clippy` still complains and this problem is tracked by https://github.com/rust-lang/rust-clippy/issues/10247

```
warning: variables can be used directly in the `format!` string
    --> crates/handlers/src/compat/login.rs:1921:9
     |
1921 | / ...   assert!(
1922 | | ...       most_recent_login_device_ids.is_subset(&remaining_active_compat_session_device_ids),
1923 | | ...       "Expected the 2 remaining active sessions ({remaining:?}) to include the 2 most recent logins ({recent:?}). (all logins: {login...
1924 | | ...       remaining = remaining_active_compat_session_device_ids,
1925 | | ...       recent = most_recent_login_device_ids,
1926 | | ...   );
     | |_______^
     |
     = help: for further information visit https://rust-lang.github.io/rust-clippy/rust-1.91.0/index.html#uninlined_format_args
     = note: `-W clippy::uninlined-format-args` implied by `-W clippy::pedantic`
     = help: to override `-W clippy::pedantic` add `#[allow(clippy::uninlined_format_args)]`
```

I'm tempted to just add this to our allowed list :fast_forward:

MadLittleMods: Added `uninlined_format_args` to the allowed Clippy lints :fast_forward:

crates/config/src/sections/experimental.rs
MadLittleMods: Ahhh, thanks for pointing out that example! I got this working now after copying the pattern from `ClientConfig`. I think the missing piece was that I wasn't using the correct path when finding the `error.metadata`.

```
Error: Session `hard_limit` must be at-least 2 when automatic `hard_limit_eviction` is set. See configuration docs for more info. for key "default.experimental.session_limit.hard_limit" in config.yaml YAML file
```

Fixed in https://github.com/element-hq/matrix-authentication-service/pull/5607/changes/e4b8a3659f831fe74e1a6e6039d15b61b1d0d795

crates/config/src/sections/experimental.rs
MadLittleMods: To resolve the original point here:

From the other [discussion](https://github.com/element-hq/matrix-authentication-service/pull/5607#discussion_r3034406794), it prompted me looking at our existing `ClientConfig` vs `ClientsConfig` and how those interact which gives a great pattern.

https://github.com/element-hq/matrix-authentication-service/blob/21566b769dc721bdb1932bfc6d8d6cfe19b4a799/crates/config/src/sections/clients.rs#L236-L255

https://github.com/element-hq/matrix-authentication-service/blob/21566b769dc721bdb1932bfc6d8d6cfe19b4a799/crates/config/src/sections/clients.rs#L96-L106

I've settled on doing the same thing with `impl ConfigurationSection for ExperimentalConfig` (only for the top-level) but still having `SessionLimitConfig::validate` :white_check_mark: 

See https://github.com/element-hq/matrix-authentication-service/pull/5607/changes/e4b8a3659f831fe74e1a6e6039d15b61b1d0d795

crates/handlers/src/compat/login.rs
MadLittleMods: Perfect :ok_hand: 

Thank you!

crates/handlers/src/compat/login.rs
MadLittleMods: ehh, that lint looks useful

Cargo.toml
MadLittleMods, line 5: `cargo-deny` CI failures are unrelated to this PR and tracked by https://github.com/element-hq/matrix-authentication-service/issues/5621

crates/handlers/src/compat/login.rs
```diff
@@ -490,12 +495,172 @@ pub(crate) async fn post(
...
+                    // ago). And by the nature of
+                    // [`mas_data_model::compat::CompatSession::id`] being a `Ulid`/`Uuid`
+                    // (the query is ordered by `compat_session_id`), the first bytes are a
+                    // timestamp so we'll be getting the 'oldest created' sessions which is
+                    // another good proxy.
+
+                    let mut edges_to_consider = Vec::new();
+
+                    // First, find the "inactive" sessions
+                    let inactive_threshold_date = clock.now() - Duration::days(90);
```
reivilibre: This 90d threshold would benefit from a rationale; is it because MAS' own webUI has some 'inactive' wording around this threshold iirc?

A cross-reference or (if appropriate) centralised constant would be good

crates/config/src/sections/experimental.rs
```diff
@@ -116,11 +120,86 @@ impl ExperimentalConfig {
...
+    #[serde(default = "default_false")]
+    pub hard_limit_eviction: bool,
+}
+
+impl SessionLimitConfig {
+    fn validate(&self) -> Result<(), Box<figment::error::Error>> {
+        // See [`SessionLimitConfig::hard_limit_eviction`] docstring
+        if self.hard_limit_eviction && self.hard_limit.get() < 2 {
+            return Err(figment::error::Error::from(
+                "Session `hard_limit` must be at-least 2 when automatic `hard_limit_eviction` is set. \
```
reivilibre:
```suggestion
                "Session `hard_limit` must be at least 2 when automatic `hard_limit_eviction` is set. \
```

no hyphen in this position, I believe (would make sense as a compound adjective)

crates/config/src/sections/experimental.rs
```diff
@@ -116,11 +120,86 @@ impl ExperimentalConfig {
...
+    ///
+    /// Disabled by default
+    ///
+    /// WARNING: Removing sessions is a potentially damaging operation. Any
+    /// end-to-end encrypted history on the device will be lost and can only
+    /// be recovered if you have another verified active device or have a
+    /// recovery key setup.
+    ///
+    /// When using [`hard_limit_eviction`], the [`hard_limit`] must be
+    /// at-least 2 to avoid catastropically losing encrypted history and digital
```
reivilibre:
```suggestion
    /// at-least 2 to avoid catastrophically losing encrypted history and digital
```

...but is also leaving the question open of 'why?' to me

<br />

Maybe a stupid question given if we log out the only other device and don't have any dehydrated devices then *of course* we lose the encrypted history, save for having an online key backup.

crates/policy/src/model.rs
```diff
@@ -70,7 +73,7 @@ impl ViolationVariant {
             Self::EmailDomainBanned => "email-domain-banned",
             Self::EmailNotAllowed => "email-not-allowed",
             Self::EmailBanned => "email-banned",
-            Self::TooManySessions => "too-many-sessions",
+            Self::TooManySessions { need_to_remove: _ } => "too-many-sessions",
```
reivilibre:
```suggestion
            Self::TooManySessions { .. } => "too-many-sessions",
```

this would also be fair and very defensible, but no preference towards either

crates/policy/src/model.rs
```diff
@@ -52,7 +52,10 @@ pub enum ViolationVariant {
     EmailBanned,
 
     /// The user has reached their session limit.
-    TooManySessions,
+    TooManySessions {
+        /// How many devices to remove
```
reivilibre:
```suggestion
        /// How many devices need to be removed to make room for the new session
```

<br />

policies/authorization_grant/authorization_grant.rego
```diff
@@ -157,6 +157,9 @@ violation contains {"msg": sprintf(
 violation contains {
 	"code": "too-many-sessions",
 	"msg": "user has too many active sessions",
+	# `+ 1` because when you're at 2 sessions, and the limit is 2, you have to make room
+	# for the new session
+	"need_to_remove": (input.session_counts.total - data.session_limit.hard_limit) + 1,
```
reivilibre: Are we missing a policy (rego) test for this? I think it would be reasonable to pop in a couple of assertions in **`policies/compat_login/compat_login_test.rego`** and `policies/authorization_grant/authorization_grant_test.rego`

crates/handlers/src/compat/login.rs
```diff
@@ -490,12 +495,172 @@ pub(crate) async fn post(
...
+                // For now, we only automatically clean up compatibility sessions.
+                // If there aren't enough sessions that we could clean up, we just
+                // throw an error with an explanation.
+                if lru_compat_sessions.len() < need_to_remove {
+                    return Err(RouteError::PolicyHardSessionLimitReached);
+                }
+
+                // Remove the sessions (only as much as necessary, `need_to_remove`)
+                for compat_session in &lru_compat_sessions[0..need_to_remove] {
+                    repo.compat_session()
```
reivilibre: it would be good to log the ID of the sessions we evict. (I mean, it'd even be kinda good to mark that info in the database for the paper trail, but I suggest that would be a better later step)

crates/handlers/src/compat/login.rs
```diff
@@ -490,12 +495,172 @@ pub(crate) async fn post(
...
+                    };
+
+                    sorted_compat_sessions
+                };
+
+                // For now, we only automatically clean up compatibility sessions.
+                // If there aren't enough sessions that we could clean up, we just
+                // throw an error with an explanation.
+                if lru_compat_sessions.len() < need_to_remove {
+                    return Err(RouteError::PolicyHardSessionLimitReached);
```
reivilibre: a log line for this is also probably a good idea, in case we get some support query about it

crates/handlers/src/compat/login.rs
```diff
@@ -490,12 +495,172 @@ pub(crate) async fn post(
...
+        // We don't trigger this if there was some other violation anyway, since
+        // that means that removing a session wouldn't actually unblock the login.
+        [
+            Violation {
+                variant: Some(ViolationVariant::TooManySessions { need_to_remove }),
+                ..
+            },
+        ] => {
+            let session_limit_config = session_limit_config
+                    .expect("We should have a `session_limit` config if we are seeing a `TooManySessions` violation. \
```
reivilibre: I am wondering if it would be worth dropping a couple of warning comments around the places someone would look, if they were going to write their own device limit policy in a way where they might forget to add the `session_limit` config...

For example, maybe just some comments in the rego files that this NEEDS `session_limit` to be configured, even if the soft/hard limits from that are ignored by the policy. (People will probably look there to crib off it, in the rare event that they write their own policy)

<br />

No real problem, just paranoia (and I can't necessarily see why people would be writing such exotic policies anyway)

crates/handlers/src/compat/login.rs
```diff
@@ -490,12 +495,172 @@ pub(crate) async fn post(
...
+            // When logging in with the compatibility API, there is no way for us to
+            // display any web UI for people to remove devices, so we instead
+            // automatically remove their oldest devices (when `hard_limit_eviction`
+            // is configured).
+            if session_limit_config.hard_limit_eviction {
+                // Find the least recently used (LRU) compat sessions
+                //
+                // In the future, it may be nice to avoid sessions with
+                // cryptographic state (what does that mean exactly? keys uploaded
+                // for device?).
```
reivilibre: I'd say so.

crates/handlers/src/compat/login.rs
```diff
@@ -490,12 +495,172 @@ pub(crate) async fn post(
...
+                        .list(
+                            CompatSessionFilter::new()
+                                .for_user(user)
+                                .active_only()
+                                .with_last_active_before(inactive_threshold_date),
+                            // We fetch a minimum of 100 sessions (more than we need in
+                            // normal cases) so we can sort by `last_active_at` after it
+                            // gets back from the database and can get even closer to
+                            // removing the oldest sessions.
+                            Pagination::first(std::cmp::max(need_to_remove, 100)),
```
reivilibre: This 100 limit makes me a little uncomfortable, because it wouldn't surprise me for practical device limits to be around the 100 or 200 mark on permissive setups like matrix.org.

If this seems like a false premise, maybe we should explain that it's OK because you wouldn't use near 100 in practice (& do we document on the config option that using eviction with such large limits is more or less buggy right now?)

Otherwise, I'd be tempted to consider how large each of these objects actually is and maybe just pull out a 'few' more into memory, though I guess it never hurts to document our assumptions.

crates/handlers/src/compat/login.rs
```diff
@@ -1512,4 +1660,410 @@ mod tests {
...
+                    panic!("Expected `device_id` to be a string")
+                }
+            };
+            login_device_ids.push(device_id);
+
+            // Wait for `last_active_at` to be set
+            state.activity_tracker.flush().await;
+
+            // Restore time
+            state.clock.advance(original_time - state.clock.now());
```
reivilibre: why rewind time instead of advancing 90d after the test starts?

crates/handlers/src/compat/login.rs
```diff
@@ -1512,4 +1660,410 @@ mod tests {
...
+            .expect("Should be able to list user's compat sessions");
+        let remaining_active_compat_session_device_ids: HashSet<String> = compat_session_page
+            .edges
+            .iter()
+            .map(|a| {
+                a.node
+                    .0
+                    .device
+                    .clone()
+                    .expect("Expected each login should havea a device")
```
reivilibre:
```suggestion
                    .expect("Expected each login should have a device")
```

crates/handlers/src/compat/login.rs
```diff
@@ -490,12 +495,172 @@ pub(crate) async fn post(
...
+                // `repo.compat_session().finish_bulk(...)` if it had the ability to
+                // limit and order.
+                let lru_compat_sessions = {
+                    // TODO: In the future, instead of all of this faff, we can simply order
+                    // by `last_active_at`
+                    //
+                    // XXX: Since we can't order by `last_active_at` yet, we instead filter
+                    // the list down to "inactive" sessions (`last_active_at` > 90 days
+                    // ago). And by the nature of
+                    // [`mas_data_model::compat::CompatSession::id`] being a `Ulid`/`Uuid`
```
reivilibre:
```suggestion
                    // [`mas_data_model::compat::CompatSession::id`] being a `Ulid`
```

Just to avoid confusing someone; Uuids (most if not all of them) don't have this property, that's why Ulid is considered nifty

crates/handlers/src/compat/login.rs
```diff
@@ -490,12 +495,172 @@ pub(crate) async fn post(
...
+                    // (the query is ordered by `compat_session_id`), the first bytes are a
+                    // timestamp so we'll be getting the 'oldest created' sessions which is
+                    // another good proxy.
```
reivilibre: To an extent.

I can easily see the following situation:

* someone has a long-lived Element Web account

* that someone sets up a hacky script that logs in every time

* now they have 200 devices; the oldest one is their genuine EW account and the remaining 199 are the hacky one-off script logins

* They come to log in and need 100 devices to be removed

By limiting the window to 100, I think this current implementation will remove all the oldest 100 devices by creation time, including the desirable Element Web session.

I don't mind whether or not we tackle this now, but if not we should definitely document it as an issue and as a warning on the config option (IMO) as it feels really quite destructive

reivilibre (changes requested):
I'm happy with this overall

crates/handlers/src/compat/login.rs
MadLittleMods: Updated :fast_forward: 

What's confusing is that it seems to be mixed usage of `Ulid` and `Uuid` for this field. Perhaps only the database interactions use `Uuid` :shrug: 

https://github.com/element-hq/matrix-authentication-service/blob/64f90e01da2de8e5a69e41f9031ab9ccf7457b85/crates/data-model/src/compat/session.rs#L70-L71

https://github.com/element-hq/matrix-authentication-service/blob/64f90e01da2de8e5a69e41f9031ab9ccf7457b85/crates/storage-pg/src/app_session.rs#L67

https://github.com/element-hq/matrix-authentication-service/blob/64f90e01da2de8e5a69e41f9031ab9ccf7457b85/crates/storage-pg/src/compat/access_token.rs#L37

https://github.com/element-hq/matrix-authentication-service/blob/64f90e01da2de8e5a69e41f9031ab9ccf7457b85/crates/data-model/src/compat/sso_login.rs#L26

crates/handlers/src/compat/login.rs
MadLittleMods: Moved the `XXX` comment closer here and added a reference to [`getNinetyDaysAgo()`](https://github.com/element-hq/matrix-authentication-service/blob/64f90e01da2de8e5a69e41f9031ab9ccf7457b85/frontend/src/utils/dates.ts#L8C14-L13) that the web UI uses :+1:

crates/config/src/sections/experimental.rs
MadLittleMods: Updated the typo :+1: 

Are you asking why require `hard_limit: 2`? It's a barebones safety net for people as if someone configures `1`, anyone who doesn't have a recovery key setup will lose their encrypted history and digital identity when they login again.

It's impossible to assert any number of devices is safe but this prevents people from the most obvious pitfall.

I think the details are explained in the docstring (let me know).

The idea of evicting devices automatically sounds really bad to me but that's the use case we want for `matrix.org`:

https://github.com/element-hq/matrix-authentication-service/blob/bc1e71baa3d7b417dc80c169ef4a65ae3fa17961/crates/config/src/sections/experimental.rs#L180-L185

Perhaps we should call this option `dangerous_hard_limit_eviction`

crates/handlers/src/compat/login.rs
MadLittleMods: I don't think it's needed. The error explains what to do: `You have reached your hard device limit. Please visit your account page to sign some out.`

crates/handlers/src/compat/login.rs
MadLittleMods: Added a log and FIXME for recording this in the database :+1:

crates/handlers/src/compat/login.rs
MadLittleMods: :+1: Thanks for confirming the hunch. I think this comment is still good to go :fast_forward:

crates/handlers/src/compat/login.rs
MadLittleMods: > This 100 limit makes me a little uncomfortable, because it wouldn't surprise me for practical device limits to be around the 100 or 200 mark on permissive setups like matrix.org.

What are your concerns?

a) We're pulling out too many which is resource/time intensive?

b) We should be trying to pull out all devices so we can get an absolutely accurate `last_active_at` sort? I'm willing to bump this number higher but I didn't want to pull out 40k devices in the pathological cases. I didn't measure how much would be reasonable and I don't know how big our largest offenders on `matrix.org` actually are.


> using eviction with such large limits is more or less buggy right now?

I think you may have read the code wrong (it's not a limit). We will pull out as many devices as we `need_to_remove` or `100` (whichever is higher, the max).

(it's not buggy in that aspect)

crates/handlers/src/compat/login.rs
MadLittleMods: Either way could work. I think I was going for whatever time the test starts at is the "current time" and we want to emulate devices being created in the past.

There could potentially be other time-tracked things in the database where "current time" could make sense :shrug:

crates/handlers/src/compat/login.rs
MadLittleMods: Related to the other conversation: https://github.com/element-hq/matrix-authentication-service/pull/5607#discussion_r3119473846

As long as the Element Web session is actively used, it won't be removed with the current code.

To explain:

 1. If we have a `hard_limit: 200`
 1. If they have 200 devices, we will first query for their devices which are "inactive" (90 days old) up-to the number of sessions that we `need_to_remove` (at least 100) (sorted by creation date, ascending).
 1. If we need to find more devices, we look at all sessions up-to the number of sessions that we `need_to_remove` (at least 100) (sorted by creation date, ascending).
 1. We combine the results from those queries and sort the sessions by `last_active_at`
 1. Remove the least-recently used sessions in the list

The Element Web session won't be one of the bottom sessions that we `need_to_remove`.

If we're talking about a script that logs in every day, the Element Web session would have to go unused for 200 days before being considered for removal.

In the case that all 200 sessions are "active" (newer than 90 days), the Element Web session would have to be the oldest created and used less recently than 99 other sessions.

policies/authorization_grant/authorization_grant.rego
MadLittleMods: Added policy tests to assert `need_to_remove` for all of the cases we have :white_check_mark: 

The tests now look like this:

```rego
# Helper utility to extract the number of sessions that they `need_to_remove`, returns 0
# if the `too-many-sessions` violation is not found
get_need_to_remove(violations) := need if {
	some v in violations
	v.code == "too-many-sessions"
	need := v.need_to_remove
} else := 0

test_session_limiting_sso_hit_soft_limit if {
	result := {
		"allow": compat_login.allow,
		"need_to_remove": get_need_to_remove(compat_login.violation),
	} with input.user as user
		with input.session_counts as {"total": 32}
		with input.login as {"type": "m.login.sso"}
		with input.session_replaced as false
		with data.session_limit as {"soft_limit": 32, "hard_limit": 64}
	not result.allow
	result.need_to_remove == 1
}
```

They have to be split up because we can't redeclare `result` (`rego_compile_error: var result assigned above`). As far as I can tell, splitting up the tests also aligns with the [test example in the docs](https://www.openpolicyagent.org/docs/policy-testing) :shrug:. Wish they had more examples to work from.

policies/authorization_grant/authorization_grant_test.rego
```diff
@@ -223,34 +223,81 @@ test_mas_scopes if {
...
+		"need_to_remove": get_need_to_remove(authorization_grant.violation),
+	} with input.user as user
 		with input.session_counts as {"total": 1}
+		# No limit configured
 		with data.session_limit as null
-
-	# Client credentials grant
-	authorization_grant.allow with input.user as user
-		with input.session_counts as null
-		with data.session_limit as {"soft_limit": 32, "hard_limit": 64}
```
MadLittleMods, line 255: Removed this test as it doesn't actually test the client credentials grant as stated. It would need to define `input.grant_type` as `"client_credentials"`. Even so, seems useless.

If we want to make sure we run into `too-many-sessions` violations for all of the OAuth 2.0 grants, we should do it at the integration level in Rust.

This was wrong since it was introduced in https://github.com/element-hq/matrix-authentication-service/pull/5221

policies/compat_login/compat_login_test.rego
```diff
@@ -10,90 +10,190 @@ import rego.v1
...
+# If the session is replacing an existing session, no need to throw any violations about
+# too many sessions
+test_no_session_limiting_sso_upon_replacement if {
+	result := {
+		"allow": compat_login.allow,
+		"need_to_remove": get_need_to_remove(compat_login.violation),
+	} with input.user as user
 		with input.session_counts as {"total": 65}
-		with input.login as {"type": "m.login.password"}
-		with input.session_replaced as false
```
MadLittleMods, line 91: Fixed these two session replacement tests to actually use `input.session_replaced as true` which should be allowed.

These were wrong since they were introduced in https://github.com/element-hq/matrix-authentication-service/pull/5287

crates/handlers/src/compat/login.rs
MadLittleMods: Just to be clear, you're worried about someone creating their own policy which emits `too-many-session` violations but their logic isn't based on `data.session_limit` at all (and doesn't define any `session_limit` config)?

If that's the case, we can drop this assertion (too-tight assertion) and handle the case of `session_limit_config` not being defined. I thought it was a good detection of a programming error if we can keep this assumption though.

I'd rather drop the assertion than add comments to the policy about `session_limit` config needing to be filled in.

(just poke again if I've understood your paranoia properly)

crates/handlers/src/compat/login.rs
reivilibre: Yes, I can see how it's confusing, but that's right, Uuids are used at the database layer because Postgres has a 128-bit UUID type that we can abuse for storing ULIDs (merely because they have the same size). sqlx then supports conversion between `Uuid` and Postgres `UUID`

crates/handlers/src/compat/login.rs
reivilibre: As far as I know everything should use the clock (or be considered a bug), so I don't think we have to worry about the concept of 'current time' (if it's like Synapse, it might start at time = 0 anyway)
Though having a clock that goes backwards (i.e. is not semi-monotonic) feels poor.
Ultimately it doesn't seem to matter for this test though; but I guess I like 'reading code line-by-line, the time only goes forward' as a general principle though.

Not fussed if you find this style clearer though

crates/handlers/src/compat/login.rs
reivilibre: My concern is (b) and my 'more or less buggy' point refers to the fact that we don't get the genuinely least-recently used devices when the user has beyond a certain amount of devices.

Given the low chance of exploitation (afaict!), I think I'd rather pull out a large number since it's a one-off.

I say 'low chance of exploitation'/'one-off' because once the device limit is in place, you can't get back to a situation where you can trigger this again.

I skimmed the struct definitions of the rows we're pulling out. `CompatSession`'s size is 192 bytes, plus it contains a couple of strings, let's call it 512 bytes total to be pessimistic.
`CompatSsoLogin`, also in each row, is also 192 bytes, with an extra `login_token` string (which is probably no more than 64 bytes but I haven't checked).
Let's call each row altogether 1 kiB as a pessimistic high bound.

Then: how much memory are we willing to temporarily spend to make this 'correct'? 1 MiB? Some operations in Synapse use way more than that, if anything 1 MiB is harsh on ourselves. But in that, we can fit 1024 rows (pessimistically). Maybe we should go further.

Out of curiosity, here is how many devices people on m.org have...

user vs devices plot

crates/handlers/src/compat/login.rs
MadLittleMods: The `MockClock` for tests starts on 2022-01-16

https://github.com/element-hq/matrix-authentication-service/blob/1d428b06243c919709b3454a815742327a9e8d00/crates/data-model/src/clock.rs#L55-L60

I was thinking of "current time" as the time on the clock when the test starts. I was thinking of it more in terms of making sure that when we create app/state, we honor the "current time".

But given we're using `/login` to create the past sessions which messes with that same app/state, it's not that much of an ideal to hold onto.

> I guess I like 'reading code line-by-line, the time only goes forward' as a general principle though.

Sounds good to me as well. Updated :white_check_mark:

crates/config/src/sections/experimental.rs
MadLittleMods: I've updated the name to `dangerous_hard_limit_eviction` to better indicate the potential danger. Especially since we don't discriminate between sessions that have cryptographic state (keys uploaded) yet.

crates/handlers/src/compat/login.rs
reivilibre: Yeah, exactly that. I guess removing the assertion does make sense

crates/handlers/src/compat/login.rs
reivilibre: I think what you're saying is that it won't be removed as long as there are devices inactive for longer than 90d (as those get preferential choice for eviction)?
But if all devices are active within the last 90d, then it is indeed creation time that decides the removal.

I am not sure I feel as strongly about it now as I did last time, though I can still see this biting, e.g. when debugging a new script, or having a script that runs hourly.

crates/handlers/src/compat/login.rs
reivilibre: Yeah this is probably on me. I don't think we have other violations at the time so I guess I didn't think too hard about what to do.

I guess I am also 'inspired' by the authorisation grant policy just showing a generic rejection page. Unclear if it's always a good thing to show the reason to the user (e.g. IP bans etc).

crates/handlers/src/compat/login.rs
reivilibre: Fair enough yeah, if the error gets to the logs I think that's fine

crates/handlers/src/compat/login.rs
reivilibre: > For all intents and purposes, non-cryptographic devices are a completely separate concept and do not exist from the perspective of the cryptography layer since they do not have identity keys, so it is impossible to send them decryption keys.
>
> — <https://spec.matrix.org/v1.18/client-server-api/#recommended-client-behaviour>

FWIW this is probably a decent spec-consistent view on the matter.

I guess the technical term is '(device) identity key' also used at <https://spec.matrix.org/v1.18/client-server-api/#get_matrixclientv3keyschanges>

crates/handlers/src/compat/login.rs
```diff
@@ -51,6 +54,10 @@ static LOGIN_COUNTER: LazyLock<Counter<u64>> = LazyLock::new(|| {
 const TYPE: Key = Key::from_static_str("type");
 const RESULT: Key = Key::from_static_str("result");
 
+/// This matches the `getNinetyDaysAgo()` used in the web UI for "inactive"
+/// sessions.
+const INACTIVE_SESSION_THRESHOLD: chrono::TimeDelta = Duration::days(90);
```
reivilibre, line 59: Thank you

crates/handlers/src/compat/login.rs
MadLittleMods: :+1: The error text is returned in the response:

```json
{
  "errcode": "M_FORBIDDEN",
  "error": "You have reached your hard device limit. Please visit your account page to sign some out."
}
```

crates/handlers/src/compat/login.rs
MadLittleMods: Added the spec reference to the FIXME comment :+1:

crates/handlers/src/compat/login.rs
MadLittleMods: I've bumped it to 2000 based on < 0.001% of accounts on matrix.org having more than 2000 sessions. And we're okay with this operation taking ~2MiB of memory.

Added `minimum_sessions_to_fetch` with a comment to capture all our reasoning behind the value.

I assume this is a good enough value to move this PR forward as-is? Given your worries were slightly tamped based on the conversation in https://github.com/element-hq/matrix-authentication-service/pull/5607#discussion_r3119473869

MadLittleMods: Further bumped this to `2160` based on a script running every hour for 90 days (`90 x 24 = 2160`) and numbers practically aligning with the other reasoning we already had.

crates/handlers/src/compat/login.rs
MadLittleMods: > But if all devices are active within the last 90d, then it is indeed creation time that decides the removal.

For the example where all 200 devices were active within the last 90 days and you need to remove 100 devices, yes :+1:, 100 compat sessions will be removed based on creation date.

Keep in mind, the [minimum fetch limit has bumped](https://github.com/element-hq/matrix-authentication-service/pull/5607#discussion_r3150176535) from `100` to `2160` now :fast_forward: so this example would have to be >= `2160` devices to remove now.

I think you understand the nuances here at this point :fast_forward:

crates/handlers/src/compat/login.rs
```diff
@@ -490,12 +499,274 @@ pub(crate) async fn post(
...
+// updating `INACTIVE_SESSION_THRESHOLD` or whatever contributing factors go
+// into `MINIMUM_SESSIONS_TO_FETCH`.
+const _: () = {
+    assert!(
+        // Update this value if you're ok with the ammount of memory that could be used.
+        MINIMUM_SESSIONS_TO_FETCH == 2160,
+        "Sanity check that you're okay with `MINIMUM_SESSIONS_TO_FETCH` x 1 KiB when fetching sessions? \
+            (read the `MINIMUM_SESSIONS_TO_FETCH` docstring)"
+    );
+};
```
MadLittleMods: This complexity is probably not worth it and instead just rely on the comment to point out the connection to `INACTIVE_SESSION_THRESHOLD`


```suggestion
/// We fetch a minimum number of sessions (2160, more than we need in normal
/// cases) so we can sort by `last_active_at` after it gets back from the
/// database and can get even closer to removing the true oldest sessions.
///
/// The 2160 number was chosen based on someone having a script that runs every
/// hour for the the 90-day `INACTIVE_SESSION_THRESHOLD`. Additionally, it also
/// aligns nicely with < 0.001% of people on matrix.org having less than 2160
/// sessions and reasoning how much memory is reasonable to spend on this
/// operation to get things right. Assuming each row is ~1 KiB (pessimistic high
/// bound, see next paragraph below) we end up at ~2 MiB of memory.
///
/// Each item in the page is `(CompatSession, Option<CompatSsoLogin>)` where
/// `CompatSession` is 192 bytes plus a couple of strings (device name and user
/// agent) (assume pessimistic 512 total bytes). And `CompatSsoLogin` which is
/// also 192 bytes with a `login_token` string which should be no more than 32
/// bytes.
const MINIMUM_SESSIONS_TO_FETCH: usize = 2160;
```

MadLittleMods: Simplified in https://github.com/element-hq/matrix-authentication-service/pull/5607/commits/2c80015fc99c1d9f88410aee2b624e40d04e5176 :white_check_mark:

crates/config/src/sections/experimental.rs
```diff
@@ -116,11 +120,87 @@ impl ExperimentalConfig {
...
+    /// legacy bots/scripts that login over and over (which ideally should
+    /// be using [personal access
+    /// tokens](https://github.com/element-hq/matrix-authentication-service/issues/4492))
+    /// and you want to avoid breaking their operation while maintaining some
+    /// level of sanity with the number of devices that people can have.
+    ///
+    /// [`hard_limit`]: Self::hard_limit
+    /// [`dangerous_hard_limit_eviction`]: Self::dangerous_hard_limit_eviction
+    #[serde(default = "default_false")]
+    pub dangerous_hard_limit_eviction: bool,
```
MadLittleMods, line 191: ## Do we even want to merge `dangerous_hard_limit_eviction`?

hmmm, based on deleting devices being non-trivial task for homeservers to tackle (lots of device list changes, `/sync`, federation, and replication traffic) (as pointed out in the issues linked from https://github.com/element-hq/matrix-authentication-service/issues/4339), we can't really use this option on `matrix.org` as-is. It would cause us to mass delete devices as soon as someone adds another session and trigger all of homeserver performance problems. And given the big offenders are automatic scripts that run regularly (hourly, daily, etc), we would see this hit, day 1 across many accounts.

I'm wondering if we even want to merge this at all 🤔  (I think no)

To stop the bleeding on `matrix.org` we can introduce a new `max_session_threshold`  (name pending) configuration option that would only apply the session limiting to people with less than that number of devices. This prevents us from acting upon a users with 1.3M devices while also enforcing that no one else can exceed the `hard_limit` we set.

The big offenders still get to grow though 😿 There are two ways to tackle these, trickle deletion one by one over time or bulk delete. It's unclear which is more performant.

If trickle deletion is fine,  perhaps instead we should have a `max_sessions_to_remove_per_login` configuration option that will subtract that number of sessions every time they add a new session until they get down to the limits.

Overall, it needs more thought and planning which is why I don't think it's really worth merging this.

The only way this is acceptable is if bulk device deletion per account is ok AND deletion across many accounts (day 1 when this config is enabled ) at once is ok.

I'll separate out the useful bits to a new PR.

reivilibre: hmm, on balance I personally am in favour of merging it + documenting the issues we have with it, if any. We can put it under the experimental config if we want to. **We don't have to use it right away.** (which I think negates most of your concerns if I understand)

It seems likely to me that this PR is a sensible base to expand from. `max_session_threshold` and `max_sessions_to_remove_per_login` feel like small changes on top of this, to the point I would be perfectly happy with those coming as potential follow-ups.

We have in the past (although I can't remember where) said we would not be happy with using errors for device limits on matrix.org, as it's bad UX (and would also break bot operators).
I don't know how much that answer changes with next-gen auth, but not everyone is using it yet.

I feel this is probably still worthwhile on balance.

crates/handlers/src/compat/login.rs
reivilibre: Yep, happy to go with that

reivilibre (approved):

crates/config/src/sections/experimental.rs
MadLittleMods: `max_session_threshold` is a good addition in all cases (already planned to be another PR follow-up even if we didn't merge this).

For `dangerous_hard_limit_eviction`, I guess we can think of it this way:


If trickle deleting devices is fine, `max_sessions_to_remove_per_login` is the better way to go. But this is an additional configuration alongside `dangerous_hard_limit_eviction` so **this PR is good to ship** (although those options probably should be tied together in some fashion):

```yaml
session_limit:
  dangerous_hard_limit_eviction:
     enabled: true
     max_sessions_to_remove_per_login: 100
```

If bulk deleting is fine, we can use `dangerous_hard_limit_eviction` as-is so **this PR is good to ship**.

MadLittleMods:
Thanks for the review @sandhose and @reivilibre 🐝

@sandhose Really good [thoughts](https://github.com/element-hq/matrix-authentication-service/pull/5607#discussion_r3057979830) for how we can workaround the lack of `last_active_at` ordering!

@reivilibre I also really appreciate the effort to [crunch the session numbers from `matrix.org`](https://github.com/element-hq/matrix-authentication-service/pull/5607#discussion_r3139237323) so we have a sense for the state of things alongside the thoughts for the right amount of sessions to pull out from the database!

crates/config/src/sections/experimental.rs
reivilibre: yup, bang on what I was thinking

crates/handlers/src/compat/login.rs
```diff
@@ -490,12 +499,245 @@ pub(crate) async fn post(
...
+                                    .device
+                                    .as_ref()
+                                    .map(mas_data_model::Device::as_str),
+                                "Automatically removing compat session for user (`dangerous_hard_limit_eviction`)"
+                            );
+
+                            // Remove the session
+                            repo.compat_session()
+                                .finish(clock, compat_session.to_owned())
+                                .await?;
```
MadLittleMods, line 597: After removing the sessions, do I need to be scheduling a `SyncDevicesJob` myself?

https://github.com/element-hq/matrix-authentication-service/blob/771d818f6fa7ca98b4ca2b15e44366b9f4f4b19a/crates/handlers/src/compat/logout.rs#L120-L123

Also, in this case, why are we scheduling `SyncDevicesJob` before we `repo.compat_session().finish(...)`?

MadLittleMods: (still looking for an answer here)

I guess this is one area where it would pay to have some Complement tests for MAS. Hit the session limit with `dangerous_hard_limit_eviction` and make sure we see a device list change down `/sync`.

sandhose: Yes you should, that's how they get cleaned up on the homeserver side. When we do a compat login, we want the device to be created synchronously on the homeserver, so that when we respond, the device already exists already. It is an `upsert` anyway, so if the device already exists for some reason (like when we're replacing it, or a concurrent device sync happening) it won't have any effect.

This is different from /logout, or deleting devices in general: we can afford a small delay, as the access token will be invalid right away, and in this case we want to do a full device list sync, because we're not sure we really want to delete the device (if there is for example a concurrent logout and login with the same device ID).

The neat thing about our job queue is that it is backed by the database, and therefore is only visible once the database transaction is committed, it doesn't matter if the session is finished on the repository before or after the job is scheduled.

MadLittleMods: Thanks for the great explanation @sandhose :innocent: 

Handling this in https://github.com/element-hq/matrix-authentication-service/pull/5679

PR #5608, opened 2026-04-03
feat(consent): allow config option to skip consent for given static clients

### LINKED ISSUE #5112: We should skip the "give this app permission to log in" screen if the user has literally just given permission to it.
**Describe the bug**

When a new user registers on (say) EW, or starts a login process, it feels very clunky and bad UX for MAS to then prompt them to give the app permission to access their account... given they just signalled their intent by trying to register/login in the first place.

Surely there should either be a way to pass that intent through to MAS to skip the redundant confirmation, or whitelist clients like EW entirely.

When hallway-testing Element at a conference a few weeks ago, every user I persuaded to signup immediately said "why on earth do I need to grant permission to something I just told it I wanted to do?!" - it feels like a major wart on something which is otherwise very polished.

defaultdino:
I think this sort of closes #5112 . it does not however apply to dynamically configured clients, but I would say that a statically configured client is at some point implicitly "trusted" anyway so one can wonder why this was not implemented from the get-go.

in essence, this allows skipping consent for a static client by adding `skip_consent: true` to the client in `config.yaml`:

```yml
clients:
  - client_id: 01KNAKZTM190PX7V9TPB8MFCT4
    client_auth_method: none
    skip_consent: true # <-- HERE
    redirect_uris:
      - https://example.domain.com/callback
```


sandhose (commented):
I am *very* reticent to the idea of skipping the consent screen. It can be a source of silent logins, especially in case of misbehaving clients.

Moreover, this is only using that on static clients, which feels weird to me, as their use case should be pretty limited, which makes me think you are using MAS not in the way it is intended? So could you please expand first what use case you are trying to solve exactly?

defaultdino:
You have got a point. I don't think it's _entirely_ dismissable to have an admin flag to skip consent for certain static clients though, but it can be debated I guess. I added a check in the last commit that determines whether a user has already given consent (based on their active OAuth sessions), which skips the consent screen when they've already granted the requested scopes, including for dynamic clients (the original case in #5112)

Would you rather I remove the logic for static clients and just keep the consent check or make both an option?

mbosner:
We ship MAS and Synapse as part of a self-hosted groupware suite. Each installation serves a single organization, with federation disabled. Users authenticate through the suite’s own OIDC provider using upstream_oauth2 with skip_confirmation: true.

The chat client is a statically configured public client from the same suite and installation. It is configured by the same operator who runs MAS. In this setup, the consent screen does not give users a meaningful choice. They have already signed in through the suite’s SSO, and the chat client is an integral part of that installation. Since the grant is not remembered, the consent screen appears again with every new session. Defining the client statically in config.yaml is already an explicit trust decision by the operator. A per-client option such as skip_consent: true would allow that decision to be represented in the configuration.

We would enable it for exactly one explicitly configured client - not globally - and can test the implementation against our deployment once it is available.

PR #5612, opened 2026-04-07
Add `oauth`.`device_code_grant_enabled` configuration option

hughns:
This allows the device code/device authorization grant to be disabled.

I couldn't see an existing section that made sense, so I've proposed a new one.

### DIFF
--- a/crates/handlers/src/oauth2/registration.rs
+++ b/crates/handlers/src/oauth2/registration.rs
@@ -21,6 +21,7 @@ use oauth2_types::{
         ClientMetadata, ClientMetadataVerificationError, ClientRegistrationResponse, Localized,
         VerifiedClientMetadata,
     },
+    requests::GrantType,
 };
 use opentelemetry::{Key, KeyValue, metrics::Counter};
 use psl::Psl;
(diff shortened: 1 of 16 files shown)


crates/handlers/src/oauth2/registration.rs
```diff
@@ -240,6 +255,13 @@ pub(crate) async fn post(
     // Validate the body
     let metadata = body.validate()?;
 
+    // Reject if the client requests the device_code grant type and it's disabled
+    if !site_config.device_code_grant_enabled
+        && metadata.grant_types().contains(&GrantType::DeviceCode)
+    {
+        return Err(RouteError::DeviceCodeGrantDisabled);
+    }
```
sandhose: Arguably, we should instead modify the client metadata on the fly but still register it, as per the RFC


```
   Additionally, the authorization server MUST return all registered
   metadata about this client, including any fields provisioned by the
   authorization server itself.  The authorization server MAY reject or
   replace any of the client's requested metadata values submitted
   during the registration and substitute them with suitable values.
   The client or developer can check the values in the response to
   determine if the registration is sufficient for use (e.g., the
   registered "token_endpoint_auth_method" is supported by the client
   software) and determine a course of action appropriate for the client
   software.  
```

sandhose (commented):
Overall looks good, fine with having a new configuration section, but unsure whether we should reject the client registration or not

crates/handlers/src/oauth2/registration.rs
hughns: > Arguably, we should instead modify the client metadata on the fly but still register it, as per the RFC

That sounds reasonable to me.

sandhose: done in [890042e](https://github.com/element-hq/matrix-authentication-service/pull/5612/commits/890042e6a905d6e09d3400ec20eb1064b2d2471f)

sandhose (approved):

PR #5639, opened 2026-04-15
Update policy violation screen when running into the session/device `soft_limit` in interactive contexts

MadLittleMods:
Update policy violation screen when running into the session/device `soft_limit` in interactive contexts (OAuth 2.0 authorization code grants, OAuth 2.0 device authorization grants, legacy `m.login.sso` compatibility login flow).

We tell people to remove X number of devices and then try signing in again. This is much better as compared to yet another interactive special flow for this scenario and figure out how to continue after you remove enough devices.

Part of https://github.com/element-hq/matrix-authentication-service/issues/4339 / https://github.com/element-hq/backend-internal/issues/199 tracking work to limit number of devices.

Follow-up to https://github.com/element-hq/matrix-authentication-service/pull/5607

When running into the session/device limit:

. | Authorization code grant (device authorization grant uses the same template) | Compat SSO login (`m.login.sso`)
--- | --- | ---
After |  | 
Before |  | 

**Note:** The extra weird space next to the client name is being fixed in https://github.com/element-hq/matrix-authentication-service/pull/5641

Then when managing your devices (being addressed in a follow-up PR): https://github.com/element-hq/matrix-authentication-service/pull/5644

Based on these Figma designs: https://www.figma.com/design/HUysJAhv6cK6p1Pc81Fxaa/Matrix-Authentication-Service--MAS-?node-id=6000-19362&p=f&t=RkVd7PMqa2Wab3fv-0

### Dev notes

Testing login in MAS: https://github.com/element-hq/matrix-authentication-service/pull/5642

Matrix spec:

 - OAuth 2.0 aware clients: https://spec.matrix.org/v1.18/client-server-api/#oauth-20-aware-clients
 - Legacy auth: https://spec.matrix.org/v1.18/client-server-api/#legacy-api
     - SSO client login/authentication: https://spec.matrix.org/v1.18/client-server-api/#sso-client-loginauthentication

---

OAuth 2.0:

```
# Both authorization code grant and device authorization grant use this policy
evaluate_authorization_grant

# Authorization code grant
crates/handlers/src/oauth2/authorization/consent.rs
PolicyViolationContext::for_authorization_grant

# Device authorization grant
crates/handlers/src/oauth2/device/consent.rs
PolicyViolationContext::for_device_code_grant

templates/pages/policy_violation.html
```

Compat SSO login:

```
crates/handlers/src/compat/login_sso_complete.rs
evaluate_compat_login
CompatLoginPolicyViolationContext::for_violations

[...]

PR #5644, opened 2026-04-17
Show session/device limit errors on account page

MadLittleMods:
Show session/device limit errors on account page

Follow-up to https://github.com/element-hq/matrix-authentication-service/pull/5639

Best to just be transparent about the session/device limit so this is always available under the info icon in the "N devices ℹ️" header with a tooltip "You've used 8/12 device slots".

If you're approaching the `soft_limit` (within 20% of the limit), we show a "heads up" warning when viewing your list of devices. There is some smarter logic here to make sure this doesn't appear if the limit is low and you have no devices (see the code).

If you've hit or are over the `soft_limit`, we show a red error when viewing your list of devices. We also show a red error indicator next to the "Devices" item in the nav bar as an action item for someone to address.

Approaching limit | Hit/over limit
--- | ---
 | 

To point out the baseline differences, we've added "N devices" header which changes to "X devices match your filter (Y total)" when you filter the list to show inactive devices.

. | Before | After
--- | --- | ---
Unfiltered |  | 
Filtered |  | 

Based on these Figma designs but I've taken some liberty here: https://www.figma.com/design/HUysJAhv6cK6p1Pc81Fxaa/Matrix-Authentication-Service--MAS-?node-id=6000-19362&p=f&t=RkVd7PMqa2Wab3fv-0 (update: this PR has subsequently gone through a design review so it's good to go on that front)

### Dev notes

Element's design system: https://github.com/element-hq/compound

```
# Device lists
frontend/src/routes/_account.sessions.index.tsx

# NavBar
frontend/src/routes/_account.tsx
```

Testing: https://testing-library.com/docs/dom-testing-library/

---

`aria-role="status"`, https://www.w3.org/WAI/WCAG22/Understanding/status-messages

### Todo

 - [x] Only show error in navbar when over `soft_limit`
 - [x] Make sure session limit error on accounts -> devices page disappears as you remove enough devices
 - [x] Figure out frontend tests that are necessary


frontend/src/routes/_account.tsx
```diff
@@ -47,7 +58,23 @@ function Account(): React.ReactElement {
...
+  if (
+    sessionLimit &&
+    viewer.unfilteredAppSessions.totalCount >= sessionLimit.softLimit
+  ) {
+    sessionLimitHitIcon = (
+      <Tooltip label={t("frontend.nav.device_limit_error")}>
+        <IconErrorSolid className={styles.navBarErrorIcon} />
+      </Tooltip>
+    );
+  }
```
MadLittleMods, line 74: I'm noticing that this error icon doesn't disappear from the `NavBar` when you remove devices to be under the limit.

I see this re-evaluated as expected when devices are added but when devices are removed, it doesn't re-evaluate this logic 🤔 

Re-evaluation happens just fine for device section though (`frontend/src/routes/_account.sessions.index.tsx`)

MadLittleMods: Fixed by https://github.com/element-hq/matrix-authentication-service/pull/5644/commits/cc7f786af3f2cd2d0878b4a4270a1a7f65460382

To explain:

`queryClient.invalidateQueries({ queryKey: ["sessionsOverview"] })` causes the GraphQL queries in `frontend/src/routes/_account.sessions.index.tsx` to fetch again.

Similarly, we just need to add `queryClient.invalidateQueries({ queryKey: ["currentUserGreeting"] })` to invalide the queries in `frontend/src/routes/_account.tsx`

frontend/tests/routes/account/sessions.test.tsx
```diff
@@ -0,0 +1,107 @@
...
+        mockSessionsOverviewQuery(() =>
+          HttpResponse.json({
+            data: {
+              viewer: {
+                __typename: "User",
+                id: "123",
+                // FIXME: Unclear how to deal with fragment masking (`$fragmentRefs`)
+                browserSessions: {
+                  totalCount: 3,
+                },
```
MadLittleMods: It's unclear how to fix this situation.

```
$ npm run lint
...
tests/routes/account/sessions.test.tsx:33:17 - error TS2353: Object literal may only specify known properties, and 'browserSessions' does not exist in type '{ __typename: "User"; id: string; unfilteredAppSessions: { __typename?: "AppSessionConnection" | undefined; totalCount: number; }; } & { ' $fragmentRefs'?: { BrowserSessionsOverview_UserFragment: BrowserSessionsOverview_UserFragment; } | undefined; }'.

33                 browserSessions: {
                   ~~~~~~~~~~~~~~~


Found 1 error in tests/routes/account/sessions.test.tsx:33
```

These are the generated types in `frontend/src/gql/graphql.ts`

```
export type SessionsOverviewQuery = { __typename?: 'Query', viewer:
    | { __typename: 'Anonymous' }
    | (
      { __typename: 'User', id: string, unfilteredAppSessions: { __typename?: 'AppSessionConnection', totalCount: number } }
      & { ' $fragmentRefs'?: { 'BrowserSessionsOverview_UserFragment': BrowserSessionsOverview_UserFragment } }
    )
  , siteConfig: { __typename?: 'SiteConfig', sessionLimit?: { __typename?: 'SessionLimitConfig', softLimit: number } | null } };
```

```
export type BrowserSessionsOverview_UserFragment = { __typename?: 'User', id: string, browserSessions: { __typename?: 'BrowserSessionConnection', totalCount: number } } & { ' $fragmentName'?: 'BrowserSessionsOverview_UserFragment' };
```

The actual response I see in the browser looks like this (includes `browserSessions`):

```json
{
    "data": {
        "viewer": {
            "__typename": "User",
            "id": "user:01KN32VDFXH7KSYS4902GF83X2",
            "browserSessions": {
                "totalCount": 3
            },
            "unfilteredAppSessions": {
                "totalCount": 11
            }
        },
        "siteConfig": {
            "sessionLimit": {
                "softLimit": 12
            }
        }
    }
}
```

frontend/tests/routes/account/sessions.test.tsx
```diff
@@ -0,0 +1,107 @@
...
+import { renderPage, server } from "../render";
+
+describe("Account sessions page", () => {
+  it("renders the page", async () => {
+    const { asFragment } = await renderPage("/sessions");
+    expect(asFragment()).toMatchSnapshot();
+  });
+
+  describe("session limit", () => {
+    it("displays an error if they've hit the session soft_limit", async () => {
```
MadLittleMods, line 26: Just a quick sanity check test that we show a session/device limit error on the account page.

I don't think it's worth having a test mirroring all of the other cases.

frontend/tests/routes/account/__snapshots__/sessions.test.tsx.snap
MadLittleMods, line 1: Don't be scared of the diff numbers. This snapshot file makes up 1.1k additions out of the total diff: +1,498
-20

crates/handlers/src/graphql/model/site_config.rs
```diff
@@ -79,6 +82,25 @@ pub enum CaptchaService {
     HCaptcha,
 }
 
+#[derive(SimpleObject)]
+pub struct SessionLimitConfig {
+    pub soft_limit: u64,
```
sandhose, line 88: Would be nice to have docstrings here so that it gets down the generated GraphQL schema and the generated typescript types

frontend/src/routes/_account.sessions.index.tsx
```diff
@@ -153,11 +166,110 @@ function Sessions(): React.ReactElement {
   // We reverse the list as we are paginating backwards
   const edges = [...appSessions.edges].reverse();
 
+  // By default, we just show a "X devices" header
+  let deviceHeaderText = t(
+    "frontend.user_sessions_overview.num_sessions_header",
+    {
+      num_sessions: appSessions.totalCount,
+    },
+  );
```
sandhose, line 177: This should be a [plural](https://www.i18next.com/translation-function/plurals) and therefore use the `count` placeholder

frontend/src/routes/_account.sessions.index.tsx
```diff
@@ -153,11 +166,110 @@ function Sessions(): React.ReactElement {
...
+  if (
+    overviewViewer.unfilteredAppSessions.totalCount !== appSessions.totalCount
+  ) {
+    deviceHeaderText = t(
+      "frontend.user_sessions_overview.num_sessions_filtered_header",
+      {
+        filtered_count: appSessions.totalCount,
+        num_sessions: overviewViewer.unfilteredAppSessions.totalCount,
+      },
+    );
```
sandhose, line 195: This should be a [plural](https://www.i18next.com/translation-function/plurals) and therefore use the `count` placeholder; you will sadly have to split it though, because you can't have multiple plurals in the same translated string. It is fine to compose them, e.g.

```
const deviceCount = t("device_count", { count: 10 });
const deviceHeaderText = t("header", { count: 20, deviceCount });
```

frontend/src/routes/_account.sessions.index.tsx
```diff
@@ -153,11 +166,110 @@ function Sessions(): React.ReactElement {
...
+  }
+  // Show a warning when they're approaching the session limit
+  else if (
+    sessionLimit &&
+    // Avoid showing a problem when they don't have any devices yet
+    overviewViewer.unfilteredAppSessions.totalCount > 0 &&
+    // When they're approaching the limit (20% headroom, round up)
+    overviewViewer.unfilteredAppSessions.totalCount +
+      Math.ceil(sessionLimit.softLimit * 0.2) >=
```
sandhose, line 268: I think that's a fine aproximation

frontend/src/routes/_account.module.css
sandhose, line 1: I see why it's there but I would like to avoid CSS modules in the `routes/` folder. Would you mind instead defining a `<NavItemErrorIcon >` or something in the `NavItem` module?

frontend/src/routes/_account.sessions.index.tsx
```diff
@@ -153,11 +166,110 @@ function Sessions(): React.ReactElement {
...
+        )}
+        data-testid="device-limit-error"
+      >
+        {t(
+          "frontend.user_sessions_overview.hit_session_limit_warning_description",
+          {
+            limit: sessionLimit.softLimit,
+            num_sessions: overviewViewer.unfilteredAppSessions.totalCount,
+          },
+        )}
```
sandhose, line 257: Plural

frontend/src/routes/_account.sessions.index.tsx
```diff
@@ -153,11 +166,110 @@ function Sessions(): React.ReactElement {
...
+        data-testid="device-limit-warning"
+      >
+        {t(
+          "frontend.user_sessions_overview.approaching_session_limit_warning_description",
+          {
+            limit: sessionLimit.softLimit,
+            num_sessions: overviewViewer.unfilteredAppSessions.totalCount,
+          },
+        )}
```
sandhose, line 290: Plural

frontend/tests/routes/account/sessions.test.tsx
sandhose: Something like this: https://github.com/element-hq/matrix-authentication-service/blob/6d3082e9dcd22cd3f35455decb87ac306d7c01c5/frontend/tests/mocks/handlers.ts#L36-L56

sandhose (changes requested):

frontend/tests/routes/account/sessions.test.tsx
MadLittleMods: :bow: Thank you! Perfect example to work from

Google wasn't being helpful. Glad we already have a solution

frontend/src/routes/_account.sessions.index.tsx
MadLittleMods: :+1: I've updated to use `count` for the main plural. For the `(Y total)` piece, I've left it as-is as I couldn't find a clear answer whether it matters in some languages. I initially made it verbose but then asking an LLM about it, it couldn't come up with an example beyond "they may want to expand `(Y total)` to `(Y devices total)`, which would then need pluralization too."

> `(Y total)` — This is a standalone parenthetical with an implicit noun. The word "total" doesn't inflect, and the noun it refers to ("devices") is already established earlier in the sentence. In practice, most i18n systems and translators treat bare (N total) as invariant. You could handle it, but it's usually unnecessary overhead and most languages won't require it in this position.

Seems like some assumptions to me but we could go with it for the sake of simplicity and change it if some translator flags it.

If we wanted to play it safe, we can do this:

```json
      "num_sessions_filtered_header": {
        "header": "{{count}} devices match your filter ({{unfiltered_session_total}})",
        "unfiltered_session_total": "{{count}} total"
      },
```

```javascript
    let unfilteredSessionTotal = t(
      "frontend.user_sessions_overview.num_sessions_filtered_header.unfiltered_session_total",
      {
        count: overviewViewer.unfilteredAppSessions.totalCount,
      },
    );
    deviceHeaderText = t(
      "frontend.user_sessions_overview.num_sessions_filtered_header.header",
      {
        count: appSessions.totalCount,
        unfiltered_session_total: unfilteredSessionTotal,
      },
    );
```

frontend/src/routes/_account.sessions.index.tsx
MadLittleMods: These are pretty tricky pluralization scenarios. I've added comments with my thinking :shrug:

crates/handlers/src/graphql/model/site_config.rs
MadLittleMods: :thinking: I'd hate to repeat what we already define in the config:

https://github.com/element-hq/matrix-authentication-service/blob/e91de1aeaf56ca535545e971192f3c467db50361/crates/config/src/sections/experimental.rs#L142-L193

I've added a little cross-link to the main type but it doesn't give us the GraphQL schema comments that you're after.

frontend/src/components/UserSessionsOverview/BrowserSessionsOverview.tsx
```diff
@@ -14,8 +14,6 @@ import styles from "./BrowserSessionsOverview.module.css";
 
 export const FRAGMENT = graphql(/* GraphQL */ `
   fragment BrowserSessionsOverview_user on User {
-    id
```
MadLittleMods, line 17: Removed as it's unused as far as I can tell

crates/handlers/src/graphql/model/site_config.rs
sandhose: Maybe a simplified description, just the first paragraph?

frontend/src/routes/_account.sessions.index.tsx
sandhose: I'd prefer the safer version, I don't want to make assumptions about how things work in various languages :D

frontend/src/routes/_account.sessions.index.tsx
sandhose: That is indeed a tricky one. FTR in Element Admin we use FormatJS instead of react-i18next, and uses the ICU message format, which handles case like this. You can have multiple plural placeholders in the same message: https://formatjs.github.io/docs/core-concepts/icu-syntax#plural-format

sandhose (approved):
The tricky plurals suck a little bit, but that is probably good enough for now; we can always iterate on this later. Approved, modulo that I do want to fix the one with `(N total)` though :)

frontend/src/routes/_account.sessions.index.tsx
MadLittleMods: Updated :white_check_mark: 

I'd say we're making just as many assumptions with the other spots as well :shrug: Overall, seems like we just need a better i18n framework for this like you mentioned in https://github.com/element-hq/matrix-authentication-service/pull/5644#discussion_r3275983163.

crates/handlers/src/graphql/model/site_config.rs
MadLittleMods: Going to skip for now :fast_forward: 

I tried it out and it looks like it doesn't actually flow down to the actual usage (only on the `SessionLimitConfig` type) so it's not useful.

And since it's something consumed inside the MAS project (which has both the Rust and frontend), people can take the hint from the reference given.

MadLittleMods:
Thanks for the review @sandhose 🐿️

PR #5645, opened 2026-04-19
Remove whitespaces from username and email

bxdxnn:
Remove whitespaces from MXID localpart and email as they both don't allow whitespaces, and people have commonly leading spaces being added by the keyboard autosuggestions.

### DIFF
--- a/crates/data-model/src/utils.rs
+++ b/crates/data-model/src/utils.rs
@@ -11,3 +11,8 @@ use crate::clock::Clock;
 pub type BoxClock = Box<dyn Clock + Send>;
 /// A boxed random number generator
 pub type BoxRng = Box<dyn CryptoRngCore + Send>;
+
+#[must_use]
+pub fn normalize_username(input: &str) -> String {
+    input.chars().filter(|c| !c.is_whitespace()).collect()
+}
--- a/crates/handlers/src/admin/v1/user_emails/add.rs
+++ b/crates/handlers/src/admin/v1/user_emails/add.rs
@@ -124,9 +124,10 @@ pub async fn handler(
         .ok_or(RouteError::UserNotFound(params.user_id))?;
 
     // Validate the email
-    if let Err(source) = lettre::Address::from_str(&params.email) {
+    let email = params.email.trim().to_owned();
+    if let Err(source) = lettre::Address::from_str(&email) {
         return Err(RouteError::EmailNotValid {
-            email: params.email,
+            email: email.clone(),
             source,
         });
     }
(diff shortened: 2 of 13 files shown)


crates/storage-pg/migrations/20260419000000_normalize_usernames.sql
sandhose, line 1: This has no reason to exist and is completely inefficient. This will completely lock the users table during the migration, which is a no-go on big server

crates/handlers/src/views/login.rs
```diff
@@ -160,7 +161,7 @@ pub(crate) async fn post(
     // Validate the form
     let mut form_state = form.to_form_state();
 
-    if form.username.is_empty() {
+    if form.username.is_empty() || normalize_username(&form.username).is_empty() {
```
sandhose, line 164: `normalize_username` called unnecessarily twice

crates/handlers/src/views/register/password.rs
sandhose, line 1: nowadays we're doing username normalization live on the field anyway for the registration flow

crates/handlers/src/compat/login.rs
sandhose, line 1: that's the responsibility of the client to do that

crates/handlers/src/admin/v1/users/add.rs
sandhose, line 1: why would we want this behaviour on the admin API?

crates/handlers/src/admin/v1/user_emails/add.rs
sandhose, line 1: why would we want this behaviour on the admin API?

crates/data-model/src/utils.rs
```diff
@@ -11,3 +11,8 @@ use crate::clock::Clock;
 pub type BoxClock = Box<dyn Clock + Send>;
 /// A boxed random number generator
 pub type BoxRng = Box<dyn CryptoRngCore + Send>;
+
+#[must_use]
+pub fn normalize_username(input: &str) -> String {
+    input.chars().filter(|c| !c.is_whitespace()).collect()
```
sandhose, line 17: this is some agressive manipulation, quite debatable. I would possibly agree on a trim, but not this

sandhose (changes requested):
Please state exactly what you are trying to fix before blindly applying a trim() everywhere

bxdxnn:
> Please state exactly what you are trying to fix before blindly applying a trim() everywhere

Having Email entered with a whutespace on login didn't trim it before

PR #5660, opened 2026-04-28
Always gate sign in `username` input behind flags

jaywink:
Currently, it seems we always show a username input on the sign in page, even if password login is disabled. This feels confusing, and typing a username here when you have multiple OIDC upstreams causes an error ("Failed to deserialize form body: missing field `password`").

image


sandhose (approved):

sandhose:
Actually closing in favor of #5603 which essentially does the same thing

PR #5664, opened 2026-05-05
Harden the security of our GitHub Actions

anoadragon453:
This PR makes a number of changes to harden our GitHub Actions workflows, in line with our internal [security hardening guidelines](https://docs.google.com/document/d/1nR8oCHyovy-YNrM9sJdEz3yj6EZdjdRNnsadVV_b6cY/edit?tab=t.0).

[`zizmor`](https://zizmor.sh/) was used to find (and automate the fixing of) most of these changes. Recommended to be reviewed commit-by-commit.

There's still a number of things that need fixing:

- Cache poisoning mitigations on release workflows (see https://github.com/element-hq/matrix-authentication-service/pull/5664/commits/72e5ae40b08e3c110462c2ed0d3ea68763442b3e as a WIP staring point)
- Excessive permissions on jobs (https://docs.zizmor.sh/audits/#excessive-permissions)
- Replace `action-gh-release` usage with built-in `gh release` command (https://docs.zizmor.sh/audits/#superfluous-actions)
- Adding the [`zizmor` GitHub Action](https://github.com/zizmorcore/zizmor-action) to CI so we don't regress any of these hardening changes accidentally.

These can be done in a follow-up PR.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -298,7 +298,7 @@ jobs:
         id: output
         run: |
           echo 'metadata<<EOF' >> $GITHUB_OUTPUT
-          echo '${STEPS_BAKE_OUTPUTS_METADATA}' | jq -c 'with_entries(select(.value | (type == "object" and has("containerimage.digest")))) | map_values({ digest: .["containerimage.digest"], tags: (.["image.name"] | split(",")) })' >> $GITHUB_OUTPUT
+          echo "$STEPS_BAKE_OUTPUTS_METADATA" | jq -c 'with_entries(select(.value | (type == "object" and has("containerimage.digest")))) | map_values({ digest: .["containerimage.digest"], tags: (.["image.name"] | split(",")) })' >> $GITHUB_OUTPUT
           echo 'EOF' >> $GITHUB_OUTPUT
         env:
           STEPS_BAKE_OUTPUTS_METADATA: ${{ steps.bake.outputs.metadata }}
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -50,6 +50,7 @@ jobs:
         with:
           # Need a full clone so that `git describe` reports the right version
           fetch-depth: 0
+          persist-credentials: false
 
       - name: Compute version and timestamp out of git history
         id: git
@@ -68,6 +69,8 @@ jobs:
     steps:
       - name: Checkout the code
         uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6
+        with:
+          persist-credentials: false
 
       - uses: ./.github/actions/build-frontend
       - uses: ./.github/actions/build-policies
@@ -113,6 +116,8 @@ jobs:
     steps:
       - name: Checkout the code
         uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6
+        with:
+          persist-credentials: false
 
       - name: Install Rust toolchain
         uses: dtolnay/rust-toolchain@stable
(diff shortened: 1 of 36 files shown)


anoadragon453:
~~The remaining CI failure is unrelated to this PR, and fixed by https://github.com/element-hq/matrix-authentication-service/pull/5665~~ Merged!

.github/workflows/ci.yaml
```diff
@@ -213,17 +229,18 @@ jobs:
...
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6
+        with:
+          persist-credentials: false
 
       - name: Install Rust toolchain
-        uses: dtolnay/rust-toolchain@1.93.0
-        with:
-          components: clippy
+        # Install the minimal toolchain, which includes rustc, rustdoc, and cargo.
+        run: rustup toolchain install stable --profile minimal --component clippy
```
sandhose: The clippy lints you had to fix is exactly why we pinned it to a particular version and regularly update that (for example https://github.com/element-hq/matrix-authentication-service/pull/4900)

sandhose (changes requested):
Please revert the Clippy lint fixes by pinning Clippy to the same version, but other than that lgtm

.github/workflows/ci.yaml
anoadragon453: Ah, that was unintentional 🤦 

Restored it to 1.93.0 and reverted the commits in 49ad5c79e1a8ab36da216538ecc818f72c6d1f58. Hopefully that commit can be used as a reference point when it comes time to update clippy separately.

.github/workflows/ci.yaml
```diff
@@ -213,17 +229,18 @@ jobs:
...
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6
+        with:
+          persist-credentials: false
 
       - name: Install Rust toolchain
-        uses: dtolnay/rust-toolchain@1.93.0
-        with:
-          components: clippy
+        # Install the minimal toolchain, which includes rustc, rustdoc, and cargo.
+        run: rustup toolchain install 1.93.0 --profile minimal --component clippy
```
sandhose: I think we need `--override` so that it makes that toolchain the "default" toolchain for this directory?

```suggestion
        # Install the minimal toolchain with clippy, pinned to a version kept in sync with Dockerfile
        run: rustup toolchain install 1.93.0 --profile minimal --component clippy --override
```

.github/workflows/build.yaml
```diff
@@ -293,8 +298,10 @@ jobs:
         id: output
         run: |
           echo 'metadata<<EOF' >> $GITHUB_OUTPUT
-          echo '${{ steps.bake.outputs.metadata }}' | jq -c 'with_entries(select(.value | (type == "object" and has("containerimage.digest")))) | map_values({ digest: .["containerimage.digest"], tags: (.["image.name"] | split(",")) })' >> $GITHUB_OUTPUT
+          echo '${STEPS_BAKE_OUTPUTS_METADATA}' | jq -c 'with_entries(select(.value | (type == "object" and has("containerimage.digest")))) | map_values({ digest: .["containerimage.digest"], tags: (.["image.name"] | split(",")) })' >> $GITHUB_OUTPUT
```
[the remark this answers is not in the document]
anoadragon453: Looks like this might be an edge case in `zizmor`. Something to report upstream.

.github/workflows/ci.yaml
anoadragon453: Ohh, I was looking for such an option. Thanks! Updated in ea9f324e7 and 451761c39.

sandhose (approved):

PR #5667, opened 2026-05-06
Fix client linking to `none` on policy violation screen

MadLittleMods:
Fix client linking to `none` on policy violation screen. This happened when the there is no `client_uri` configured.

`config.yaml`
```yaml
clients:
  - client_id: 00000000000000000000SEC0ND
    client_name: "my-test-client"
    client_auth_method: none
```

Before:

```
<a target="_blank" href="none" class="cpd-link" data-kind="primary">my-test-client</a>
```

After (plain text):

```
my-test-client
```

---

Spawning from design review of [device limits](https://github.com/element-hq/matrix-authentication-service/issues/4339) in MAS and @amshakal asking the question "why is this a link?". At the time, since clicking on the link lead to the consent screen (`http://localhost:8080/consent/none`), this appeared like some weird loop and the link isn't necessary (especially when `client_uri` isn't defined).

But it looks like we have a pattern for this kind of thing to fallback to the `redirect_uri` if the `client_uri` isn't defined. Having some URL is nice to better disambiguate and identify which client you're interacting with. See https://github.com/element-hq/matrix-authentication-service/pull/5667#discussion_r3201605734 for more discussion on what patterns we should be using.

### Dev notes

Policy violation screen for reference:

Testing:

```
DATABASE_URL='postgresql://postgres:postgres@localhost/mas' cargo nextest run -p mas-templates tests::check_builtin_templates
```

uses all of the `TemplateContext`/`sample(...)`/`samples(...)` logic throughout the code to render the templates.

Relevant:

 - [`PolicyViolationContext`](https://github.com/element-hq/matrix-authentication-service/blob/1345f6b5020db1a3df7447770b0e633b14c00fd0/crates/templates/src/context.rs#L803-L843), [`ConsentContext`](https://github.com/element-hq/matrix-authentication-service/blob/1345f6b5020db1a3df7447770b0e633b14c00fd0/crates/templates/src/context.rs#L740-L770),  [`DeviceConsentContext`](https://github.com/element-hq/matrix-authentication-service/blob/1345f6b5020db1a3df7447770b0e633b14c00fd0/crates/templates/src/context.rs#L1823-L1858)
 -  https://github.com/element-hq/matrix-authentication-service/blob/1345f6b5020db1a3df7447770b0e633b14c00fd0/crates/data-model/src/oauth2/client.rs#L177-L227

### DIFF
--- a/crates/templates/src/context.rs
+++ b/crates/templates/src/context.rs
@@ -247,7 +247,7 @@ impl<T: TemplateContext> TemplateContext for WithCsrf<T> {
 }
 
 /// Context with a user session in it
-#[derive(Serialize)]
+#[derive(Serialize, Debug)]
 pub struct WithSession<T> {
     current_session: BrowserSession,
 
@@ -783,7 +783,7 @@ impl ConsentContext {
     }
 }
 
-#[derive(Serialize)]
+#[derive(Serialize, Debug)]
 #[serde(tag = "grant_type")]
 enum PolicyViolationGrant {
     #[serde(rename = "authorization_code")]
@@ -793,7 +793,7 @@ enum PolicyViolationGrant {
 }
 
 /// Context used by the `policy_violation.html` template
-#[derive(Serialize)]
+#[derive(Serialize, Debug)]
 pub struct PolicyViolationContext {
     grant: PolicyViolationGrant,
     client: Client,
--- a/frontend/src/styles/cpd-link.css
+++ b/frontend/src/styles/cpd-link.css
@@ -18,33 +18,33 @@ there is something to interact with */
 .cpd-link[href] {
   text-decoration: underline;
 
-  .cpd-link[data-kind="primary"] {
+  &[data-kind="primary"] {
     color: var(--cpd-color-text-primary);
   }
 
-  .cpd-link[data-kind="critical"] {
+  &[data-kind="critical"] {
     color: var(--cpd-color-text-critical-primary);
   }
 
   @media (hover) {
-    .cpd-link[data-kind="primary"]:hover {
+    &[data-kind="primary"]:hover {
       background: var(--cpd-color-gray-300);
     }
 
-    .cpd-link[data-kind="critical"]:hover {
+    &[data-kind="critical"]:hover {
       background: var(--cpd-color-red-300);
     }
   }
 
-  .cpd-link:active {
+  &:active {
     color: var(--cpd-color-text-on-solid-primary);
   }
 
-  .cpd-link[data-kind="primary"]:active {
+  &[data-kind="primary"]:active {
     background: var(--cpd-color-text-primary);
   }
 
-  .cpd-link[data-kind="critical"]:active {
+  &[data-kind="critical"]:active {
     background: var(--cpd-color-text-critical-primary);
   }
 }
--- a/frontend/src/styles/cpd-link.css
+++ b/frontend/src/styles/cpd-link.css
@@ -48,4 +48,3 @@ there is something to interact with */
     background: var(--cpd-color-text-critical-primary);
   }
 }
-
--- a/frontend/src/styles/cpd-link.css
+++ b/frontend/src/styles/cpd-link.css
@@ -7,44 +7,39 @@
 
 .cpd-link {
   display: inline-block;
+  text-decoration: underline;
   color: var(--cpd-color-text-primary);
   font-weight: var(--cpd-font-weight-medium);
   border-radius: var(--cpd-radius-pill-effect);
   padding-inline: 0.25rem;
 }
 
-/* Only underline and add hover/active styles if it actually links to something and
-there is something to interact with */
-.cpd-link[href] {
-  text-decoration: underline;
+.cpd-link[data-kind="primary"] {
+  color: var(--cpd-color-text-primary);
+}
 
-  &[data-kind="primary"] {
-    color: var(--cpd-color-text-primary);
+.cpd-link[data-kind="critical"] {
+  color: var(--cpd-color-text-critical-primary);
+}
+
+@media (hover) {
+  .cpd-link[data-kind="primary"]:hover {
+    background: var(--cpd-color-gray-300);
   }
 
-  &[data-kind="critical"] {
-    color: var(--cpd-color-text-critical-primary);
+  .cpd-link[data-kind="critical"]:hover {
+    background: var(--cpd-color-red-300);
   }
+}
 
-  @media (hover) {
-    &[data-kind="primary"]:hover {
-      background: var(--cpd-color-gray-300);
-    }
+.cpd-link:active {
+  color: var(--cpd-color-text-on-solid-primary);
+}
 
-    &[data-kind="critical"]:hover {
-      background: var(--cpd-color-red-300);
-    }
-  }
+.cpd-link[data-kind="primary"]:active {
+  background: var(--cpd-color-text-primary);
+}
 
-  &:active {
-    color: var(--cpd-color-text-on-solid-primary);
-  }
-
-  &[data-kind="primary"]:active {
-    background: var(--cpd-color-text-primary);
-  }
-
-  &[data-kind="critical"]:active {
-    background: var(--cpd-color-text-critical-primary);
-  }
+.cpd-link[data-kind="critical"]:active {
+  background: var(--cpd-color-text-critical-primary);
 }
--- a/frontend/src/styles/cpd-link.css
+++ b/frontend/src/styles/cpd-link.css
@@ -7,39 +7,45 @@
 
 .cpd-link {
   display: inline-block;
-  text-decoration: underline;
   color: var(--cpd-color-text-primary);
   font-weight: var(--cpd-font-weight-medium);
   border-radius: var(--cpd-radius-pill-effect);
   padding-inline: 0.25rem;
 }
 
-.cpd-link[data-kind="primary"] {
-  color: var(--cpd-color-text-primary);
-}
+/* Only underline and add hover/active styles if it actually links to something and
+there is something to interact with */
+.cpd-link[href] {
+  text-decoration: underline;
 
-.cpd-link[data-kind="critical"] {
-  color: var(--cpd-color-text-critical-primary);
-}
-
-@media (hover) {
-  .cpd-link[data-kind="primary"]:hover {
-    background: var(--cpd-color-gray-300);
+  .cpd-link[data-kind="primary"] {
+    color: var(--cpd-color-text-primary);
   }
 
-  .cpd-link[data-kind="critical"]:hover {
-    background: var(--cpd-color-red-300);
+  .cpd-link[data-kind="critical"] {
+    color: var(--cpd-color-text-critical-primary);
+  }
+
+  @media (hover) {
+    .cpd-link[data-kind="primary"]:hover {
+      background: var(--cpd-color-gray-300);
+    }
+
+    .cpd-link[data-kind="critical"]:hover {
+      background: var(--cpd-color-red-300);
+    }
+  }
+
+  .cpd-link:active {
+    color: var(--cpd-color-text-on-solid-primary);
+  }
+
+  .cpd-link[data-kind="primary"]:active {
+    background: var(--cpd-color-text-primary);
+  }
+
+  .cpd-link[data-kind="critical"]:active {
+    background: var(--cpd-color-text-critical-primary);
   }
 }
 
-.cpd-link:active {
-  color: var(--cpd-color-text-on-solid-primary);
-}
-
-.cpd-link[data-kind="primary"]:active {
-  background: var(--cpd-color-text-primary);
-}
-
-.cpd-link[data-kind="critical"]:active {
-  background: var(--cpd-color-text-critical-primary);
-}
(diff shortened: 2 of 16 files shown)


templates/pages/policy_violation.html
```diff
@@ -29,7 +29,7 @@ <h1 class="title">{{ _("mas.policy_violation.heading") }}</h1>
         {{ icon.web_browser() }}
       </div>
       {% endif %}
-      <a target="_blank" href="{{ client.client_uri }}" class="cpd-link" data-kind="primary">{{ client.client_name or client.client_id }}</a>
+      <a target="_blank" href="{{ (client.client_uri or grant.redirect_uri) | simplify_url }}" class="cpd-link" data-kind="primary">{{ client.client_name or client.client_id }}</a>
```
MadLittleMods: This matches our other template usage:

https://github.com/element-hq/matrix-authentication-service/blob/1345f6b5020db1a3df7447770b0e633b14c00fd0/templates/pages/consent.html#L32

https://github.com/element-hq/matrix-authentication-service/blob/1345f6b5020db1a3df7447770b0e633b14c00fd0/templates/pages/device_consent.html#L35

crates/templates/src/context.rs
```diff
@@ -247,7 +247,7 @@ impl<T: TemplateContext> TemplateContext for WithCsrf<T> {
 }
 
 /// Context with a user session in it
-#[derive(Serialize)]
+#[derive(Serialize, Debug)]
```
MadLittleMods, line 250: Added these `Debug` traits so I could print out the template context and try to find out where the `none` is coming from. It turns out `client_uri` is actually `None` and [`minininja`](https://docs.rs/minijinja/latest/minijinja/syntax/index.html) is the one doing the `none` magic

templates/pages/policy_violation.html
MadLittleMods: ### :x: Template render  failure

```
$ DATABASE_URL='postgresql://postgres:postgres@localhost/mas' cargo nextest run -p mas-templates tests::check_builtin_templates
...
failures:
    tests::check_builtin_templates

test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 1 filtered out; finished in 0.32s


--- TRY 2 STDERR:        mas-templates tests::check_builtin_templates ---

thread 'tests::check_builtin_templates' (3267342) panicked at crates/templates/src/lib.rs:540:61:
called `Result::unwrap()` on an `Err` value: Failed to render sample template "pages/policy_violation.html"-SampleIdentifier { components: [("index", "3"), ("browser-session", "0"), ("locale", "cs")] } with context {"lang":"cs","csrf_token":"fake_csrf_token","current_session":{"id":"01KQZC5PW4C048SZ366JYG0YJA","user":{"id":"01KQZC5PW48GQ219WF5A711WAD","username":"john","sub":"123-456","created_at":"2026-05-06T19:27:56.292412967Z","locked_at":null,"deactivated_at":null,"can_request_admin":false,"is_guest":false},"created_at":"2026-05-06T19:27:56.292412967Z","finished_at":null,"user_agent":"Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/93.0.0.0 Safari/537.36","last_active_at":"2026-05-06T19:27:56.292412967Z","last_active_ip":null},"grant":{"grant_type":"urn:ietf:params:oauth:grant-type:device_code","id":"01KQZC5PW45MM6WCPW0DNFN5Y4","state":"pending","client_id":"01KQZC5PW4W0JJR6NKMZYNX9V0","scope":"openid","user_code":"D3ANKP","device_code":"knxlsBJk42KmxQsskPIIu7LikHCJxCx0","created_at":"2026-05-06T19:22:56.292412967Z","expires_at":"2026-05-06T19:52:56.292412967Z","ip_address":null,"user_agent":null},"client":{"id":"01KQZC5PW4W0JJR6NKMZYNX9V0","client_id":"client2","metadata_digest":null,"encrypted_client_secret":null,"application_type":"native","redirect_uris":["https://client2.example.com/redirect"],"grant_types":["authorization_code","refresh_token"],"client_name":null,"logo_uri":null,"client_uri":null,"policy_uri":null,"tos_uri":null,"jwks":null,"id_token_signed_response_alg":null,"userinfo_signed_response_alg":null,"token_endpoint_auth_method":null,"token_endpoint_auth_signing_alg":null,"initiate_login_uri":null},"action":{"kind":"continue_device_code_grant","id":"01KQZC5PW45MM6WCPW0DNFN5Y4"}}

Caused by:
    0: could not render template "pages/policy_violation.html"
    1: invalid operation: value is not a string (in pages/policy_violation.html:32)
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

(a device authorization grant doesn't have `grant.redirect_uri` either)

MadLittleMods: Fixed by having better fallbacks, https://github.com/element-hq/matrix-authentication-service/pull/5667/changes/9b64b69bb4fd4d21c10d9e75ecefb8f8ee523f78

frontend/src/styles/cpd-link.css
```diff
@@ -7,39 +7,44 @@
...
   border-radius: var(--cpd-radius-pill-effect);
   padding-inline: 0.25rem;
 }
 
-.cpd-link[data-kind="primary"] {
-  color: var(--cpd-color-text-primary);
-}
+/* Only underline and add hover/active styles if it actually links to something and
+there is something to interact with */
+.cpd-link[href] {
```
MadLittleMods: Nested all of the interactive styles under `.cpd-link[href]` so the link only looks interactive when it actually as an href attribute (links somewhere).

This also aligns with how the default browser user agent stylesheet works.

This change is spawning from needing to render the `templates/pages/policy_violation.html` template with a device authorization grant which has no URL to use.

templates/pages/policy_violation.html
```diff
@@ -29,7 +29,19 @@ <h1 class="title">{{ _("mas.policy_violation.heading") }}</h1>
...
+      {# Make it clear to the user which client they're dealing with. This template has
+      to deal with authorization grants and device authorization grants.
+      `client.client_uri` is the best but it's optional so we fallback to
+      `grant.redirect_uri` (required for authorization grants) which is where the user
+      will end up (probably a good approximation). But for device authorization grants,
+      there is no `redirect_uri` so we have to fallback all the way to an empty string.
+
+      We're using the `default('')` filter otherwise we see an `undefined value` error
+      when `grant.redirect_uri` isn't defined for device authorization grants #}
+      {% set client_href = (client.client_uri or grant.redirect_uri | default('') or '') | simplify_url %}
```
MadLittleMods: Since we're using the `default('')` filter, we could probably get away with dropping the last fallback but it's there for good measure (i.e. it should be there if we didn't have to use the default filter)

```suggestion
      {% set client_href = (client.client_uri or grant.redirect_uri | default('')) | simplify_url %}
```

templates/pages/policy_violation.html
```diff
@@ -29,7 +29,19 @@ <h1 class="title">{{ _("mas.policy_violation.heading") }}</h1>
...
+      <a target="_blank" {% if client_href %}href="{{ client_href }}"{% endif %} class="cpd-link"
```
MadLittleMods: Made the `href` attribute only render if we actually have a `href` to use.

A blank `href` links to itself (the same page) which we don't want.

translations/en.json
```diff
@@ -6,11 +6,11 @@
     },
     "cancel": "Cancel",
     "@cancel": {
-      "context": "pages/consent.html:79:11-29, pages/device_consent.html:148:13-31, pages/policy_violation.html:46:13-31"
+      "context": "pages/consent.html:83:11-29, pages/device_consent.html:153:13-31, pages/policy_violation.html:58:13-31"
     },
```
MadLittleMods, line 10: Automatic changes because the templates changed

frontend/src/styles/cpd-link.css
```diff
@@ -7,39 +7,44 @@
...
+  text-decoration: underline;
 
-.cpd-link[data-kind="critical"] {
-  color: var(--cpd-color-text-critical-primary);
-}
+  &[data-kind="primary"] {
```
MadLittleMods: I just copy pasted the styles inside and changed `.cpd-link` to `&` as it's nested inside `.cpd-link[href]` now

templates/pages/policy_violation.html
sandhose: feels wrong to use the `redirect_uri` as `href` in this case. It can be used as a client name fallback, but in case we don't have a `client_uri`, we should just display whatever name we *can* display but without it being a link?

templates/pages/policy_violation.html
sandhose: Feels like it should just not be a link if we don't have a `client_uri`

templates/pages/consent.html
```diff
@@ -29,6 +29,10 @@ <h1 class="title">
         {{ _('mas.consent.continue_to', client_name=client_name) }}
       </h1>
       <p class="text [&>span]:whitespace-nowrap [&>span]:text-[var(--cpd-color-text-link-external)]">
+        {# Make it clear to the user which client they're dealing with. This template
+        only has to deal with authorization grants. `client.client_uri` is the best but
+        it's optional so we fallback to `grant.redirect_uri` (required for authorization
+        grants) which is where the user will end up (probably a good approximation). #}
         {{ _("mas.consent.this_will_setup", client_name=client_name, client_uri=((client.client_uri or grant.redirect_uri) | simplify_url), server_name=branding.server_name) }}
```
sandhose: here the `client_uri | redirect_uri` is just being used for display, not as a link, that's why we do that

frontend/src/styles/cpd-link.css
sandhose, line 1: feels weird to handle ".cpd-link in case it's not a link", I don't think we should do that

sandhose (changes requested):
I really feel like this should be simplified into something like

```jinja
{% set client_name = client.client_name or ((client.client_uri or grant.redirect_uri) | simplify_url %}
{% if client.client_uri %}
  <a href="{{ client.client_uri }}">{{ client_name }}</a>
{% else %}
  {{ client_name }}
{% endif %}
```

The `client_uri` being absent is an edge case in case you're defining static clients in the config. In practice, Matrix clients have to set a `client_uri` when they register

frontend/src/styles/cpd-link.css
MadLittleMods: It's a bit weird. I assumed we wanted to keep the bold and spacing styles that are tied to [`.cpd-link`](https://github.com/element-hq/matrix-authentication-service/blob/31f00df71b786f6dfa1b82acab348822da281338/frontend/src/styles/cpd-link.css#L8-L15) but perhaps, that's just how normal links should look in the middle of a paragraph :shrug: 

What does `cpd` stand for? Compound?

templates/pages/policy_violation.html
MadLittleMods: I feel like using `redirect_uri` for display is just as weird as linking to it. My preference would be to only show/link `client_uri` in all places.

templates/pages/consent.html
MadLittleMods: Related discussion: https://github.com/element-hq/matrix-authentication-service/pull/5667#discussion_r3201605734

frontend/src/styles/cpd-link.css
MadLittleMods: (no longer relevant with the latest diff)

frontend/src/styles/cpd-link.css
MadLittleMods: Overall, the approach wasn't that crazy to me. CSS also has [`:link`](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Selectors/:link) and [`:any-link`](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Selectors/:any-link) pseudo classes to cover this kind of thing (which are probably better to use if we wanted to continue down this road).

But I've updated to conditionally use a link as you suggested: 

```jinja2
{% if client.client_uri %}
  <a target="_blank" href="{{ client.client_uri }}" class="cpd-link" data-kind="primary">
    {{ client_display_name }}
  </a>
{% else %}
  {{ client_display_name }}
{% endif %}
```

templates/pages/policy_violation.html
MadLittleMods: Updated with the following rules:

 - Only link with `client.client_uri`
 - Use `{% set client_display_name = client.client_name or client.client_id %}` for text
 - Use `{% set client_display_uri = ((client.client_uri or "") | simplify_url) or "<no client_uri provided>" %}` where we want to display the URI as text

templates/pages/policy_violation.html
MadLittleMods: Updated to conditionally render `<a>` vs text

templates/pages/policy_violation.html
MadLittleMods: (outdated)

frontend/src/styles/cpd-link.css
MadLittleMods: Now looks like this when there is no `client_uri` (no link):

sandhose (approved):

MadLittleMods:
Thanks for the review @sandhose 🦒

PR #5670, opened 2026-05-08
Add `max_session_threshold` configuration (only apply `session_limit` to people under the threshold)

MadLittleMods:
Add `max_session_threshold` configuration to prevent new users from passing session limits while not disrupting the people already past the limit.

> This is most applicable in scenarios where your homeserver has many legacy bots/scripts that login over and over (which ideally should be using [personal access tokens](https://github.com/element-hq/matrix-authentication-service/issues/4492)) and you want to avoid breaking their operation while maintaining some level of sanity with the number of devices that people can have. This will prevent anyone else from crossing the limit.

Spawning from https://github.com/element-hq/matrix-authentication-service/pull/5607#discussion_r3150765827

### Dev notes

See [`docs/development/contributing.md`](https://github.com/element-hq/matrix-authentication-service/blob/main/docs/development/contributing.md)

Testing:

```
DATABASE_URL='postgresql://postgres:postgres@localhost/mas' cargo nextest run -p mas-handlers compat::login::tests
```

```
DATABASE_URL='postgresql://postgres:postgres@localhost/mas' cargo nextest run --nocapture --retries 0 -p mas-handlers compat::login::tests
```

```
cd policies
make DOCKER=1 test
```

---

`TestState.reset()` -> `TestState.restart()`, see https://github.com/element-hq/matrix-authentication-service/pull/5670#discussion_r3222243619

---

OPA / rego:

 - https://www.openpolicyagent.org/docs/policy-reference/keywords/not

> When evaluating rule bodies, OPA searches for variable bindings that make all of the expressions true. There may be multiple sets of bindings that make the rule body true. The rule body can be understood intuitively as:
> ```
> expression-1 AND expression-2 AND ... AND expression-N
> ```
>
> *-- https://www.openpolicyagent.org/docs/policy-language#the-basics*

### Todo

 - [ ] ~~Add Rust tests for authorization grant side~~
     - It looks like there are no existing tests for the interactive login flows. And it's a lot more complicated than `m.login.password` so it's probably something we should handle in its own PR. For example, I started on a test that uses `m.login.sso` in https://github.com/element-hq/matrix-authentication-service/pull/5675 but OAuth 2.0 authorization grant is a step beyond that as well.
     - Discussion: https://github.com/element-hq/matrix-authentication-service/pull/5670#discussion_r3244078712


crates/handlers/src/compat/login.rs
```diff
@@ -1860,10 +1870,87 @@ mod tests {
...
+        for _ in 0..(upcoming_hard_limit + 1) {
+            let request = Request::post("/_matrix/client/v3/login")
+                .json(serde_json::json!(password_login_json.clone()));
+            let response = state.request(request.clone()).await;
+            response.assert_status(StatusCode::OK);
+        }
+
+        // Update the app state to configure a `max_session_threshold`
+        // FIXME: This creates an entirely new state
+        let state = TestState::from_pool_with_site_config(
```
MadLittleMods: What's the proper way to do this?

I basically want to use the same app (database, etc) but restart with a different config?

crates/config/src/sections/experimental.rs
```diff
@@ -162,6 +162,17 @@ pub struct SessionLimitConfig {
...
+    /// When set, only accounts with <= `max_session_threshold` sessions have the
+    /// session limits applied.
+    ///
+    /// This is most applicable in scenarios where your homeserver has many legacy
+    /// bots/scripts that login over and over (which ideally should be using [personal
+    /// access
+    /// tokens](https://github.com/element-hq/matrix-authentication-service/issues/4492))
+    /// and you want to avoid breaking their operation while maintaining some level of
+    /// sanity with the number of devices that people can have. This will prevent anyone
+    /// else from crossing the limit.
```
MadLittleMods: Similar to the language for `dangerous_hard_limit_eviction`

The new part is `This will prevent anyone else from crossing the limit.`

crates/handlers/src/compat/login.rs
MadLittleMods: As some rough notes, @sandhose suggested making the `state` mutable, change `state.site_config` inside, then call recreate (or something like that)

I think the method in mind was [`TestState.reset()`](https://github.com/element-hq/matrix-authentication-service/blob/9db7cf0aacd9ec50cab15a0f3f37dd7fdddcff71/crates/handlers/src/test_utils.rs#L302-L318) but it doesn't seem to behave as I would expect. Even the [single test](https://github.com/element-hq/matrix-authentication-service/blob/9db7cf0aacd9ec50cab15a0f3f37dd7fdddcff71/crates/handlers/src/compat/login.rs#L1272-L1273) it's used in seems to be flawed. The test looks for `Invalid username/password` but that's the same error message that's used for `RouteError::UserNotFound` which is what we're actually seeing instead of the password errors that are expected. The same database doesn't seem to be shared across the reset (or something important).

(all of the errors use the same message)
https://github.com/element-hq/matrix-authentication-service/blob/9db7cf0aacd9ec50cab15a0f3f37dd7fdddcff71/crates/handlers/src/compat/login.rs#L278-L282

It looks like `TestState.reset()` was introduced in the [same PR](https://github.com/element-hq/matrix-authentication-service/pull/4075) where the test usage came from and it was never caught.

And currently, it looks like calling `TestState::from_pool_with_site_config(...)` is basically the same as `TestState.reset()` anyway.

MadLittleMods: ### A fix (maintain mock homeserver state across MAS restarts)

@sandhose was able to spot the problem and I've created a separate PR for the fix, https://github.com/element-hq/matrix-authentication-service/pull/5678


### Disambiguating what we're running into

With our new test, this is what we're actually running into:

```
DATABASE_URL='postgresql://postgres:postgres@localhost' cargo nextest run --retries 0 -p mas-handlers compat::login::tests::test_session_limit_past_max_session_threshold
...
ERROR mas_handlers::compat::login: failed to provision device \x1b[3mmessage\x1b[0m\x1b[3m.sources\x1b[0m\x1b[2m=\x1b[0m[User not found]
test compat::login::tests::test_session_limit_past_max_session_threshold ... FAILED
```

`failed to provision device` is coming from `ProvisionDeviceFailed` here with the underlying `User not found` coming from [`MockHomeserverConnection::upsert_device(...)`](https://github.com/element-hq/matrix-authentication-service/blob/aecb172209d46d20409a80149ccb5c6b53b0a0cd/crates/matrix/src/mock.rs#L137-L148)

https://github.com/element-hq/matrix-authentication-service/blob/566d67b76f694f7767dd1683f73a04b462542a4c/crates/handlers/src/compat/login.rs#L458-L483

Which with hindsight, makes it even more obvious as a mock homeserver problem.


### Existing test 

To compare this with the [existing test](https://github.com/element-hq/matrix-authentication-service/blob/9db7cf0aacd9ec50cab15a0f3f37dd7fdddcff71/crates/handlers/src/compat/login.rs#L1272-L1273) using `TestState.reset()`, it's not the exact same problem as I thought. Although, the language is confusingly similar.

```
DATABASE_URL='postgresql://postgres:postgres@localhost' cargo nextest run --retries 0 -p mas-handlers compat::login::tests::test_user_password_login --success-output immediate
...
2026-05-13T19:20:17.740462Z  WARN mas_handlers::compat::login: user not found
2026-05-13T19:20:17.759225Z  WARN mas_handlers::compat::login: user not found
2026-05-13T19:20:17.780139Z  WARN mas_handlers::compat::login: user not found
```

The `user not found` is coming from `UserNotFound` [when the user is deactivated](https://github.com/element-hq/matrix-authentication-service/blob/566d67b76f694f7767dd1683f73a04b462542a4c/crates/handlers/src/compat/login.rs#L889-L895).

You can reproduce the same problem with that test if you just copy-paste the correct login request below the `TestState.reset()` like we're doing in the new test. Which is something I did when I found my test failing while seeing the other test pass. So the compounding of being able to reproduce while also seeing `user not found` with this test made me think it was also on the same track.

policies/authorization_grant/authorization_grant.rego
```diff
@@ -167,9 +167,32 @@ violation contains {
...
 	# sessions to return under the limit.
 	data.session_limit.soft_limit <= input.session_counts.total
 }
+
+# The session limits only apply to accounts within the `max_session_threshold`.
+#
+# True if the `max_session_threshold` isn't configured or <= `max_session_threshold`.
+passes_session_threshold if {
+	# If no `session_limit` configured, automatically passes (undefined)
+	not data.session_limit
```
MadLittleMods, line 185: Instead of dealing with `undefined`, is there a way to force a strict mode that all `data`/`input` used in the policy are defined? Solely relying on our Rust types like `mas_policy::CompatLoginInput` also sucks.

We could define everything manually like this at the top but keeping this manually up-to-date with the usage in the policy sucks.

```rego
allow if {
	# Specify what should be defined (even if null)
	data.session_limit
	data.session_limit.soft_limit
	data.session_limit.hard_limit
	input.session_replaced
	input.session_counts
    # etc

	count(violation) == 0
}
```

crates/handlers/src/compat/login.rs
```diff
@@ -1742,13 +1742,15 @@ mod tests {
...
     ///
     /// `soft_limit` is for when we allow the user to remove devices in
-    /// interactive contexts. With the compatibility login API, there is no
-    /// opportunity for us to present a web UI.
+    /// interactive contexts. If someone uses the `m.login.password`
+    /// compatibility login API, there is no opportunity for us to present a
+    /// web UI.
     #[sqlx::test(migrator = "mas_storage_pg::MIGRATOR")]
-    async fn test_soft_limit_does_not_affect_compat_login(pool: PgPool) {
+    async fn test_session_soft_limit_does_not_affect_non_interactive_login(pool: PgPool) {
```
MadLittleMods, line 2079: Added `session_` as a prefix to all of the session limit related tests so it's easier to see and reference them all.

crates/handlers/src/compat/login.rs
```diff
@@ -1824,31 +1830,89 @@ mod tests {
...
+                .expect("Expected errror response to include an `errcode`"),
+            "M_FORBIDDEN",
+            "Expected `errcode` to be `M_FORBIDDEN`"
+        );
+    }
+
+    /// Test that session limits are enforced for anyone who is *under* the
+    /// `max_session_threshold`
+    #[sqlx::test(migrator = "mas_storage_pg::MIGRATOR")]
+    async fn test_session_limit_under_max_session_threshold(pool: PgPool) {
```
MadLittleMods: New test

crates/handlers/src/compat/login.rs
```diff
@@ -1860,10 +1924,77 @@ mod tests {
         );
     }
 
+    /// Test that session limits are not enforced (logins are allowed) for
+    /// anyone who is already *past* the `max_session_threshold`
+    #[sqlx::test(migrator = "mas_storage_pg::MIGRATOR")]
+    async fn test_session_limit_past_max_session_threshold(pool: PgPool) {
```
MadLittleMods: New test

crates/handlers/src/compat/login.rs
```diff
@@ -1860,10 +1924,77 @@ mod tests {
...
+        session_limit_config.max_session_threshold.as_ref().expect("Expected `session_limit.max_session_threshold` to be configured at this point in the test");
+
+        // Since we're already above `max_session_threshold`, the `session_limit` won't
+        // stop us from adding another session.
+        let request = Request::post("/_matrix/client/v3/login")
+            .json(serde_json::json!(password_login_json.clone()));
+        let response = state.request(request.clone()).await;
+        response.assert_status(StatusCode::OK);
+    }
+
```
MadLittleMods, line 2457: Ideally, we'd also have some tests for interactive login flows like `m.login.sso` and OAuth 2.0 authorization grants but it looks like that isn't a thing in the codebase yet.

Interactive login is lot more complicated than `m.login.password` (non-interactive) so it's probably something we should handle in its own PR. For example, I started on a test that uses `m.login.sso` in https://github.com/element-hq/matrix-authentication-service/pull/5675 and OAuth 2.0 authorization grant is a step beyond that as well.

We should be testing more end-to-end like this.

policies/compat_login/compat_login_test.rego
```diff
@@ -21,6 +21,32 @@ need_to_remove_sessions(violations) := need if {
...
 # =========================================================================
+test_no_session_limiting_undefined if {
+	result := {
+		"allow": compat_login.allow,
+		"need_to_remove_sessions": need_to_remove_sessions(compat_login.violation),
+	} with input.user as user
+		with input.session_counts as {"total": 1}
+		with input.login as {"type": "m.login.sso"}
+		with input.session_replaced as false
+		with input._noop as null # The main thing we're trying to test (`data.session_limit` as undefined)
```
MadLittleMods, line 32: Using this noop syntax as otherwise `make DOCKER=1 fmt` forces the comment to unindent and be placed on the expressions below. We can't even use comments in between these `with ...` expressions either (annoying formatter)

crates/handlers/src/compat/login.rs
```diff
@@ -1824,31 +1830,89 @@ mod tests {
...
 
         let _user = user_with_password(&state, "alice", "password", false).await;
+        let password_login_json = serde_json::json!({
+            "type": "m.login.password",
+            "identifier": {
+                "type": "m.id.user",
+                "user": "alice",
+            },
+            "password": "password",
+        });
```
MadLittleMods, line 2166: Just updating the test to share this `password_login_json` across the multiple places we make requests.

It could be more test-wide constant but 🤷 reads fine as-is and the user is created right here which is what it pertains to. I think the duplication is best.

crates/handlers/src/compat/login.rs
```diff
@@ -1824,31 +1830,89 @@ mod tests {
...
 
         // Keep logging in to add more sessions, up to the `hard_limit`
-        #[allow(clippy::range_plus_one)]
```
MadLittleMods, line 2154: Annoying that `clippy` doesn't call out unused ignores

sandhose: Using `#[expect(…lint…)]` instead of `#[allow(…lint…)]` will do exactly that

crates/handlers/src/compat/login.rs
sandhose: In general interactive testing is a big thing we're missing. Ideally we would do some browser-based integration testing for some of those based on playwright.

The Tchap team sent me a bunch of E2E tests they wrote: https://github.com/tchapgouv/tchap-e2e-playwright

policies/authorization_grant/authorization_grant.rego
sandhose: I wanted to point out that there's a JSON schema derived… but that's only for the input, not the data… :(

Part of the complexity is dealing with `null` honestly, because else if you don't care about `null`s, your `else if` chain becomes only `not data.session_limit.max_session_threshold`

sandhose: I would honestly be fine with just ignoring `null` values and expect that people don't set things to null

sandhose (approved):

crates/handlers/src/compat/login.rs
MadLittleMods: hmmm, it looks like there is a rule to suggest this since `1.70.0`: [`clippy::allow_attributes`](https://rust-lang.github.io/rust-clippy/master/index.html#allow_attributes) but we don't see anything because it's allowed by default. I've created https://github.com/element-hq/matrix-authentication-service/pull/5691 to enforce this lint :fast_forward:

policies/authorization_grant/authorization_grant.rego
MadLittleMods: `null` is an expected value for `None` config in Rust.

The `undefined` case is the tricky one that I wish we could strictly enforce didn't happen.

crates/handlers/src/compat/login.rs
MadLittleMods: Since https://github.com/element-hq/matrix-authentication-service/pull/5675 was merged, I added some tests for interactive login with `m.login.sso`. Still a testing coverage hole with the OAuth 2.0 authorization grant side though.

Having some even better end-to-end tests with an actual browser sounds nice (preferably driven through the Rust side though). Or if we expect MAS to support more than just Synapse, probably with Complement tests with browsers driven on the Go side.

MadLittleMods:
Thanks for the review @sandhose 🦭

PR #5672, opened 2026-05-11
Update Authelia Sample for OIDC upstream

andrepintorj:
Adding Authelia configuration fix so that preferred_username is populated accordingly. Looks like the same issue happens in Grafana.
More info at [Grafana's Authelia Documentation](https://www.authelia.com/integration/openid-connect/clients/grafana/#configuration-escape-hatch)


reivilibre (approved):
Thanks!

PR #5677, opened 2026-05-13
Allow passing arbitrary values from clients down to upstream OAuth providers

sandhose:

This allows arbitrary parameters passed by clients down to the upstream OAuth provider through Jinja templates.

This extends the existing `additional_authorization_parameters` to use Jinja templates:

```yaml
upstream_oauth2:
  providers:
    - id: …
      additional_authorization_parameters:
        # Static value
        kc_idp_hint: "saml"
        # Forward the downstream `login_hint`, if any
        login_hint: "{{ params.login_hint }}"
```


crates/handlers/src/oauth2/authorization/mod.rs
```diff
@@ -4,10 +4,13 @@
 // SPDX-License-Identifier: AGPL-3.0-only OR LicenseRef-Element-Commercial
 // Please see LICENSE files in the repository root for full details.
 
+use std::collections::BTreeMap;
+
 use axum::{
-    extract::{Form, State},
```
sandhose, line 8: Note that the `Form` extractor behaves like the `Query` extractor on `GET` requests, hence switching to it

crates/handlers/src/oauth2/authorization/mod.rs
```diff
@@ -118,7 +121,10 @@ pub(crate) async fn get(
     activity_tracker: BoundActivityTracker,
     mut repo: BoxRepository,
     cookie_jar: CookieJar,
-    Form(params): Form<Params>,
+    // Extract the query parameters twice: once to get the raw query string,
+    // and once to parse it into a structured `Params` object.
+    Query(raw_parameters): Query<BTreeMap<String, String>>,
+    Query(params): Query<Params>,
```
reivilibre, line 127: do we actually want the fields in `Params` to be also included in `raw_parameters`?

docs/setup/sso.md
```diff
@@ -67,6 +67,38 @@ The template has the following variables available:
...
+
+Templates that render to an empty string are dropped rather than forwarded, so referencing a downstream parameter that was not supplied does not produce a stray empty query parameter.
+
+Plain strings without `{{ … }}` are valid templates that render to themselves, so static values continue to work without changes.
+
+```yaml
+upstream_oauth2:
+  providers:
+    - id: …
+      additional_authorization_parameters:
```
reivilibre, line 88: do we need to do something about the backwards compatibility hazard that now we interpret these as templates whereas they previously were literals?

crates/handlers/src/upstream_oauth2/authorize.rs
```diff
@@ -147,3 +176,114 @@ pub(crate) async fn get(
...
+    provider_id: Ulid,
+    templates: &'a [(String, String)],
+    raw_parameters: &BTreeMap<String, String>,
+) -> impl Iterator<Item = (&'a str, String)> {
+    let env = environment();
+    let ctx = context! { params => raw_parameters };
+
+    templates.iter().filter_map(move |(key, template)| {
+        match env.render_str(template, &ctx) {
+            Ok(value) if !value.trim().is_empty() => Some((key.as_str(), value)),
```
reivilibre: hmm, this feels a bit weird to both trim and yet forward the untrimmed version. Not sure what's correct

reivilibre (approved):

crates/handlers/src/oauth2/authorization/mod.rs
sandhose: I'd say yes, else we'll have to document which parameter is included and which one isn't

docs/setup/sso.md
sandhose: I don't think anyone would have Jinja-like syntax in there in existing deployments, probably not worth handling

crates/handlers/src/upstream_oauth2/authorize.rs
sandhose: Fixed in 772360bb4efa

sandhose:
ESS test is flaky, force-merging

PR #5679, opened 2026-05-13
Schedule `SyncDevicesJob` after `dangerous_hard_limit_eviction` so the homeserver stays up to date

MadLittleMods:
Schedule `SyncDevicesJob` after `dangerous_hard_limit_eviction` so the homeserver stays up to date

Follow-up to https://github.com/element-hq/matrix-authentication-service/pull/5607#discussion_r3189451555

### DIFF
--- a/crates/handlers/src/compat/login.rs
+++ b/crates/handlers/src/compat/login.rs
@@ -458,13 +458,14 @@ pub(crate) async fn post(
     // Now we can create the device on the homeserver, without holding the
     // transaction
     //
-    // Normally, devices get synced to the homeserver in a `SyncDevicesJob` but we want
-    // the device to be created synchronously on the homeserver, so that when we
-    // respond, the device already exists. If the device doesn't exist on Synapse, token
-    // introspection from Synapse to MAS will work, but then the device won't exist, so
-    // Synapse will return a 401. We're using an upsert so if the device already exists
-    // for some reason (like when we're replacing it, or a concurrent device sync
-    // happening) it won't have any effect.
+    // Normally, devices get synced to the homeserver in a `SyncDevicesJob` but we
+    // want the device to be created synchronously on the homeserver, so that
+    // when we respond, the device already exists. If the device doesn't exist
+    // on Synapse, token introspection from Synapse to MAS will work, but then
+    // the device won't exist, so Synapse will return a 401. We're using an
+    // upsert so if the device already exists for some reason (like when we're
+    // replacing it, or a concurrent device sync happening) it won't have any
+    // effect.
     if let Err(err) = homeserver
         .upsert_device(
             &user.username,
--- a/crates/handlers/src/compat/login.rs
+++ b/crates/handlers/src/compat/login.rs
@@ -460,12 +460,14 @@ pub(crate) async fn post(
     //
     // Normally, devices get synced to the homeserver in a `SyncDevicesJob` but we
     // want the device to be created synchronously on the homeserver, so that
-    // when we respond, the device already exists. If the device doesn't exist
-    // on Synapse, token introspection from Synapse to MAS will work, but then
-    // the device won't exist, so Synapse will return a 401. We're using an
-    // upsert so if the device already exists for some reason (like when we're
-    // replacing it, or a concurrent device sync happening) it won't have any
-    // effect.
+    // when we respond, the access token works completely. If the device doesn't
+    // exist on the homeserver side, token introspection from Synapse to MAS
+    // will work but Synapse will return a 401 because it doesn't see the
+    // device.
+    //
+    // We're using an upsert so if the device already exists for some reason (like
+    // when we're replacing it, or a concurrent device sync happening) it won't
+    // have any effect.
     if let Err(err) = homeserver
         .upsert_device(
             &user.username,
--- a/crates/handlers/src/compat/login.rs
+++ b/crates/handlers/src/compat/login.rs
@@ -457,6 +457,12 @@ pub(crate) async fn post(
 
     // Now we can create the device on the homeserver, without holding the
     // transaction
+    //
+    // Normally, devices get synced to the homeserver in a `SyncDevicesJob` but we want
+    // the device to be created synchronously on the homeserver, so that when we
+    // respond, the device already exists (TODO: Why important?). We're using an upsert
+    // so if the device already exists for some reason (like when we're replacing it, or
+    // a concurrent device sync happening) it won't have any effect.
     if let Err(err) = homeserver
         .upsert_device(
             &user.username,
--- a/crates/handlers/src/compat/login.rs
+++ b/crates/handlers/src/compat/login.rs
@@ -508,6 +508,7 @@ pub(crate) async fn post(
 /// Given the violations from [`Policy::evaluate_compat_login`], return the
 /// appropriate `RouteError` response.
 async fn process_violations_for_compat_login(
+    rng: &mut (dyn RngCore + Send),
     clock: &dyn Clock,
     repo: &mut BoxRepository,
     session_limit_config: Option<&SessionLimitConfig>,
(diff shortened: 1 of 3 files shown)


crates/handlers/src/compat/login.rs
```diff
@@ -596,6 +603,11 @@ async fn process_violations_for_compat_login(
                                 .finish(clock, compat_session.to_owned())
                                 .await?;
                         }
+
+                        // Schedule a device sync with the homeserver
+                        repo.queue_job()
+                            .schedule_job(rng, clock, SyncDevicesJob::new_for_id(user.id))
+                            .await?;
```
MadLittleMods, line 615: Is it necessary to call `repo.save().await?;` here as well?

For example we do that here:

https://github.com/element-hq/matrix-authentication-service/blob/aecb172209d46d20409a80149ccb5c6b53b0a0cd/crates/handlers/src/compat/login.rs#L468-L480

If so, why can't we rely on the behavior of transaction being created for each request and committed? Where does that happen anyway? Perhaps that assumption is wrong and it's just wherever we create a repo and save a repo. And so the `repo.save().await?;` already on the outside covers us (I think this is the correct way to think about it):

https://github.com/element-hq/matrix-authentication-service/blob/aecb172209d46d20409a80149ccb5c6b53b0a0cd/crates/handlers/src/compat/login.rs#L446

What's the litmus test and mental model I should be using?



---

Also if necessary, how should I be managing the `repo` as `repo.save()` doesn't like being a reference. Pass around the `repository_factory`? I think this question is probably irrelevant though if my assumption above is correct.

<details>
<summary><code>error[E0507]: cannot move out of `*repo` which is behind a mutable reference</code></summary>

```
error[E0507]: cannot move out of `*repo` which is behind a mutable reference
   --> crates/handlers/src/compat/login.rs:612:25
    |
612 |                         repo.save().await?;
    |                         ^^^^ ------ `*repo` moved due to this method call
    |                         |
    |                         move occurs because `*repo` has type `Box<dyn Repository<RepositoryError> + Send + Sync>`, which does not implement the `Copy` trait
    |
note: `mas_storage::RepositoryTransaction::save` takes ownership of the receiver `self`, which moves `*repo`
   --> /home/eric/Documents/github/element/matrix-authentication-service/crates/storage/src/repository.rs:89:13
    |
 89 |     fn save(self: Box<Self>) -> BoxFuture<'static, Result<(), Self::Error>>;
    |             ^^^^
    = note: the full name for the type has been written to '/tmp/rust-analyzer-check/debug/deps/mas_handlers-b11b22e52e346042.long-type-3978556458229716060.txt'
    = note: consider using `--verbose` to print the full type name to the console
help: you could `clone` the value and consume it, if the following trait bounds could be satisfied: `dyn mas_storage::Repository<mas_storage::RepositoryError> + std::marker::Send + std::marker::Sync: std::marker::Sized` and `dyn mas_storage::Repository<mas_storage::RepositoryError> + std::marker::Send + std::marker::Sync: std::clone::Clone`
    |
612 |                         <std::boxed::Box<dyn mas_storage::Repository<mas_storage::RepositoryError> + std::marker::Send + std::marker::Sync> as Clone>::clone(&repo).save().await?;
    |                         ++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++    +
```

</details>

crates/handlers/src/compat/logout.rs
```diff
@@ -117,13 +117,19 @@ pub(crate) async fn post(
...
 
+    // This will make the access token invalid
+    repo.compat_session().finish(&clock, session).await?;
+
     // Schedule a job to sync the devices of the user with the homeserver
+    //
+    // Doing this in a background job is ok as the access token will be invalid right
+    // away (from the session being finished above) and we do actually want to do a full
+    // device list sync, because we're not sure whether we want to delete the device (if
+    // there is for example a concurrent logout and login with the same device ID).
```
MadLittleMods: This reasoning comes from @sandhose's explanation in https://github.com/element-hq/matrix-authentication-service/pull/5607#discussion_r3232971115

crates/handlers/src/compat/logout.rs
```diff
@@ -117,13 +117,19 @@ pub(crate) async fn post(
         // XXX: this is probably not the right error
         .ok_or(RouteError::InvalidAuthorization)?;
 
+    // This will make the access token invalid
+    repo.compat_session().finish(&clock, session).await?;
```
MadLittleMods, line 121: As @sandhose explained in https://github.com/element-hq/matrix-authentication-service/pull/5607#discussion_r3232971115, the ordering doesn't matter (because both things are committed in a single transaction) but I've updated this just so it reads more plainly and obviously.

crates/handlers/src/compat/login.rs
```diff
@@ -457,6 +457,13 @@ pub(crate) async fn post(
 
     // Now we can create the device on the homeserver, without holding the
     // transaction
+    //
+    // Normally, devices get synced to the homeserver in a `SyncDevicesJob` but we
+    // want the device to be created synchronously on the homeserver, so that
+    // when we respond, the device already exists (TODO: Why important?). We're
+    // using an upsert so if the device already exists for some reason (like
+    // when we're replacing it, or a concurrent device sync happening) it won't
+    // have any effect.
```
MadLittleMods, line 470: This reasoning comes from @sandhose's explanation in https://github.com/element-hq/matrix-authentication-service/pull/5607#discussion_r3232971115

Overall, this makes sense to me but just so I'm not guessing/assuming, do you know the exact answer for `TODO: Why important?`

crates/handlers/src/compat/logout.rs
```diff
@@ -117,13 +117,21 @@ pub(crate) async fn post(
...
+
     // Schedule a job to sync the devices of the user with the homeserver
+    //
+    // Doing this in a background job is ok as the access token will be invalid
+    // right away (from the session being finished above) and we do actually
+    // want to do a full device list sync (as opposed to
+    // `homeserver.delete_device(...)`), because we're not sure whether we want
+    // to delete the device (if there is for example a concurrent logout and
+    // login with the same device ID).
```
MadLittleMods, line 130: This reasoning comes from @sandhose's explanation in https://github.com/element-hq/matrix-authentication-service/pull/5607#discussion_r3232971115

crates/handlers/src/compat/login.rs
sandhose: It is important because when we return to the client with an access token, the access token *has* to work. If the device doesn't exist on Synapse, token introspection from Synapse to MAS will work, but then the device won't exist, so Synapse will return a 401

crates/handlers/src/compat/login.rs
sandhose: By design, the `repo` holds an open transaction, and all the operations on it need `&mut self` access.
`repo.save()` and `repo.cancel()` "consume" the repo (as they take `self` access) to save/rollback the transaction, and make sure you don't do anything with it once you've committed or rolled back the transaction.

Dropping the repo will also rollback the transaction (but asynchronously)

The repository factory is there for cases where in a single request we need multiple transactions. This is useful whenever we're talking to external services that may take some time to respond; we usually don't want to hold a running transaction whilst we're waiting for Synapse to respond for example, because we risk exhausting the database pool quite quickly if we do that too much.

In your particular case, your branch returns `Ok(())`, which means it will continue and usually hit `repo.save()`

sandhose (approved):

crates/handlers/src/compat/login.rs
MadLittleMods: Sounds good, works as I expected. Versus the off-hand generalization that each request gets a transaction (magically) that I've heard in passing.

crates/handlers/src/compat/login.rs
MadLittleMods: Thanks @sandhose! Added to the comment here :white_check_mark:

MadLittleMods:
Thanks for the review @sandhose 🐦

PR #5681, opened 2026-05-14
Check whether policy evaluation result is `valid()` when we `process_violations_for_compat_login(...)`

MadLittleMods:
Check whether policy evaluation result is `valid()` when we `process_violations_for_compat_login(...)`

This regressed in https://github.com/element-hq/matrix-authentication-service/pull/5607 but because the policy is only being used for session limiting (https://github.com/element-hq/matrix-authentication-service/pull/5287), no behavior has actually regressed. This is just important if we start enforcing other logic in these policies.

### DIFF
--- a/crates/handlers/src/compat/login.rs
+++ b/crates/handlers/src/compat/login.rs
@@ -499,8 +499,9 @@ pub(crate) async fn post(
     }))
 }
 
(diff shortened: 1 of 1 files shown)


sandhose (approved):

MadLittleMods:
Thanks for the review @sandhose 🐫

PR #5682, opened 2026-05-15
Add authenticated-only GraphQL introspection option

### LINKED ISSUE #5635: Add config option to disable GraphQL introspection for unauthenticated requests
## Feature request

MAS exposes a GraphQL endpoint at `/graphql` which is used by the self-service account UI.

This endpoint allows unauthenticated clients to run introspection queries:

```bash
curl -X POST https://auth.matrix.example.com/graphql -H "Content-Type: application/json" -d '{"query":"{ __schema { types { name } } }"}'
```

This returns the full schema (129 types) without any authentication. While actual data queries correctly return `viewer: Anonymous` for unauthenticated requests, the schema itself is fully discoverable.

## Request

Add a configuration option to disable GraphQL introspection for unauthenticated (anonymous) requests, similar to how `playground: false` disables the playground UI.

For example:

```yaml
http:
  listeners:
    - name: graphql
      playground: false
      introspection: authenticated_only  # or: false / public / authenticated_only
```

##  Context

- The schema is already publicly documented, so this is a defence-in-depth measure
- The playground option already exists as precedent for this kind of configuration
- The self-service UI does not require introspection for anonymous users; it only needs it after the user has authenticated via session cookie
- Operators deploying MAS in regulated environments (healthcare, government) may have privacy requirements that prohibit exposing API schemas without authentication

ppmaskaluk:
## Summary
- add an `http.listeners[].resources[].introspection` GraphQL setting with `public` and `authenticated_only` modes
- block anonymous GraphQL schema introspection while still allowing authenticated browser sessions and normal anonymous queries
- add config/runtime tests and document the new listener option

Closes #5635.


sandhose:
Hey 👋 

Thanks for this contribution, although as commented on the issue, I'm not sure this is really useful? As in, the schema is anyway public, so what is this trying to achieve?

sandhose (changes requested):
I don't believe this is needed as the GraphQL schema is already public? Could you explain why you want this?

sandhose:
Closing because of the lack of response from the author

PR #5691, opened 2026-05-20
Enforce `clippy::allow_attributes` (prefer `#[expect]` over `#[allow]`)

MadLittleMods:
Enforce [`clippy::allow_attributes`](https://rust-lang.github.io/rust-clippy/master/index.html#allow_attributes)  which prefers `#[expect(...)]` over `#[allow(...)]` when you want to make a linter exception.

`#[expect(unused_mut)]` is better because it will complain about the lint being unused if there is not `mut` usage anymore.

> Example
> ```rust
> #[allow(unused_mut)]
> fn foo() -> usize {
>     let mut a = Vec::new();
>     a.len()
> }
> ```
> 
> Use instead:
> ```rust
> #[expect(unused_mut)]
> fn foo() -> usize {
>     let mut a = Vec::new();
>     a.len()
> }
> ```

Spawning from https://github.com/element-hq/matrix-authentication-service/pull/5670#discussion_r3244177988


Cargo.toml
```diff
@@ -32,6 +32,9 @@ too_many_lines = "allow"
 # Sometimes variables are long and annoying to inline in the format string.
 # And this lint also complains about aliases, https://github.com/rust-lang/rust-clippy/issues/10247
 uninlined_format_args = "allow"
+# Prefer `#[expect(unused_mut)]` over `#[allow(unused_mut)]` because it will complain if
+# the lint ignore is not used.
+allow_attributes = "deny"
```
MadLittleMods, line 37: By default, this lint defaults to [`allow`](https://rust-lang.github.io/rust-clippy/master/index.html#allow_attributes) (see the green `allow` in the top-right of the lint section)

crates/tower/src/metrics/make_attributes.rs
```diff
@@ -136,7 +136,11 @@ macro_rules! impl_for_tuple {
             $($rest: MetricsAttributes<T> + 'static,)*
         {
             type Iter<'a> = chain_for!($first $(, $rest)*);
+            // Lint-ignore for the `#[allow(...)]` usage below
+            #[expect(clippy::allow_attributes)]
             fn attributes<'a>(&'a self, t: &'a T) -> Self::Iter<'a> {
+                // This has to be a an `#[allow(...)]` as it doesn't get fulfilled here but is
+                // relevant when the macro expansion occurs.
```
MadLittleMods: Here is the only weird exception we had to make

crates/storage-pg/src/oauth2/authorization_grant.rs
```diff
@@ -35,7 +35,6 @@ impl<'c> PgOAuth2AuthorizationGrantRepository<'c> {
     }
 }
 
-#[allow(clippy::struct_excessive_bools)]
```
MadLittleMods, line 38: We were able to remove a good number of linter exceptions.

I first updated everything from `#[allow(...)]` to `#[expect(...)]`, then ran the linter again and removed any lint expectations that were unfulfilled.

crates/i18n/src/sprintf/mod.rs
```diff
@@ -63,16 +63,16 @@ macro_rules! sprintf {
...
+#[expect(unused_imports)]
 pub(crate) use arg_list_inner;
-#[allow(unused_imports)]
+#[expect(unused_imports)]
 pub(crate) use sprintf;
 
 #[derive(Debug, thiserror::Error)]
 #[error(transparent)]
-#[allow(dead_code)]
+#[expect(dead_code)]
```
MadLittleMods: All of these appear correct (`cargo clippy --workspace` doesn't complain) but my code editor (VSCode) complains about these being unfulfilled.

```
warning: this lint expectation is unfulfilled
  --> crates/i18n/src/sprintf/mod.rs:66:10
   |
66 | #[expect(unused_imports)]
   |          ^^^^^^^^^^^^^^
   |
   = note: `#[warn(unfulfilled_lint_expectations)]` on by default
```

Although, it looks like CI complains:

:x: https://github.com/element-hq/matrix-authentication-service/actions/runs/26254998386/job/77275564475?pr=5691
```
error: this lint expectation is unfulfilled
  --> crates/i18n/src/sprintf/mod.rs:66:10
   |
66 | #[expect(unused_imports)]
   |          ^^^^^^^^^^^^^^
   |
   = note: `-D unfulfilled-lint-expectations` implied by `-D warnings`
   = help: to override `-D warnings` add `#[allow(unfulfilled_lint_expectations)]`

error: this lint expectation is unfulfilled
  --> crates/i18n/src/sprintf/mod.rs:68:10
   |
68 | #[expect(unused_imports)]
   |          ^^^^^^^^^^^^^^

error: this lint expectation is unfulfilled
  --> crates/i18n/src/sprintf/mod.rs:75:10
   |
75 | #[expect(dead_code)]
   |          ^^^^^^^^^
```

MadLittleMods: It looks like this probably has to do with `arg_list!(...)` and friends only being used in tests. I've updated them to be gated behind `#[cfg(test)]` alongside the tests.

Separately, it seems like `arg_list!(...)` isn't really worth the complexity. Perhaps I'm missing something.

crates/i18n/src/sprintf/mod.rs
```diff
@@ -63,21 +63,20 @@ macro_rules! sprintf {
...
 #[derive(Debug, thiserror::Error)]
 #[error(transparent)]
-#[allow(dead_code)]
+#[cfg(test)]
 enum Error {
     Format(#[from] self::formatter::FormatError),
     Parse(Box<self::parser::Error>),
 }
 
+#[cfg(test)]
```
MadLittleMods, line 79: See https://github.com/element-hq/matrix-authentication-service/pull/5691#discussion_r3284557046 for context behind these changes

sandhose (approved):

MadLittleMods:
Thanks for the review @sandhose 🐛

PR #5704, opened 2026-05-26
Disable the device code auto-filling by default and match the device authorization grant with the designs

sandhose:
This should be reviewed commit by commit.

This does essentially two things:

 - make the 'auto-fill' behaviour of the device code grant configurable, defaulting to false
 - change the device code grant flow to match the latest designs

There are some drive-by changes, like fixing the page heading text size to match the design across the app.


frontend/src/styles/cpd-alert.css
sandhose, line 1: This is copied (with a `cpd-` prefix on classes) from Compound Web directly

reivilibre (approved):

PR #5707, opened 2026-05-29
Admin API: filter users based on client usage

sandhose:
This adds a filter in the admin API on users to see users that have active compat sessions as well as sessions with a particular client

### DIFF
--- a/crates/storage-pg/src/user/mod.rs
+++ b/crates/storage-pg/src/user/mod.rs
@@ -163,6 +163,23 @@ impl Filter for UserFilter<'_> {
                     )
                 },
             ))
+            .add_option(self.has_active_oauth2_session().map(|has| -> SimpleExpr {
+                let exists = Expr::exists(
+                    Query::select()
+                        .expr(Expr::cust("1"))
+                        .from(OAuth2Sessions::Table)
+                        .and_where(
+                            Expr::col((OAuth2Sessions::Table, OAuth2Sessions::UserId))
+                                .equals((Users::Table, Users::UserId)),
+                        )
+                        .and_where(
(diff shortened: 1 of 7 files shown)


crates/storage-pg/src/user/mod.rs
```diff
@@ -132,6 +132,71 @@ impl Filter for UserFilter<'_> {
             .add_option(self.search().map(|search| {
                 Expr::col((Users::Table, Users::Username)).ilike(format!("%{search}%"))
             }))
+            .add_option(self.active_oauth2_session_for_any_of_clients().map(
+                |clients| -> SimpleExpr {
+                    if clients.is_empty() {
+                        // "Has an active OAuth2 session for any of these
+                        // zero clients" matches no row.
+                        return Expr::val(false).into();
```
reivilibre: slightly surprised we need this special case logic here; would expect Postgres to be smart enough and since it's not a sane use case anyway, do we benefit having code for it?

reivilibre (approved):

crates/storage-pg/src/user/mod.rs
sandhose: You're right, I don't think this optimization makes sense. Tried the following:

```
auth=# select count(*) from users where user_id = ANY(ARRAY[]::uuid[]);
 count
-------
     0
(1 row)

auth=# select count(*) from users;
 count
-------
     1
(1 row)
```

sandhose: Fixed in [d951d0e](https://github.com/element-hq/matrix-authentication-service/pull/5707/commits/d951d0eed50cbab9f2a47dfd9c057bd10a9e77fc) and there is a test covering this case

PR #5709, opened 2026-05-29
Admin API: allow filtering of sessions by creation and last-activity time

sandhose:
This allows all session types to be filtered by creation time and 'last active at' time in the admin API.

Changes to the storage layer and to the API are split; it can be reviewed commit by commit


crates/storage-pg/src/ulid_at.rs
```diff
@@ -0,0 +1,80 @@
...
+//! without adding a separate `created_at` index.
+
+use chrono::{DateTime, Utc};
+use ulid::Ulid;
+use uuid::Uuid;
+
+/// Build the smallest possible [`Uuid`] for a ULID generated at the given time.
+///
+/// Use this as the upper bound when filtering rows created strictly *before*
+/// `t`: `id < min_ulid_at(t)`.
```
reivilibre, line 23: checked and Postgres orders by bytes, which is what we want indeed

reivilibre (approved):

PR #5710, opened 2026-05-29
Admin API: allow listing and viewing of OAuth 2.0 clients

sandhose:
This adds an admin API to list and get OAuth 2.0 clients, with a few filters:

 - if the client is dynamic or statically registered
 - search on the client_uri
 - search on the client_name
 - whether it supports a particular grant_type
 - whether it has active sessions or not

I haven't throughly verified that all the queries are efficient or not, but I'm not too concerned as this is an admin API and therefore won't really be a DoS vector. I prefer getting this deployed in real-world scenarios and see there how those filters behave before adding a bunch of indexes

### DIFF
--- a/crates/storage/src/oauth2/client.rs
+++ b/crates/storage/src/oauth2/client.rs
@@ -22,10 +22,10 @@ use crate::{Page, Pagination, repository_impl};
 #[derive(Clone, Copy, Debug, PartialEq, Eq, Default)]
 pub struct OAuth2ClientFilter<'a> {
     state: Option<bool>,
-    _lifetime: std::marker::PhantomData<&'a ()>,
+    client_name: Option<&'a str>,
 }
 
-impl OAuth2ClientFilter<'_> {
+impl<'a> OAuth2ClientFilter<'a> {
     /// Create a new [`OAuth2ClientFilter`] with default values
     #[must_use]
     pub fn new() -> Self {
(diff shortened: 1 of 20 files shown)


crates/storage/src/oauth2/client.rs
```diff
@@ -15,7 +16,113 @@ use rand_core::RngCore;
 use ulid::Ulid;
 use url::Url;
 
-use crate::repository_impl;
+use crate::{Page, Pagination, repository_impl};
+
+/// Filter parameters for listing OAuth 2.0 clients
+#[derive(Clone, Copy, Debug, PartialEq, Eq, Default)]
+pub struct OAuth2ClientFilter<'a> {
+    state: Option<bool>,
```
reivilibre: needs a docstring given the true/false isn't intrinsically meaningful

reivilibre (approved):
sgtm otherwise

crates/storage/src/oauth2/client.rs
sandhose: right, should probably be an enum instead. The meaningful docstrings are really on the `only_static_clients`, `only_dynamic_clients` and `state` methods. Fixing that to use an enum instead

sandhose: done in [256c5e3](https://github.com/element-hq/matrix-authentication-service/pull/5710/commits/256c5e3ff5628d6332641fa91a4c50021fdaa682)

PR #5722, opened 2026-06-03
feat(OAuth): delete all upstream links for a deactivated user

### LINKED ISSUE #5436: Allow upstream IdP identity to be reused when previously linked user is deleted
The current behaviour observed is:

- Register a user `@user1:homeserver.local` on MAS using an upstream IdP (e.g. GitHub)
- Deactivate/delete `@user1:homeserver.local`
- Attempt to register a new user `@user2:homeserver.local` using the same upstream IdP
- Instead of succeeding, you get an error `This account (@user1:homeserver.local) has been deleted. If this is not expected, contact your server administrator.`
- This indicates that the upstream IdP identity is still linked to the deleted user.

Allow an upstream IdP identity to be reused where the previously linked user has been deleted.

None.

**Additional context**
None.

defaultdino:
potentially closes #5436


sandhose (commented):
I think overall this would be a good idea, although I've got two reasons in mind to push back.

First, having the link still around makes it possible for us to show an informative fallback page to the user. "Your account has been deactivate, please contact your administrator" or something like that.

Second, this wouldn't cleanup any existing link. Ideally, we would retroactively clean up the links of existing deactivated users when doing this.

This is making me think that we should probably introduce a two-phase cleanup: deactivating the account does as currently, but after 30 days of the account being deactivated, we clean up the rest, including upstream OAuth links and password hashes.

defaultdino:
@sandhose I implemented the functionality you described in a new runnable job `CleanupDeactivatedUsersJob`

PR #5735, opened 2026-06-16
Record and log the requester on the request log line

sandhose:
Adds a `Requester` to the `LogContext` and records it in various places, so each request log line can be attributed to who made it.

There are 3 kinds of requesters being recorded:

 - a `user:<id>(<username>)` if the request was made by an end-user directly. Note that we try to include the `username` if we have it, but it's not always the case
 - an `oauth2-client:<id>` if the request was made by an OAuth 2.0 client. This includes request to the token endpoint and the introspection endpoint (which might be a bit weird?)
 - the `homeserver` if the request was made by the homeserver (on the introspection endpoint only)

Those requesters are visible on the request log line as `requester=…`

Can be reviewed commit by commit.

Part of a series of request-logging improvements. See the documentation PR to understand the "big picture" of this series of PRs: https://github.com/element-hq/matrix-authentication-service/pull/5741


crates/axum-utils/src/log_context.rs
```diff
@@ -0,0 +1,61 @@
...
+//! [`User`], [`BrowserSession`], [`Session`], [`CompatSession`] or [`Client`]
+//! instead of reaching for [`LogContext::maybe_set_requester`] + [`Requester`]
+//! directly. Each call site should record the entity that *authenticated the
+//! request* exactly once; [`LogContext`] is first-writer-wins, so an accidental
+//! second call is harmless.
+
+use mas_context::{LogContext, Requester};
+use mas_data_model::{BrowserSession, Client, CompatSession, Session, User};
+
+/// Record `self` as the [`Requester`] on the current [`LogContext`], if any.
```
reivilibre: doc should be on function not on trait

reivilibre (approved):

crates/axum-utils/src/log_context.rs
sandhose: done in [fc798b2](https://github.com/element-hq/matrix-authentication-service/pull/5735/commits/fc798b2831765c7cdb9901c73c273fd6b87d257e)

PR #5743, opened 2026-06-17
Fix ESS test CI job by pinning the setup-ess-cluster-action to the latest version

sandhose:
The `setup-ess-cluster-action` installs Helm via an unpinned `azure/setup-helm`, which now resolves to Helm v4. Helm v4 writes `Pulled:`/`Digest:` progress lines to stdout, corrupting the JSON that pyhelm3 parses, so the ESS test fixtures fail, surfacing as a JSONDecodeError or a hang to the 600s test timeout.

I fixed that upstream, this pins to the right fixed version

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -1,3 +1,4 @@
+# Copyright 2025, 2026 Element Creations Ltd.
 # Copyright 2025 New Vector Ltd.
 #
 # SPDX-License-Identifier: AGPL-3.0-only OR LicenseRef-Element-Commercial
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yaml
```diff
@@ -379,6 +380,16 @@ jobs:
         with:
           ess-helm-ref: ${{ steps.check-branch.outputs.ref }}
 
+      # The setup action's `azure/setup-helm` is unpinned, so it installs the latest Helm (v4).
```
reivilibre: feels weird to have a comment that says 'A is B' when that was true of the state before the change, but not after the introduction of the comment ?
(or do I misunderstand ?)

reivilibre (approved):
sgtm in principle

.github/workflows/ci.yaml
sandhose: What I meant is the "setup-ess-cluster-action's `azure/setup-helm`", which I think makes more sense

sandhose: Ended up fixing the thing upstream

PR #5745, opened 2026-06-19
Track and log database statistics for every request and job

sandhose:
This adds `LogContext`-level tracking of database statistics, including:

 - how much total time we spent waiting on the database
 - how many queries were executed
 - how many rows were fetched

This is done by wrapping the `Executor` (`PgConnection`, `Transaction<…>`, `PgPool`…) into a `RecordingExecutor`, which captures those operations.

Sample request and job completion log lines after this change:

```
2026-06-19T13:52:59.706373Z  INFO http.server.response POST-1074 - "POST /oauth2/token HTTP/2.0" 200 OK "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/605.1.15 (KHTML, like Gecko) Version/26.5 Safari/605.1.15" [polls: 22, cpu: 34.4ms, db: 8.8ms, elapsed: 81.6ms, queries: 10, fetched: 6] requester=oauth2-client:01KVG2FJE7ATZSVGRDWCY4FFAT client.address=2a01:e0a:f5c:5b02:b516:e9ff:72ae:c003 trace.id=a0d0b25960fb2d2cc0f762403245ec30
2026-06-19T13:53:37.223928Z  INFO mas_tasks::new_queue:934 job-sync-devices-1122 - Job completed [polls: 11, cpu: 6.0ms, db: 4.3ms, elapsed: 46.3ms, queries: 5, fetched: 5] job.id=01KVG2H5WHWYCXPWK25G1FRH5Y job.queue.name=sync-devices job.attempt=0 trace.id=28c0c37bf6de3690e0949b39ea2620dd
```

Another motivation for this PR is that `sqlx` 0.9 changed the `Execute` trait, so that `Execute::sql` now takes `self` instead of `&self`, which means we cannot use the same trick we were previously using to record the `db.query.text` in the current span.

I initially also made it show the number of "affected" rows for INSERT/UPDATE/DELETE queries, but this was a little bit too much work, as there is no trait on sqlx query results which covers this, it's per-database impl.

This should be reviewed commit by commit.


reivilibre (approved):
should be good

PR #5748, opened 2026-06-22
Align the configuration reference documentation, document missing options

### LINKED ISSUE #5057: Error for invalid well-known in mas doctor links to docs which don't mention the well-known
https://github.com/element-hq/matrix-authentication-service/blob/dd7d401c03ff61f9e29bb0c5c6f5f4be8bfe347d/crates/cli/src/commands/doctor.rs#L138

This links to https://element-hq.github.io/matrix-authentication-service/setup/homeserver.html, but there the well-known isn't documented.

In general, the docs for the well-known seem to be missing.

sandhose:
This is a round of cleaning up of the documentation, to make sure all configuration options are correctly documented.

Fixes #5057 
Replaces #4576 

It also removes mentions to the `.well-known` stuff that we had in `mas-cli doctor` but were removed a while ago from the spec. It also removes a stale `branding.logo_uri` option which did nothing.

### DIFF
--- a/docs/reference/configuration.md
+++ b/docs/reference/configuration.md
@@ -12,6 +12,20 @@ http:
   # OIDC issuer advertised by the service. Defaults to `public_base`
   issuer: https://example.com/
 
+  # List of trusted reverse proxies that are allowed to set the
+  # `X-Forwarded-For` header.
+  #
+  # Defaults to the usual private IP ranges:
+  #   192.168.0.0/16, 172.16.0.0/12, 10.0.0.0/8, 127.0.0.0/8,
+  #   fd00::/8 and ::1/128
+  #trusted_proxies:
+  #  - 192.168.0.0/16
+  #  - 172.16.0.0/12
+  #  - 10.0.0.0/8
+  #  - 127.0.0.0/8
+  #  - fd00::/8
+  #  - ::1/128
+
   # List of HTTP listeners, see below
   listeners:
     # ...
@@ -46,6 +60,9 @@ http:
           path: ./share/assets/
         # Serve the admin API on the /api/admin/v1/ path. Disabled by default
         #- name: adminapi
+
+      # Optional URL prefix to mount all the resources of this listener under
+      #prefix: /auth
 
       # List of addresses and ports to listen to
       binds:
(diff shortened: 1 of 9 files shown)


docs/reference/configuration.md
```diff
@@ -928,4 +1047,37 @@ experimental:
...
+
+     # Hard limit on the number of sessions, enforced in all contexts
+     # (interactive and non-interactive). When reached, new logins are refused
+     # unless `dangerous_hard_limit_eviction` is set.
+     #hard_limit: 50
+
+     # If set, only accounts with at most this many sessions have the session
+     # limits applied. Useful to avoid breaking legacy bots/scripts that log in
+     # repeatedly, while keeping a sane limit for everyone else.
+     #max_session_threshold: 100
```
MadLittleMods, line 1072: Having a similar but different docstring for these options is a bit 😬 

As I mentioned in [`#element-backend-internal:matrix.org`](https://matrix.to/#/!SGNQGPGUwtcPBUotTL:matrix.org/$hyX3FHTNbAlppzuvMv-ua3hJUdeYroYmcG-7jAuvoVI?via=jki.re&via=element.io&via=matrix.org), I see how having a shorter reference document could be useful to better digest, understand, and copy from (how it's currently setup). And a 👍️ on adding code-gen so we can have a single source of truth; at-least as a separate full reference would be good to link off to. For maintenance burden ease, I would be fine with just the code-gen version


Config docstring for comparison:

https://github.com/element-hq/matrix-authentication-service/blob/1beee9fec91f9cc4859f6d8538a22742be73dfde/crates/config/src/sections/experimental.rs#L168-L178

docs/reference/configuration.md
MadLittleMods, line 1: Haven't really scrutinized the options beyond `session_limit` that I'm familiar with. Leaning on your expertise and assuming it's correct ⏩

MadLittleMods (approved):

docs/reference/configuration.md
sandhose: It's probably not the only place where there is a diff here. I did intentionally reword it for now because I find the docstring itself to be quite verbose. If we're looking into codegen-ing this, maybe we should distinguish between a short description that ends up in this reference config and a longer description that ends up in the schema

sandhose:
Thanks for the review @MadLittleMods ; I'll definitely look into codegen-ing all of this 🤝

PR #5752, opened 2026-06-25
Upgrade aide, schemars and axum

sandhose:
This bumps aide on an alpha version and schemars and axum to their latest version.

There are some minor changes in the OpenAPI schema output, but they are negligible, and don't significantly change the generated API client we use in Element Admin.

### DIFF
--- a/docs/api/spec.json
+++ b/docs/api/spec.json
@@ -91,14 +91,7 @@
             "description": "Retrieve the items before the given ID",
             "schema": {
               "description": "Retrieve the items before the given ID",
-              "anyOf": [
-                {
-                  "$ref": "#/components/schemas/ULID"
-                },
-                {
-                  "type": "null"
-                }
-              ]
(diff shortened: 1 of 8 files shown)


docs/api/spec.json
```diff
@@ -91,14 +91,7 @@
...
             "schema": {
               "description": "Retrieve the items before the given ID",
-              "anyOf": [
-                {
-                  "$ref": "#/components/schemas/ULID"
-                },
-                {
-                  "type": "null"
-                }
-              ]
```
reivilibre, line 101: hm hm, does this mean we no longer (claim to) accept explicit null, but only absence?

reivilibre (approved):
I think that's OK though, as we probably don't want you sending `null`s anyway and it's not like we have to be as pedantic as something that was specced

docs/api/spec.json
sandhose: I think so, yes, but it's definitely fine

PR #5843, opened 2026-07-06
Keep the auth action context when changing accounts

### LINKED ISSUE #5763: The login flow sometimes 'forgets' what the end goal was
There are some edge cases where when the current session is logged out, or the user is deactivated, or the user is locked, we show a screen that lets you sign out.

Image

The problem is, at this point we're not carrying the context of what the user was doing, so when they log back in, they won't be continuing whatever they were doing (like logging in a client). We should carry that context in those cases to avoid breaking user login flows

sandhose:
Fixes #5763 

There are a few contexts where we check whether:

 - the session is still valid
 - the account is not locked
 - the account is not deactivated

and we show a fallback with a 'Sign out' button in those cases. Problem is that this button didn't retain what the current 'action' was, so once logged out, it would redirect to the login flow, but with the context lost.

An example flow that is being fixed: if I start a login flow but the account I'm already signed in MAS is deactivated, it tells me so, but choosing another account used to 'forget' that I was trying to log in Element. With this bug fixed, you see it continuing the login flow after re-login in with another account.

**Before:**

https://github.com/user-attachments/assets/2a23f4ab-2f5c-4ec5-bdd7-d932a7eb97a1

**After:**

https://github.com/user-attachments/assets/9e31320f-127c-4882-8119-3d8b867c7218

The video shows an example with a deactivated account, but it could be a locked account or just a browser session being remotely signed out (which can even happen automatically with the session expiration experimental feature)

### DIFF
--- a/crates/handlers/src/compat/login_sso_complete.rs
+++ b/crates/handlers/src/compat/login_sso_complete.rs
@@ -1,3 +1,4 @@
+// Copyright 2025, 2026 Element Creations Ltd.
 // Copyright 2024, 2025 New Vector Ltd.
 // Copyright 2022-2024 The Matrix.org Foundation C.I.C.
 //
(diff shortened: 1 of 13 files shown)


crates/handlers/src/compat/login_sso_complete.rs
```diff
@@ -1,3 +1,4 @@
+// Copyright 2025, 2026 Element Creations Ltd.
 // Copyright 2024, 2025 New Vector Ltd.
 // Copyright 2022-2024 The Matrix.org Foundation C.I.C.
 //
```
MadLittleMods, line 4: Seems reasonable. I feel like I can guess and assume what this does but a video would clarify it completely.

MadLittleMods (approved):

crates/handlers/src/compat/login_sso_complete.rs
sandhose: That's a good point, I've added screen recording of the bug this fixes

PR #5858, opened 2026-07-10
Replace Unix socket on startup & allow setting permissions

### LINKED ISSUE #2424: Unix socket listener cannot bind on restart
*This issue was originally created by [**@AndrewFerr**](https://github.com/AndrewFerr) at <https://github.com/matrix-org/matrix-authentication-service/issues/2424>.*

System: Fedora Linux 39 (Workstation Edition)

When using a Unix socket as a listener (via a config of `http.listeners[*].binds.socket`), the socket file is not deleted when MAS is shut down.  This prevents MAS from starting up again later with an error of:

```
Error: could not bind socket

Caused by:
    Address already in use (os error 98)
```

Only when I manually delete the socket file can MAS be started again.

Perhaps relevant: when created, the socket file has file mode bits of 0755 (user read-write, group/others read-only).

### LINKED ISSUE #5572: Be able to set permissions for Unix socket
I need to change permissions and ownership of Unix socket to connect

Have a config field to configure permissions and group/owner of Unix socket

Use TCP bind

sandhose:
Fixes #5572
Fixes #2424

This makes it so that MAS binds on a temporary path on startup, sets permissions and then moves it to the configured path. This means that we don't error out anymore if there is a stale socket file, and that it is safe to rollout a new process without killing the old one first, without breaking connections.

### DIFF
--- a/crates/cli/src/server.rs
+++ b/crates/cli/src/server.rs
@@ -7,7 +7,7 @@
 
 use std::{
     net::{IpAddr, Ipv4Addr, Ipv6Addr, SocketAddr, TcpListener, ToSocketAddrs},
-    os::unix::net::UnixListener,
+    os::unix::{fs::PermissionsExt as _, net::UnixListener},
     time::Duration,
 };
 
@@ -16,6 +16,7 @@ use axum::{
     Extension, Router,
     extract::{FromRef, MatchedPath},
 };
+use camino::Utf8PathBuf;
 use headers::{CacheControl, HeaderMapExt as _, UserAgent};
 use hyper::{Method, Request, Response, StatusCode, Version, header::USER_AGENT};
 use listenfd::ListenFd;
@@ -419,6 +420,27 @@ pub fn build_tls_server_config(config: &HttpTlsConfig) -> Result<ServerConfig, a
     Ok(config)
 }
 
+/// Utility for removing a file on drop.
+struct RemoveOnDrop<'a>(Option<&'a Utf8PathBuf>);
+
+impl Drop for RemoveOnDrop<'_> {
+    fn drop(&mut self) {
+        if let Some(path) = self.0 {
+            std::fs::remove_file(path).ok();
+        }
+    }
+}
+
+impl<'a> RemoveOnDrop<'a> {
+    fn new(path: &'a Utf8PathBuf) -> Self {
+        Self(Some(path))
+    }
+
+    fn disarm(mut self) {
+        self.0 = None;
+    }
+}
+
 pub fn build_listeners(
     fd_manager: &mut ListenFd,
     configs: &[HttpBindConfig],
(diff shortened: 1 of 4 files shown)


crates/cli/src/server.rs
```diff
@@ -454,9 +476,37 @@ pub fn build_listeners(
...
+
+                // We first bind to a temporary socket, then rename it to the desired path.
+                // This lets us replace an existing socket (binding on an existing socket
+                // doesn't work) and change the permissions of the socket before it being
+                // available.
+                let pid = std::process::id();
+                let tmp_socket = Utf8PathBuf::from(format!("{socket}.{pid}.tmp"));
+
+                // Delete the temporary socket on drop, to avoid leaving it around if we fail.
+                let guard = RemoveOnDrop::new(&tmp_socket);
```
MadLittleMods, line 494: Comment about this being a best-effort thing (not fool-proof)

MadLittleMods, line 494: To be clear, this isn't the main fix, this is just a house keeping thing right?

crates/cli/src/server.rs
```diff
@@ -454,9 +476,37 @@ pub fn build_listeners(
...
+                    .map(|mode| u32::from_str_radix(mode, 8))
+                    .transpose()?
+                    .map(std::fs::Permissions::from_mode);
+
+                // We first bind to a temporary socket, then rename it to the desired path.
+                // This lets us replace an existing socket (binding on an existing socket
+                // doesn't work) and change the permissions of the socket before it being
+                // available.
+                let pid = std::process::id();
+                let tmp_socket = Utf8PathBuf::from(format!("{socket}.{pid}.tmp"));
```
erikjohnston: If we're using pathbufs anyway we may as well do:

```suggestion
                let socket = Utf8PathBuf::from(socket);
                let tmp_socket = socket.with_added_extension(format!("{pid}.tmp"));
```

Because a) it's the "proper" way of working with paths, and b) it confused me a bit since it wasn't immediately obvious that `{socket}` would render as the proper path (albeit obvious in hindsight)

erikjohnston (approved):

crates/cli/src/server.rs
sandhose: Oh, TIL that method exists, thanks! Have done that in fc94b550

MadLittleMods (commented):
(pending review that never got submitted but still curious on the answer)

crates/cli/src/server.rs
sandhose: It's indeed not fool-proof: if during the very short amount of time where the process has created the temp socket but hasn't been renamed yet, it gets killed by something else, the temp socket will stay around.

But it has to be external: panic'ing here or returning an error will clean up the socket

sandhose: The `RemoveOnDrop` yes. The main fix is that we're always listening to a new socket then replacing the old one. An alternative fix would have been to `unlink` (delete) the old socket first then reopening right away. But the advantage of 'listening first then rename' is that there is zero amount of time where someone trying to contact the socket errors out.

By doing this, you can rollout new versions by spinning up a new instance *then* killing the old one. The old one will happily serve existing connections, and when it disappears, things will reconnect on the new socket, therefore on the new instance

PR #5859, opened 2026-07-10
Expose a `logged_out` variable to upstream authorization parameter templates

### LINKED ISSUE #1569: Signing out from the service does not force re-logging in from the upstream IdP
*This issue was originally created by [**@hughns**](https://github.com/hughns) at <https://github.com/matrix-org/matrix-authentication-service/issues/1569>.*

A current experience:

https://github.com/matrix-org/matrix-authentication-service/assets/6955675/9654700d-4c96-44ce-9d18-9f78128b9051

sandhose:
When a browser signs out (or discovers its session was ended), the session cookie now remembers when that happened; the flag is cleared by the next successful login. The `additional_authorization_parameters` templates can read it as `logged_out`, so operators can force a fresh prompt at the upstream provider after sign-out, e.g.:

```yaml
    additional_authorization_parameters:
      prompt: "{% if logged_out %}login{% endif %}"
```

This addresses the "sign out signs me back in" loop with upstream SSO providers that still hold a live session.

Fixes #1569


crates/config/src/sections/upstream_oauth2.rs
```diff
@@ -1,4 +1,4 @@
-// Copyright 2026 Element Creations Ltd.
+// Copyright 2025, 2026 Element Creations Ltd.
```
anoadragon453, line 1: Was this intentional?

Likewise below.

docs/setup/sso.md
```diff
@@ -98,6 +99,26 @@ upstream_oauth2:
...
+```yaml
+upstream_oauth2:
+  providers:
+    - id: …
+      additional_authorization_parameters:
+        # Send `prompt=login` only right after the browser signed out.
+        prompt: "{% if logged_out %}login{% endif %}"
+```
+
+`logged_out` is `true` only for a browser that has signed out and has not signed back in — a fresh browser, or one with an active session (such as during account linking), renders the template to an empty string, so no `prompt` parameter is forwarded in that case.
```
anoadragon453: nit:

```suggestion
`logged_out` is `true` only for a browser that has signed out and has not signed back in. A fresh browser, or one with an active session (such as during account linking), renders the template to an empty string, so no `prompt` parameter is forwarded in that case.
```

crates/handlers/src/upstream_oauth2/authorize.rs
```diff
@@ -71,6 +73,16 @@ pub(crate) async fn get(
...
 ) -> Result<impl IntoResponse, RouteError> {
+    // Load the session info from the cookie jar. We use this to know whether
+    // the browser recently signed out, which we expose to the
+    // `additional_authorization_parameters` templates as `logged_out` so that
+    // operators can force a fresh prompt at the upstream provider. We consider
+    // that we were logged out recently if we logged out within the last 5 minutes.
+    let (session_info, cookie_jar) = cookie_jar.session_info();
+    let logged_out = session_info
+        .logged_out_at()
+        .is_some_and(|t| t.signed_duration_since(clock.now()) < Duration::minutes(5));
```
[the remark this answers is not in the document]
anoadragon453: And add a unit test for checking that `logged_out` is false when enough time has passed.

anoadragon453 (changes requested):
Minor bits below. Nice QoL fix!

crates/config/src/sections/upstream_oauth2.rs
sandhose: This is because this is the right copyright. The New Vector Ltd. -> Element Creations Ltd. happened during 2025, so the proper rule is essentially:

 - when updating an existing file that was created in 2025 or before, the last copyright holder should be `Copyright 2025, 2026 Element Creations Ltd.`
 - for new files, it should be `Copyright 2026 Element Creations Ltd.`

my agents review now tends to auto-correct those because I explained this rule whenever I touch a file

crates/handlers/src/upstream_oauth2/authorize.rs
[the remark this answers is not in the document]
sandhose: Duh. Fixed in [07857f5](https://github.com/element-hq/matrix-authentication-service/pull/5859/commits/07857f591f3ef693008a1c2246b9307aa5cec9ec) including a test

anoadragon453 (approved):
LGTM, thank you!

PR #5862, opened 2026-07-10
Provide more specific password login failure reasons in metrics

reivilibre:
This is probably not ready to merge as-is, but I just got paged because we had a large rate of 'error' password logins.

Problem is, this seems to not be a very useful metric currently and it seems like it could just be someone hitting their ratelimit and spamming requests.

I'm opening this as a discussion starter, but we probably want to figure out what we actually want from the alerting on this one.

### DIFF
--- a/crates/handlers/src/views/login.rs
+++ b/crates/handlers/src/views/login.rs
@@ -170,7 +170,7 @@ pub(crate) async fn post(
 
     if !form_state.is_valid() {
         tracing::warn!("Invalid login form: {form_state:?}");
-        PASSWORD_LOGIN_COUNTER.add(1, &[KeyValue::new(RESULT, "error")]);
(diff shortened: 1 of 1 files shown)


crates/handlers/src/views/login.rs
```diff
@@ -170,7 +170,7 @@ pub(crate) async fn post(
 
     if !form_state.is_valid() {
         tracing::warn!("Invalid login form: {form_state:?}");
-        PASSWORD_LOGIN_COUNTER.add(1, &[KeyValue::new(RESULT, "error")]);
```
sandhose, line 173: I think to avoid breaking the existing metric, I would rather add a new label? like `{result="error", reason="form-invalid"}`? Most dashboard and alerting rules should anyway `sum on (result) …` or similar since MAS scales horizontally and usually has multiple instances, so adding cardinality with a new label isn't a breaking change

sandhose (changes requested):

PR #5863, opened 2026-07-13
Escape whole translations by default on server-side templates

sandhose:
We've got a few translations that have HTML markups in them, which led us to manually escape placeholders only, then mark the whole translated string as safe.

This however shouldn't be the default, as most translations *don't* have HTML tags in them. This is particularly problematic in places where we inject translations in HTML attributes, as a translation could end the attribute with a `"` that won't be escaped.

This makes it so that by default, the whole string is auto-escaped, and callers can opt-out by using `_(…, html=true)` in templates.

### DIFF
--- a/crates/templates/src/functions.rs
+++ b/crates/templates/src/functions.rs
@@ -1,4 +1,4 @@
-// Copyright 2026 Element Creations Ltd.
+// Copyright 2025, 2026 Element Creations Ltd.
 // Copyright 2024, 2025 New Vector Ltd.
 // Copyright 2021-2024 The Matrix.org Foundation C.I.C.
 //
@@ -291,8 +291,13 @@ impl Object for TranslateFunc {
                 ))?
         };
 
+        // Whether the translation is trusted to contain HTML markup
+        let html = kwargs.get::<Option<bool>>("html")?.unwrap_or(false);
+
         let res: Result<ArgumentList, Error> = kwargs
             .args()
+            // `html` is a control flag, not a message argument
+            .filter(|name| *name != "html")
             .map(|name| {
                 let value: Value = kwargs.get(name)?;
                 let value = serde_json::to_value(value).map_err(|e| {
@@ -309,6 +314,14 @@ impl Object for TranslateFunc {
             Error::new(ErrorKind::InvalidOperation, "Could not format message").with_source(e)
         })?;
 
+        if !html {
+            // Return a plain value, letting auto-escape escape the whole
+            // message at output time
+            return Ok(Value::from(formatted.to_string()));
+        }
+
+        // The literal text is trusted markup; only the placeholders are
+        // escaped, and the result is marked safe
         let mut buf = String::with_capacity(formatted.len());
         let mut output = make_string_output(&mut buf);
         for part in formatted.parts() {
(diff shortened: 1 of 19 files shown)


crates/templates/src/functions.rs
```diff
@@ -642,3 +655,87 @@ impl Object for Counter {
...
+
+    use super::*;
+
+    /// Build a `MiniJinja` environment with the translation function registered
+    /// against a small in-memory set of English messages.
+    fn environment() -> minijinja::Environment<'static> {
+        let en: TranslationTree = serde_json::from_value(serde_json::json!({
+            "plain": "Hello %(name)s!",
+            "markup": "Hello <strong>%(name)s</strong>!",
+            "uses_html": "value=%(html)s",
```
MadLittleMods, line 671:
```suggestion
            // We expect this to fail to render because TODO
            "using_html_arg_is_not_allowed": "value=%(html)s",
```

crates/templates/src/functions.rs
```diff
@@ -642,3 +655,87 @@ impl Object for Counter {
...
+    #[test]
+    fn test_not_escaped_in_text_templates() {
+        // With auto-escape off (`.txt` templates, email bodies/subjects), the
+        // default path emits the message untouched
+        let rendered = render_as(
+            "test.txt",
+            r#"{% set _ = translator("en") %}{{ _("plain", name="<b>&") }}"#,
+        )
+        .unwrap();
+        assert_eq!(rendered, "Hello <b>&!");
```
MadLittleMods, line 725: Why is auto-escape off here? I would expect auto-escape to be off when `html=true`

MadLittleMods: It's described above as "Whether the translation is trusted to contain HTML markup" 🤔 

Either way, it seems ambiguous. Better name?

templates/emails/verification.html
```diff
@@ -16,4 +17,4 @@
 
 {{ _("mas.emails.greeting", username=(username|default("user"))) }}<br />
 <br />
-{{ _("mas.emails.verify.body_html", code=authentication_code.code) }}<br />
+{{ _("mas.emails.verify.body_html", code=authentication_code.code, html=true) }}<br />
```
MadLittleMods, line 20: Would it make more sense to define this in the translation itself? 🤔 

Feels like translations are very usage specific anyway.

crates/templates/src/functions.rs
sandhose: It's minijinja which has different escaping depending on the extension. `.html` applies escaping according to HTML escaping rules, `.json` with JSON escaping rules… and `.txt` has no escaping. It you're right that it does make the `html=true` sound weird… maybe `dont_escape=true`? `safe=true`? `escape=false`?

templates/emails/verification.html
sandhose: Well I prefer the usage site to reflect 'oh, this translation possibly has HTML tags in it' rather than having to dig in the translation itself

MadLittleMods: hmmm, the `_html` convention used here is good for translators (and even us) to know as well but we don't use it everywhere (not something we need to change everywhere in this PR, just a note).

crates/templates/src/functions.rs
```diff
@@ -642,3 +655,87 @@ impl Object for Counter {
...
+    use mas_i18n::{Translator, translations::TranslationTree};
+
+    use super::*;
+
+    /// Build a `MiniJinja` environment with the translation function registered
+    /// against a small in-memory set of English messages.
+    fn environment() -> minijinja::Environment<'static> {
+        let en: TranslationTree = serde_json::from_value(serde_json::json!({
+            "plain": "Hello %(name)s!",
+            "markup": "Hello <strong>%(name)s</strong>!",
```
MadLittleMods, line 670:
```suggestion
            "html_markup": "Hello <strong>%(name)s</strong>!",
```

crates/templates/src/functions.rs
MadLittleMods: If it's a generic escape toggle in every language, then `escape=false` makes sense.

PR #5864, opened 2026-07-13
Validate downloaded translations against the English source

sandhose:
Add a check that every translated server-side string (`translations/*.json`) has exactly the same `sprintf` placeholders and HTML tags/attributes as the English source. Server templates render translation markup as trusted raw HTML, so a compromised/incorrect translation could otherwise inject arbitrary markup.

### DIFF
--- a/crates/i18n/src/check.rs
+++ b/crates/i18n/src/check.rs
@@ -1,0 +1,921 @@
+// Copyright 2026 Element Creations Ltd.
+//
+// SPDX-License-Identifier: AGPL-3.0-only OR LicenseRef-Element-Commercial
+// Please see LICENSE files in the repository root for full details.
+
+//! Validation that downloaded translations match the English source.
+//!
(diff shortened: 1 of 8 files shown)


crates/i18n/src/check.rs
```diff
@@ -0,0 +1,921 @@
...
+                        continue;
+                    };
+                    // Compare against the source's *same* category when it has
+                    // one (a source `one` and `other` legitimately differ), and
+                    // only fall back to a reference for categories the source
+                    // lacks (target locales have more categories than English).
+                    let reference = source_map
+                        .get(category)
+                        .and_then(Value::as_str)
+                        .or_else(|| plural_reference(source_map));
```
MadLittleMods, line 315: How do plurals work?

I'd imagine the placeholders around plurals is bound to differ.

Also, we have some cases where we provide placeholder data but don't use it in the English translation but it could be useful for the translated version.

PR #5866, opened 2026-07-15
Move "Full release process" section higher up

MadLittleMods:
Move "Full release process" section higher up

When you get linked to https://element-hq.github.io/matrix-authentication-service/development/releasing.html#releasing it's confusing to know what to do. Based on descriptions of the GitHub actions, I managed to do everything right but then I noticed this was actually documented in the "Full release process" section at the bottom.

I think instead it makes sense to give the steps first and then have the in-depth descriptions of each workflow below.


sandhose (approved):

MadLittleMods:
Thanks for the review @sandhose 🐠

PR #5867, opened 2026-07-16
Fix the password complexity calculation always returning 0, preventing password changes

sandhose:
This is a regression introduced by upgrading zxcvbn in #5802 

Turns out, their types don't agree with their built files. I've opened https://github.com/zxcvbn-ts/zxcvbn/issues/329

I've added a test to catch such regression

Context: https://matrix.to/#/!yHWhpxlXVaLcsgDUKb:matrix.org/$eX4OPBG8n6C_k--pnvEKdhdWvPbkbWeRgkp5UrQkFNU?via=matrix.org&via=element.io&via=vector.modular.im

### DIFF
--- a/frontend/src/utils/password_complexity/index.test.ts
+++ b/frontend/src/utils/password_complexity/index.test.ts
@@ -1,0 +1,52 @@
+// Copyright 2026 Element Creations Ltd.
+//
+// SPDX-License-Identifier: AGPL-3.0-only OR LicenseRef-Element-Commercial
+// Please see LICENSE files in the repository root for full details.
+
+import type { TFunction } from "i18next";
+import { describe, expect, it } from "vitest";
+
+import { estimatePasswordComplexity } from "./index";
+
+// Fake translator that echoes the key back, so we can assert which
+// warning/suggestion strings the estimator produced.
+const t = ((key: string) => key) as unknown as TFunction;
+
+describe("estimatePasswordComplexity()", () => {
+  it("scores a strong password highly", async () => {
+    const result = await estimatePasswordComplexity(
+      "correcthorsebatterystaple",
+      t,
+    );
+    expect(result.score).toBeGreaterThanOrEqual(3);
+  });
+
+  it("scores a common password as the weakest", async () => {
+    const result = await estimatePasswordComplexity("password", t);
+    expect(result.score).toBe(0);
+  });
+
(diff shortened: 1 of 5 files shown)


frontend/src/utils/password_complexity/index.ts
```diff
@@ -42,10 +42,21 @@ const l33tTable = {
   z: ["2"],
 };
 
+// The ESM build of @zxcvbn-ts/language-common nests the graphs under a `default`
+// export its type declarations don't advertise, so the top-level qwerty/dvorak/…
+// read as `undefined`. Feeding those to zxcvbn crashes its spatial matcher and
+// silently disables password strength, so unwrap `default` when it's there.
+// See https://github.com/zxcvbn-ts/zxcvbn/issues/329
```
MadLittleMods: It looks like https://github.com/zxcvbn-ts/zxcvbn/issues/329 was fixed so this is no longer necessary if we want to upgrade the dependency

MadLittleMods (approved):

frontend/src/utils/password_complexity/index.ts
sandhose: Yeah but they have not released yet as far as I can tell; I'll keep an eye on it and remove that workaround when we upgrade

MadLittleMods: Ahh, I saw the tags has something new from 38 minutes ago but I guess that's not the one we care about :fast_forward: (lots of things being tagged in that repo)

sandhose: Nevermind, they tagged in the meantime :D
Let's upgrade then

frontend/src/utils/password_complexity/index.test.ts
```diff
@@ -0,0 +1,52 @@
...
+import type { TFunction } from "i18next";
+import { describe, expect, it } from "vitest";
+
+import { estimatePasswordComplexity } from "./index";
+
+// Fake translator that echoes the key back, so we can assert which
+// warning/suggestion strings the estimator produced.
+const t = ((key: string) => key) as unknown as TFunction;
+
+describe("estimatePasswordComplexity()", () => {
```
MadLittleMods, line 15: > [...] password complexity calculation always returning 0, preventing password changes

Why was it defaulting to `0` if we were running into an error?

PR #5869, opened 2026-07-16
Admin API: allow searching in upstream OAuth links on 'human account name'

sandhose:
This adds a filter on the upstream OAuth links admin API to do a substring case insensitive search on the 'human account name' on upstream OAuth links

Context: https://github.com/element-hq/legal-compliance/issues/547#issuecomment-4993944225

### DIFF
--- a/crates/handlers/src/admin/v1/upstream_oauth_links/list.rs
+++ b/crates/handlers/src/admin/v1/upstream_oauth_links/list.rs
@@ -1,3 +1,4 @@
+// Copyright 2025, 2026 Element Creations Ltd.
 // Copyright 2025 New Vector Ltd.
 //
 // SPDX-License-Identifier: AGPL-3.0-only OR LicenseRef-Element-Commercial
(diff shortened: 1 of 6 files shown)


crates/handlers/src/admin/v1/upstream_oauth_links/list.rs
```diff
@@ -63,6 +69,11 @@ impl std::fmt::Display for FilterParams {
             sep = '&';
         }
 
+        if let Some(human_account_name) = &self.human_account_name {
+            write!(f, "{sep}filter[human-account-name]={human_account_name}")?;
```
MadLittleMods, line 73: Would be nice to reference `filter[human-account-name]` from a single source of truth

crates/storage-pg/src/upstream_oauth2/mod.rs
MadLittleMods, line 43: Prior art: Massive combined test

MadLittleMods (approved):

crates/handlers/src/admin/v1/upstream_oauth_links/list.rs
sandhose: I wish we could, but serde's replace doesn't let us use constants, so we can't have share it with the other place that has this constant (outside of tests)

PR #5874, opened 2026-07-24
Allow setting a `client_uri` for static clients in the config

V02460:
Implements configuration of the `client_uri` field for static clients. The field is required to show a user-facing URL for the client on the consent page.


sandhose (approved):
Not sure why the frontend test suite doesn't work, that seems unrelated

PR #5876, opened 2026-07-27
Fix time-dependent snapshot test in account sessions page

sandhose:
The mocked sessions have a hardcoded `lastActiveAt` of `2026-04-23`, which crossed the 90-day inactivity threshold on `2026-07-22`, flipping the rendered label from "Active <date>" to "Inactive for 90+ days" and breaking the committed snapshot. Freeze the clock in the test file so the fixture stays within the active window.


hughns (approved):
👍

PR #5881, opened 2026-07-28
Accept `[` and `~` in OAuth 2.0 scope tokens

### LINKED ISSUE #5878: device ID scopes that include `~` are not accepted
**Describe the bug**
MAS throws an error (`Failed to deserialize query string: Invalid scope format`) if any device ID scope includes `~`, despite that being allowed by spec: [https://spec.matrix.org/v1.19/client-server-api/#device-id-allocation](https://spec.matrix.org/v1.19/client-server-api/#device-id-allocation)

**To Reproduce**
Try to go to an auth URL with a device ID scope that includes `~`.

 For a reproducer, https://nexus.federated.nexus hits this pretty often. You can restart the app then tap Matrix.org, and repeat until it fails. It tends to have about a 50/50 shot.

**Expected behavior**
It should accept the device ID, as it is valid.

 - OS: NixOS
 - Browser: Firefox
 - Version: 152.0.6-1

sandhose:
The `nqchar` predicate used exclusive ranges where RFC 6749's NQCHAR ranges (%x23-5B / %x5D-7E) are inclusive on both ends, so scopes containing `[` or `~` were rejected with "Invalid scope format". In particular this broke `urn:matrix:client:device:` scopes for device IDs containing `~`, which the Matrix spec explicitly allows.

Fixes #5878

### DIFF
--- a/crates/oauth2-types/src/scope.rs
+++ b/crates/oauth2-types/src/scope.rs
@@ -228,6 +228,10 @@ mod tests {
         assert_eq!(ScopeToken::from_str("openid"), Ok(OPENID));
 
         assert_eq!(ScopeToken::from_str("invalid\\scope"), Err(InvalidScope));
+
+        // Regression test for #5878: Matrix device IDs may contain any
+        // RFC 3986 unreserved character, including `~`.
+        assert!(ScopeToken::from_str("urn:matrix:client:device:AA~bb1234_-.").is_ok());
     }
 
     #[test]
@@ -244,19 +248,14 @@ mod tests {
             );
         }
 
-        // Boundary characters affected by the off-by-one bug.
-        for c in ['!', '#', '[', ']', '~'] {
-            assert!(ScopeToken::from_str(&c.to_string()).is_ok());
-        }
-
-        // Outside NQCHAR.
-        for s in [" ", "\"", "\\", "\x7F", ""] {
+        // Outside NQCHAR:
+        //  - space is 0x20
+        //  - " (double quote) is 0x22
+        //  - \ (backslash) is 0x5C
+        //  - DEL is 0x7F
+        for s in [" ", "\"", "\\", "\x7F"] {
             assert_eq!(ScopeToken::from_str(s), Err(InvalidScope));
         }
-
-        // Regression test for #5878: Matrix device IDs may contain any
-        // RFC 3986 unreserved character, including `~`.
-        assert!(ScopeToken::from_str("urn:matrix:client:device:AA~bb1234_-.").is_ok());
     }
 
     #[test]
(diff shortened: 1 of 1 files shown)


crates/oauth2-types/src/scope.rs
```diff
@@ -1,4 +1,4 @@
-// Copyright 2026 Element Creations Ltd.
+// Copyright 2025, 2026 Element Creations Ltd.
```
sandhose, line 1: This is a pass-by change: the copyright holder changed during 2025, so existing files are *supposed* to say 'Element Creations Ltd' *since 2025* for existing. My review agent is told to care about this, so it tends to insert those whenever they are wrong

crates/oauth2-types/src/scope.rs
```diff
@@ -228,6 +230,35 @@ mod tests {
...
+                ScopeToken::from_str(&c.to_string()).is_ok(),
+                "expected {c:?} (0x{:02X}) to be a valid NQCHAR",
+                c as u32
+            );
+        }
+
+        // Boundary characters affected by the off-by-one bug.
+        for c in ['!', '#', '[', ']', '~'] {
+            assert!(ScopeToken::from_str(&c.to_string()).is_ok());
+        }
```
MadLittleMods: This is a sanity check on top of of the loop above?

crates/oauth2-types/src/scope.rs
```diff
@@ -228,6 +230,35 @@ mod tests {
...
+
+        // Outside NQCHAR.
+        for s in [" ", "\"", "\\", "\x7F", ""] {
+            assert_eq!(ScopeToken::from_str(s), Err(InvalidScope));
+        }
```
MadLittleMods: A bit hard to know if these are correct (for example `"` being `\x22`). Perhaps we should just use the hex version for all

crates/oauth2-types/src/scope.rs
```diff
@@ -228,6 +230,35 @@ mod tests {
...
+
+        // Regression test for #5878: Matrix device IDs may contain any
+        // RFC 3986 unreserved character, including `~`.
+        assert!(ScopeToken::from_str("urn:matrix:client:device:AA~bb1234_-.").is_ok());
```
MadLittleMods: Separate test

Could be part of `parse_scope_token`

MadLittleMods (approved):

crates/oauth2-types/src/scope.rs
sandhose: yeah, maybe not necessary. Removed it in b458faed

crates/oauth2-types/src/scope.rs
sandhose: I've sticked in a comment in  b458faed

crates/oauth2-types/src/scope.rs
sandhose: Have done that in b458faed

sandhose:
Thanks for the review @MadLittleMods! 🦦

crates/oauth2-types/src/scope.rs
MadLittleMods: It could be fine to keep it as well but just call it out as a sanity check on top of the loop

PR #5885, opened 2026-07-30
Render the registration CAPTCHAs with React

sandhose:
This moves the CAPTCHA rendering logic to React components, in preparation of the registration form migration to React in #5591 

Most of the complexity is around how to load nicely the remote scripts, and make the CAPTCHA component nicely suspend until it's ready


frontend/src/components/Captcha.tsx
```diff
@@ -0,0 +1,390 @@
...
+  | "cloudflare_turnstile"
+  | "hcaptcha";
+
+export type CaptchaConfig = {
+  service: CaptchaService;
+  site_key: string;
+};
+
+// Cache of script URLs to their load promises, so each script is only loaded
+// once even if multiple components request it.
```
reivilibre, line 13: minor eyebrow raise at having to do ourselves. I think dynamic `import()` would do this itself, but I guess that would involve adding an extra network trip of latency because we'd have to `import()` our own module that then adds the `<script>` tag?

reivilibre (approved):
The PR title is misleading, right? This only sets up a component that would let us do that, but the jinja captcha template that loads the script in the header is still what's in use?

frontend/src/components/Captcha.tsx
sandhose: This is a requirement of t[he React `use` API](https://react.dev/reference/react/use#use-promise): you can use it to suspend the render, but the promise you pass to `use` *must* be stable, hence it can't be recreated on each render

sandhose:
> The PR title is misleading, right? This only sets up a component that would let us do that, but the jinja captcha template that loads the script in the header is still what's in use?

dang, that was one refactoring too much done in the context of another PR on top of this. I restored that in [ee1a104](https://github.com/element-hq/matrix-authentication-service/pull/5885/commits/ee1a104705639fe6600916d3d26cf1d11c20c267) and simplified back down the whole thing in [d0370a6](https://github.com/element-hq/matrix-authentication-service/pull/5885/commits/d0370a6b3fa7edc71958596d1379cdea1cb0be64)

PR #5886, opened 2026-07-30
Read javascript entrypoints configuration from data attributes

sandhose:
Instead of injecting configuration through inline `<script>` tags setting window globals, server-rendered templates now carry the configuration as data attributes on the entrypoint mount node, validated client-side with a valibot schema over `element.dataset`. This removes all inline scripts from the templates, which is friendlier to CSP, and gives the frontend runtime-validated, typed configuration.

This is in preparation of the registration form migration to React in #5591, as this makes managing the config for multiple JS entrypoints less magic, more logical


frontend/src/entrypoints/main.tsx
```diff
@@ -1,33 +1,55 @@
...
+const router = makeRouter(config.root, queryClient);
+
+createRoot(rootElement).render(
   <StrictMode>
     <QueryClientProvider client={queryClient}>
       <ErrorBoundary>
         <TooltipProvider>
           <Suspense fallback={<LoadingScreen />}>
             <I18nextProvider i18n={i18n}>
-              <RouterProvider router={router} context={{ queryClient }} />
```
sandhose, line 30: This was actually already set in the router, no need to override the context here

frontend/src/graphql.ts
```diff
@@ -1,26 +1,22 @@
...
-import { QueryClient } from "@tanstack/react-query";
 import type { ExecutionResult } from "graphql";
-import appConfig from "./config";
 import type { TypedDocumentString } from "./gql/graphql";
 
-let graphqlEndpoint: string;
-if (import.meta.env.TEST && typeof window === "undefined") {
-  graphqlEndpoint = new URL(
-    appConfig.graphqlEndpoint,
-    "http:://localhost/",
```
sandhose, line 16: There was a typo here

frontend/src/graphql.ts
```diff
@@ -66,16 +67,8 @@ export const graphqlRequest = async <TData, TVariables>({
...
 
   if (!json.data) {
-    throw new Error(`GraphQL request to ${graphqlEndpoint} returned no data`);
+    throw new Error(`GraphQL request to ${endpoint} returned no data`);
   }
 
   return json.data;
 };
-
-export const queryClient = new QueryClient({
```
sandhose, line 75: We now prefer to create it in the entrypoint & pass it in the router + QueryClientProvider there

frontend/src/router.tsx
```diff
@@ -1,28 +1,28 @@
...
-// Create a new router instance
-export const router = createRouter({
-  routeTree,
-  scrollRestoration: true,
-  basepath: config.root,
-  defaultPendingComponent: LoadingScreen,
-  defaultPreload: "intent",
-  context: { queryClient },
-});
+export const makeRouter = (basepath: string, queryClient: QueryClient) =>
```
sandhose, line 13: Because there is no 'global config' we can import from, we need to inject things like the base path of the app as a parameter

frontend/index.html
```diff
@@ -14,15 +15,10 @@
     <meta charset="UTF-8" />
     <meta name="viewport" content="width=device-width, initial-scale=1.0" />
     <title>matrix-authentication-service</title>
-    <script type="application/javascript">
```
sandhose, line 17: This is the main advantage for now: no more bare <script> tags in the app anymore, which will allow us to set [a strict CSP](https://developer.mozilla.org/en-US/docs/Web/HTTP/Guides/CSP)

reivilibre (approved):

PR #5902, opened 2026-08-04
Update opa-wasm to update wasmtime and fix RUSTSEC-2026-0222

sandhose:
`cargo-deny` is currently failing because of that vulnerability. Updating opa-wasm also brings a bunch of updated dependencies (mostly RustCrypto crates but also `rand`), which means I had to add a bunch of exceptions to the `deny.toml` duplicate check skip list for now.


sandhose:
Closing in favor of #5903

PR #5903, opened 2026-08-04
Ignore RUSTSEC-2026-0222 as we are not affected by it

sandhose:
See https://rustsec.org/advisories/RUSTSEC-2026-0222.html and https://github.com/bytecodealliance/wasmtime/security/advisories/GHSA-hgjw-h833-99q9

Evidence as why we're not affected, the advisory states:

> This bug is not triggerable by guest WebAssembly programs. It requires the embedder to create two Engines and mix objects between them. 

And we're only ever creating a `wasmtime::Engine` once [here](https://github.com/element-hq/matrix-authentication-service/blob/19f3d146a5cc1c8ec12df74ffe774a8648b0528c/crates/policy/src/lib.rs#L231)

### DIFF
--- a/deny.toml
+++ b/deny.toml
@@ -23,6 +23,10 @@ ignore = [
     # Rand 0.8.5 unsoundness with custom logger + thread_rng reseeding.
     # Only triggers when the `log` feature is enabled, which we don't use.
     "RUSTSEC-2026-0097",
+    # A bug in wasmtime, but we are not affected, as we only have a single
+    # Engine/Store, and as the advisory says explicitly:
+    # > This bug is not triggerable by guest WebAssembly programs.
+    "RUSTSEC-2026-0222",
 ]
 
 [licenses]


deny.toml
```diff
@@ -23,6 +23,10 @@ ignore = [
     # Rand 0.8.5 unsoundness with custom logger + thread_rng reseeding.
     # Only triggers when the `log` feature is enabled, which we don't use.
     "RUSTSEC-2026-0097",
+    # A bug in wasmtime, but we are not affected, as we only have a single
+    # Engine/Store, and as the advisory says explicitly:
+    # > This bug is not triggerable by guest WebAssembly programs.
+    "RUSTSEC-2026-0222",
```
MadLittleMods, line 29: For context, this was also [discussed](https://matrix.to/#/!SGNQGPGUwtcPBUotTL:matrix.org/$mhxRL4WzcY4U4UwjgxdGZOQx1XfJigMDUA0_0_P_4rc?via=jki.re&via=element.io&via=matrix.org) in the internal backend room.

We chose to ignore as the `carg-deny` CI job is failing and we want to unblock it so we can continue the release. Ignoring is better as some dependency upgrades have caused regressions recently.

The alternative to ignoring is to update `opa-wasm` -> `wasmtime` (but also updates a bunch of other child deps)

Since we're not affected either way, ignoring seems extra fine for the release.

MadLittleMods (approved):

PR #5911, opened 2026-08-06
docs: Merge error description for the same status code

V02460:
Merge docs for the two error cases of POST /api/admin/v1/users with HTTP status code 409.

Aide’s `response_with` method overrides infos for a HTTP status code if already given before. In this case this results in the docs not mentioning the error case of an already existing user.


reivilibre (approved):
Thanks!

Issue #5913, opened 2026-08-07
Set a strict Content Security Policy


sandhose:
We should have a strict [Content Security Policy](https://developer.mozilla.org/fr/docs/Web/HTTP/Guides/CSP) and restrict from where resources can be loaded

 - #5912 
 - #5885 
 - #5886 
 - #5914

PR #5915, opened 2026-08-07
Load frontend translations with JS dynamic imports rather than fetching as JSON

sandhose:
This changes the translation loading logic to bundle things as JS files instead of JSON. This behaves better in terms of caching and will allow us restricting the `connect-src` on the CSP to only the GraphQL endpoint (for #5913)

The weird part about this PR is essentially for emitting the `modulepreload` intelligently


crates/templates/src/functions.rs
```diff
@@ -485,6 +485,20 @@ impl Object for IncludeAsset {
...
+            mas_spa::FileType::Script if main.is_dynamic_entry() => {
+                // Chunks which are only reachable through a dynamic `import()`
+                // (like the translation files) must not be evaluated eagerly,
+                // we only hint the browser to fetch them
+                let integrity = main.integrity_attr();
+                let src = main.src(assets_base);
+                if tracker.mark_preloaded(&src) {
+                    writeln!(
+                        output,
+                        r#"<link rel="modulepreload" href="{src}" crossorigin="anonymous"{integrity} />"#
```
reivilibre, line 497: is there actually benefit in preloading all the translations? I'd expect most users to only need one set of translations and so most of them will be unused.

reivilibre (approved):

crates/templates/src/functions.rs
sandhose: It's not preloading all the translations, only the detected one. A little bit confusingly, this is the thing doing the preload magic: https://github.com/element-hq/matrix-authentication-service/blob/8892cc7eb84accdaf75f8507f65aae0a5b8c3a12/templates/app.html#L21

PR #5926, opened 2026-08-18
Add unittest for database connection via pgpass

V02460:
Adds a unittest that checks whether database connections can be established via a pgpass-provided user password. I currently rely on this behavior as the method for passing the password via file in combination with a password-less database URI (that in turn allows custom connection arguments). Because it’s a bit niche, I want to make sure that it does not break by accident.

I added a custom helper for setting environment variables, but a dedicated crate like [temp-env](https://crates.io/crates/temp-env) could be used instead, if desired. Because sqlx’s `DatabaseConnectOptions` struct has crate-only access modifiers, the only direct way I could think of inspecting it for the right password was via its debug output. Alternatively, one could attempt an actual database connection in the test case, but I would need advice on how to implement that.

### DIFF
--- a/crates/cli/src/util.rs
+++ b/crates/cli/src/util.rs
@@ -627,4 +627,61 @@ mod tests {
         let manager = password_manager_from_config(&config).await;
         assert!(manager.is_err());
     }
+
+    /// RAII guard that removes an environment variable when dropped,
+    /// ensuring cleanup even if the test panics.
+    struct EnvVarGuard(&'static str);
+
+    impl EnvVarGuard {
+        fn set(key: &'static str, value: impl AsRef<std::ffi::OsStr>) -> Self {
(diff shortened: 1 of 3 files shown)


crates/cli/src/util.rs
```diff
@@ -627,4 +627,61 @@ mod tests {
...
+
+        let opts = DatabaseConnectOptions {
+            log_slow_statements: false,
+        };
+
+        let result = database_connect_options_from_config(&config, &opts).await;
+        assert!(result.is_ok());
+        let debug = format!("{:?}", result.unwrap());
+        assert!(
+            debug.contains("testpassword"),
```
reivilibre, line 683: I hope they don't eventually censor the debug output or something, but I guess we can deal with that if/when it comes to it.

reivilibre (approved):

PR #5932, opened 2026-08-21
Format log timestamps in the local timezone

### LINKED ISSUE #5931: Log timestamps are hardcoded to UTC and cannot be made timezone-aware
**Version:** v1.22.0 (`ghcr.io/element-hq/matrix-authentication-service:latest`, rev `f28451c`)

## Problem

MAS always logs UTC. In a Docker Compose stack where every other service logs in the host's
timezone, MAS is the only one offset from the host clock, which makes cross-service correlation
difficult:

```
$ date
Fri Aug 21 09:44:19 PM CEST 2026

$ docker compose logs --tail=1 synapse mas
synapse-1  | 2026-08-21 21:44:27,522 - synapse.app.homeserver - 213 - INFO - main - Running
mas-1      | 2026-08-21T19:44:11.244661Z  INFO mas_tasks::new_queue:608 worker-run-loop-1 - I'm the leader now
```

Neither `TZ=Europe/Berlin` nor bind-mounting `/etc/localtime:/etc/localtime:ro` changes the output,
and there is no config key for it. (`RUST_LOG` only affects filtering.)

## Root cause

`EventFormatter` calls the UTC-only timer directly, so it is not reachable via
`fmt::layer().with_timer(..)` — a custom `FormatEvent` impl owns the whole line:

```rust
// crates/context/src/fmt.rs:101
SystemTime.format_time(&mut writer)?;
```

```rust
// crates/cli/src/main.rs:99
let fmt_layer = tracing_subscriber::fmt::layer()
    .with_writer(log_writer)
    .event_format(mas_context::EventFormatter)   // <- overrides any .with_timer()
    .with_ansi(with_ansi);
```

`tracing_subscriber::fmt::time::SystemTime` has no timezone concept at all — it converts
`duration_since(UNIX_EPOCH)` to a civil datetime with a port of musl's `__secs_to_tm` and
hardcodes the `Z` suffix in its `Display` impl
([`fmt/time/datetime.rs`](https://github.com/tokio-rs/tracing/blob/master/tracing-subscriber/src/fmt/time/datetime.rs)):

```rust
write!(f, "-{:02}-{:02}T{:02}:{:02}:{:02}.{:06}Z", ...)
```

Secondly, the timezone-aware timers aren't compiled in — they live behind the
`chrono` / `time` / `local-time` features:

```toml
# Cargo.toml:689
[workspace.dependencies.tracing-subscriber]
version = "0.3.23"
features = ["env-filter"]
```

drott:
Fixes #5931.

`EventFormatter` calls `SystemTime.format_time(..)` directly, which has no timezone concept — it renders `duration_since(UNIX_EPOCH)` as a civil datetime with a hardcoded `Z`. Because a custom `FormatEvent` impl owns the whole line, this is not reachable via `fmt::layer().with_timer(..)`, so neither `TZ` nor a bind-mounted `/etc/localtime` has any effect today.

`ChronoLocal` resolves the zone from `TZ`, then `/etc/localtime`, then `/usr/share/zoneinfo/<name>`, and falls back to UTC when all three miss ([`offset/local/unix.rs`](https://github.com/chronotope/chrono/blob/main/src/offset/local/unix.rs)):

```rust
// default to UTC if no local timezone can be found
TimeZone::local(var).ok().or_else(fallback_timezone).unwrap_or_else(TimeZone::utc)
```

The published image is distroless with neither tzdata nor `/etc/localtime`, so existing deployments keep logging UTC and this is opt-in by mounting a zone file.

Notes:

- `chrono` is already a workspace dependency (`Cargo.toml:171`), so this only enables an existing `tracing-subscriber` feature. The `Cargo.lock` change is a single line adding `chrono` to `tracing-subscriber`'s dependency list — no new packages.
- The explicit format string keeps the current 6-digit precision; `ChronoLocal::rfc_3339()` would switch to auto-precision nanoseconds. The offset suffix necessarily changes, `Z` → `+00:00`.
- `LazyLock` avoids rebuilding the format string on every event.
- Sentry and OTLP timestamps are deliberately untouched — both are wire formats that are UTC by spec and localized by the backend.

### Testing

`cargo clippy -p mas-context --all-targets` and `cargo +nightly fmt -p mas-context -- --check` are clean.

Comparing the old and new timer side by side:

```
TZ=<unset>          before=2026-08-21T20:12:52.889229Z  after=2026-08-21T23:12:52.889470+03:00
TZ=UTC              before=2026-08-21T20:13:08.022080Z  after=2026-08-21T20:13:08.022403+00:00
TZ=Europe/Berlin    before=2026-08-21T20:13:08.206784Z  after=2026-08-21T22:13:08.206977+02:00
TZ=America/New_York before=2026-08-21T20:13:08.395140Z  after=2026-08-21T16:13:08.395427-04:00
```

The UTC-fallback behaviour on a zone-file-less image is from reading chrono's source, linked above; I have not built the distroless image to confirm it end to end.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6804,6 +6804,7 @@ version = "0.3.23"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "cb7f578e5945fb242538965c2d0b04418d38ec25c79d160cd279bf0731c8d319"
 dependencies = [
+ "chrono",
  "matchers",
  "nu-ansi-term",
  "once_cell",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -688,7 +688,7 @@ features = ["cors", "fs", "add-extension", "set-header"]
 version = "0.1.41"
 [workspace.dependencies.tracing-subscriber]
 version = "0.3.23"
-features = ["env-filter"]
+features = ["env-filter", "chrono"]
 [workspace.dependencies.tracing-appender]
 version = "0.2.5"
 
--- a/crates/context/src/fmt.rs
+++ b/crates/context/src/fmt.rs
@@ -4,6 +4,8 @@
 // SPDX-License-Identifier: AGPL-3.0-only OR LicenseRef-Element-Commercial
 // Please see LICENSE files in the repository root for full details.
 
+use std::sync::LazyLock;
+
 use console::{Color, Style};
 use opentelemetry::{TraceId, trace::TraceContextExt as _};
 use tracing::{Level, Subscriber};
@@ -12,12 +14,15 @@ use tracing_subscriber::{
     fmt::{
         FormatEvent, FormatFields,
         format::{DefaultFields, Writer},
-        time::{FormatTime, SystemTime},
+        time::{ChronoLocal, FormatTime},
     },
     registry::LookupSpan,
 };
 
 use crate::LogContext;
+
+static TIMER: LazyLock<ChronoLocal> =
+    LazyLock::new(|| ChronoLocal::new("%Y-%m-%dT%H:%M:%S%.6f%:z".to_owned()));
 
 /// An event formatter usable by the [`tracing_subscriber`] crate, which
 /// includes the log context and the OTEL trace ID.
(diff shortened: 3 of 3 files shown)


reivilibre (approved):
Thanks!

PR #5935, opened 2026-08-25
Update `taiki-e/install-action` so it knows about newer `cargo-zigbuild` with fix for `--fix-cortex-a53-843419`

MadLittleMods:
Update `taiki-e/install-action` so it knows about newer `cargo-zigbuild` with fix for `--fix-cortex-a53-843419`

Spawning from this [CI failure](https://github.com/element-hq/matrix-authentication-service/actions/runs/32890477663/job/97946328369) about `error: unsupported linker arg: --fix-cortex-a53-843419`:

```
Install Rust toolchain
[...]
info: latest update on 2026-08-20 for version 1.98.0 (88d9e12ae 2026-08-18)

Install cargo-zibuild
[...]
  + cargo-zigbuild --version
  cargo-zigbuild 0.22.3

Build the binary
[...]
error: linking with `/home/runner/.cache/cargo-zigbuild/0.22.3/wrappers/d7b6/zigcc-aarch64-unknown-linux-gnu.2.17-aa8f.sh` failed: exit status: 1
  |
[...]

PR #5938, opened 2026-08-26
Pin the Rust toolchain for CI build jobs and let dependabot update it

sandhose:
This is spawning off #5935, as an update of the stable Rust toolchain broke our release build pipeline. 

We already pinned Rust for the Clippy job in CI, as Clippy tends to introduce new rules in the pedantic ruleset regularly, which made that job regularly break.

This introduces a `rust-toolchain.toml` to pin the Rust toolchain for everyone: your local `rustup` should pick it up automatically, and both the Dockerfile and the CI workflows have been reworked to use it as well. This also has one other benefit: dependabot can automatically update the toolchain in rust-toolchain.toml, so we'll have automatic toolchain upgrade PRs, with the proper CI failures (including clippy) contained in there

cc @MadLittleMods

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,3 +1,4 @@
+# Copyright 2025, 2026 Element Creations Ltd.
 # Copyright 2025 New Vector Ltd.
 #
 # SPDX-License-Identifier: AGPL-3.0-only OR LicenseRef-Element-Commercial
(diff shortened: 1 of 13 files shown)


reivilibre (approved):
I'm not really a huge fan of using rustup in the docker build, but it's a weak preference and I guess that's a matter of opinion.

PR #5956, opened 2026-09-02
Stop wrapping comments in rustfmt, drop redundant options

sandhose:
The rustfmt subtree sync that reached nightly on 2026-08-30 (rust-lang/rust#161964) includes rust-lang/rustfmt#6802: `wrap_comments` now wraps comments longer than `comment_width` that older rustfmt left alone, so the rustfmt CI job fails on every PR.

`comment_width` counts the whole line, indentation included, which wraps more than we want. Turn `wrap_comments` off instead of reformatting ~135 files. With it gone, `max_width = 100`, `comment_width = 80` and `use_small_heuristics = "Default"` are rustfmt's defaults, so drop them too. Only the two nightly-only import options remain, and the tree formats clean under the current nightly.

### DIFF
--- a/.rustfmt.toml
+++ b/.rustfmt.toml
@@ -1,11 +1,8 @@
+# Copyright 2025, 2026 Element Creations Ltd.
 # Copyright 2025 New Vector Ltd.
 #
 # SPDX-License-Identifier: AGPL-3.0-only OR LicenseRef-Element-Commercial
 # Please see LICENSE files in the repository root for full details.
 
-max_width = 100
-comment_width = 80
-wrap_comments = true
 imports_granularity = "Crate"
-use_small_heuristics = "Default"
 group_imports = "StdExternalCrate"


.rustfmt.toml
```diff
@@ -1,11 +1,8 @@
+# Copyright 2025, 2026 Element Creations Ltd.
 # Copyright 2025 New Vector Ltd.
 #
 # SPDX-License-Identifier: AGPL-3.0-only OR LicenseRef-Element-Commercial
 # Please see LICENSE files in the repository root for full details.
 
-max_width = 100
```
sandhose, line 6: Already the default value: https://rust-lang.github.io/rustfmt/?version=v1.10.0&search=#max_width

.rustfmt.toml
```diff
@@ -1,11 +1,8 @@
...
-comment_width = 80
```
sandhose, line 7: Already the default value but doesn't do anything unless `wrap_comments = true`: https://rust-lang.github.io/rustfmt/?version=v1.10.0&search=#comment_width

.rustfmt.toml
```diff
@@ -1,11 +1,8 @@
...
-wrap_comments = true
```
sandhose, line 8: That's the main change: https://rust-lang.github.io/rustfmt/?version=v1.10.0&search=#wrap_comments

.rustfmt.toml
```diff
@@ -1,11 +1,8 @@
...
 imports_granularity = "Crate"
-use_small_heuristics = "Default"
```
sandhose, line 10: Already the default value: https://rust-lang.github.io/rustfmt/?version=v1.10.0&search=#use_small_heuristics

.rustfmt.toml
MadLittleMods, line 8: While I am fine personally wrapping comments in my editor (I just have a shortcut), I don't really want to nit other people about their comment wrapping.

Does `rustfmt` still wrap comments with our new config?

MadLittleMods (approved):

.rustfmt.toml
sandhose: I don't think it will… but also the behavior of `wrap_comments` is not really good IMO. I would have hoped it would wrap comments to 80 cols *ignoring* indent, and still apply a max width of 100 for everything else, but that not the case. The result was a massive reformatting where lots of comments where wrapping at like ~60 characters, which was really too short

Issue #3206, opened 2024-09-15
End2end encrypted mautrix-signal bridge fails with "Matrix connector: homeserver does not support appservice"


xundeenergie:
**Describe the bug**

I setup a new mautrix-signal bridge with end2end-encryption. I configured double-puppeting. All according to the docs from mautrix-bridge.

I changed from matrix-synapse with directly configured oidc and ldap authentication to matrix-authentication-service successfully. After that, my mautrix-signal bridge stopped working... so i setup a new bridge to test it for this bugreport.

I asked in the matrix-room for the mautrix-signal bridge and got pointed to this issue:
https://github.com/spantaleev/matrix-docker-ansible-deploy/issues/3493

The conclusion there is, that my problem is probably with matrix-authentication-service.

What i did:

Created an example-config from the mautrix-signal directly and filled in the neccessary information from an old bridge, which worked before migration to MAS.

I activated end2end-encryption and double puppetting. 

When i start the bridge, i get this in the bridge-log and combined the nginx-log for synapse/matrix with tail to those 2 logfiles:

```
==> /var/log/nginx/matrix.schuerz.at.access.log <==                                                                                                                                                                
127.0.0.0 - - [16/Sep/2024:00:25:13 +0200] "GET /_matrix/client/versions?user_id=%40signalbotprimary%3Aschuerz.at HTTP/2.0" 200 343 "-" "mautrix-signal/0.7.0 mautrix-go/v0.20.0 go/1.22.6"                        
                                                                                                                                                                                                                   
==> /var/log/mautrix-signal/primary/mautrix-signal.log <==                                                                                                                                                         
{"level":"debug","req_id":1,"method":"GET","url":"https://matrix.schuerz.at/_matrix/client/versions?user_id=%40signalbotprimary%3Aschuerz.at","duration":57.545853,"status_code":200,"response_length":1038,"respon
se_mime":"application/json","time":"2024-09-16T00:25:13.949786427+02:00","message":"Request completed"}                                                                                                            
                                                                                                                                                                                                                   
==> /var/log/nginx/matrix.schuerz.at.access.log <==                                                                                                                                                                
127.0.0.0 - - [16/Sep/2024:00:25:13 +0200] "GET /_matrix/client/v3/account/whoami?user_id=%40signalbotprimary%3Aschuerz.at HTTP/2.0" 200 79 "-" "mautrix-signal/0.7.0 mautrix-go/v0.20.0 go/1.22.6"                
[...]

PR #5961, opened 2026-09-03
Return `M_APPSERVICE_LOGIN_UNSUPPORTED` for `m.login.application_service`

barodeur:
TLDR: return `M_APPSERVICE_LOGIN_UNSUPPORTED` error code instead of the generic `M_UNKNOWN` when `/login` is called with the `m.login.application_service` login type.

> Application services MUST NOT use the `/login` endpoint if the server doesn't support the Legacy authentication API. If `/login` is called with the `m.login.application_service` login type the server MUST return a 400 HTTP status code with an `M_APPSERVICE_LOGIN_UNSUPPORTED` error code.
>
> — [Matrix v1.19, Application Service API](https://spec.matrix.org/v1.19/application-service-api/#registration)

Currently this login type falls through the generic unsupported-credentials path, which an appservice written against Matrix 1.17 can't distinguish from an unrelated failure.

Before:

```
POST /_matrix/client/v3/login {"type": "m.login.application_service", ...}
  400 {"errcode": "M_UNKNOWN", "error": "Invalid login type"}
```

After:

```
POST /_matrix/client/v3/login {"type": "m.login.application_service", ...}
  400 {"errcode": "M_APPSERVICE_LOGIN_UNSUPPORTED", "error": "Application services can't log in through the legacy authentication API on this server"}
```

Sister PR in Synapse, which returned this error code with an unstable prefix on `/login` and `/register`: element-hq/synapse#20180.

Both are needed. When MAS is deployed, the legacy `/_matrix/client/*/login`, `/logout` and `/refresh` endpoints are proxied to MAS instead of being handled by the homeserver (see [Set up the compatibility layer](https://github.com/element-hq/matrix-authentication-service/blob/main/docs/setup/homeserver.md#set-up-the-compatibility-layer)).

Related: #3206, https://github.com/matrix-org/matrix-authentication-service/issues/2580

### DIFF
--- a/crates/handlers/src/compat/login.rs
+++ b/crates/handlers/src/compat/login.rs
@@ -1710,6 +1710,10 @@ mod tests {
 
         let request = Request::post("/_matrix/client/v3/login").json(serde_json::json!({
             "type": "m.login.application_service",
+            "identifier": {
+                "type": "m.id.user",
+                "user": "_irc_example",
+            },
         }));
 
         let response = state.request(request).await;
--- a/crates/handlers/src/compat/login.rs
+++ b/crates/handlers/src/compat/login.rs
@@ -1686,6 +1686,27 @@ mod tests {
         "###);
     }
 
+    /// Test the response of an `m.login.application_service` login.
+    #[sqlx::test(migrator = "mas_storage_pg::MIGRATOR")]
+    async fn test_appservice_login_unsupported(pool: PgPool) {
+        setup();
+        let state = TestState::from_pool(pool).await.unwrap();
+
+        let request = Request::post("/_matrix/client/v3/login").json(serde_json::json!({
+            "type": "m.login.application_service",
+        }));
+
+        let response = state.request(request).await;
+        response.assert_status(StatusCode::BAD_REQUEST);
+        let body: serde_json::Value = response.json();
+        insta::assert_json_snapshot!(body, @r###"
+        {
+          "errcode": "M_APPSERVICE_LOGIN_UNSUPPORTED",
+          "error": "Application services can't log in through the legacy authentication API on this server"
+        }
+        "###);
+    }
+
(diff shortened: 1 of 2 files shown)


crates/handlers/src/compat/login.rs
```diff
@@ -1686,6 +1702,27 @@ mod tests {
...
 
+    /// Test the response of an `m.login.application_service` login.
+    #[sqlx::test(migrator = "mas_storage_pg::MIGRATOR")]
+    async fn test_appservice_login_unsupported(pool: PgPool) {
+        setup();
+        let state = TestState::from_pool(pool).await.unwrap();
+
+        let request = Request::post("/_matrix/client/v3/login").json(serde_json::json!({
+            "type": "m.login.application_service",
+        }));
```
sandhose, line 1717: I would make this request a believable appservice login

```suggestion
        let request = Request::post("/_matrix/client/v3/login").json(serde_json::json!({
            "type": "m.login.application_service",
            "identifier": {
                "type": "m.id.user",
                "user": "_irc_example",
            },
        }));
```

docs/as-login.md
```diff
@@ -1,7 +1,7 @@
 # About Application Services login
 
 Encrypted Application Services/Bridges currently leverage the `m.login.application_service` login type to create devices for users.
-This API is *not* available in the Matrix Authentication Service.
+This API is *not* available in the Matrix Authentication Service: as per [Matrix 1.19](https://spec.matrix.org/v1.19/application-service-api/#registration), calling `/login` with this login type returns a 400 HTTP status code with an `M_APPSERVICE_LOGIN_UNSUPPORTED` error code.
```
sandhose, line 4: Technically 1.17 but fine to quote the latest spec version

sandhose (approved):

PR #5964, opened 2026-09-05
Fix positional arguments in manage CLI docs

noahziheng:
## Summary

- Document `device_id` as a positional argument for `issue-compatibility-token`.
- Document `username` as a positional argument for `register-user`.
- Add an example for non-interactive user registration.

## Testing

- Ran `git diff --check`.
- Compared the examples with the current CLI argument definitions.


sandhose (approved):
Thanks!

PR #5969, opened 2026-09-10
Remember which username was chosen when doing a registration through a social-login provider

sandhose:
This refactors the registration screen to use a `<form method=POST>` when choosing an upstream OAuth 2.0 provider. This is in preparation of moving the whole registration form to an interactive React component (see #5591), but also gives us an interesting feature: we now carry around the username that the user chose on the initial registration.


crates/handlers/src/upstream_oauth2/link.rs
```diff
@@ -454,34 +458,47 @@ pub(crate) async fn get(
                 )?
             };
 
+            let forced_or_required = provider.claims_imports.localpart.is_forced_or_required();
+
+            // What the user typed on the registration page beats the claim rendered
+            // from the upstream response, even when that claim is ignored. A provider
+            // which forces or requires the localpart never lets the user pick one.
```
reivilibre, line 465: This comment feels about 2× as complex as the logic it's describing. Better to put a 1-liner on the `None` branch?

crates/handlers/src/views/register/mod.rs
```diff
@@ -93,3 +129,395 @@ pub(crate) async fn get(
...
+    mut repo: BoxRepository,
+    Query(query): Query<OptionalPostAuthAction>,
+    cookie_jar: CookieJar,
+    Form(form): Form<ProtectedForm<RegisterForm>>,
+) -> Result<Response, RouteError> {
+    let post_auth_action = query.post_auth_action;
+
+    let Ok(form) = cookie_jar.verify_form(&clock, form) else {
+        // The CSRF token expires after an hour, which a registration page left
+        // open in a tab will outlive.
```
reivilibre, line 165: Low-quality comment, but not in a significant way at least

crates/handlers/src/views/register/mod.rs
```diff
@@ -93,3 +129,395 @@ pub(crate) async fn get(
...
+        tracing::debug!("Invalid CSRF token on the registration form, redirecting to a fresh one");
+        let destination = Register::from(post_auth_action);
+        return Ok((cookie_jar, url_builder.redirect(&destination)).into_response());
+    };
+
+    let username = form.username.trim();
+    let username = if username.len() <= MAX_USERNAME_LENGTH {
+        username
+    } else {
+        ""
```
reivilibre, line 175: Though a small comment here may not be terrible.
Is this to avoid storing unbounded stuff?
Can a user hit this in practice or will there be a `maxlength` on the `<input>`?

reivilibre (approved):
LGTM, cosmetics only

PR #5972, opened 2026-09-11
Rate limit device code link attempts

hughns:
Add a dedicated `device_code_link` rate limiter for verification of the user code from the device code grant. This applies *per source IP* rather than globally.

This is to help mitigate the User Code Brute Forcing attack as described by [RFC 8628 section 5.1](https://datatracker.ietf.org/doc/html/rfc8628#section-5.1).

Proposed default is 1 per minute and burst of 10. This can be overridden by the server admin.

New configuration option:

https://github.com/element-hq/matrix-authentication-service/blob/357f83c96a85f771ea70fcd301f568d74f2fda7d/docs/reference/configuration.md?plain=1#L644-L650


reivilibre (approved):
Thanks!

PR #5973, opened 2026-09-14
Rate limit and slow down device code token requests

hughns:
Enforce the polling interval we advertise in the device authorization response: a client which polls a still-pending grant faster than that is told to `slow_down` rather than `authorization_pending`, as described in RFC 8628 section 3.5. The check is keyed per grant and only applies whilst the grant is pending, so an impatient client is never stopped from picking up a grant the user has already approved.

Also bound how many unknown device codes a single source address may present to the token endpoint, to limit enumeration of device codes (RFC 8628 section 5.2). This only counts codes which don't match a grant, so clients polling with a valid device code are unaffected by it.

The advertised interval and the one the token endpoint enforces now both come from a new constant `DEFAULT_DEVICE_AUTHORIZATION_INTERVAL`, so they cannot drift.

New configuration option:

https://github.com/element-hq/matrix-authentication-service/blob/41344927b5324d7c9638c869b401b260ae69ef4a/docs/reference/configuration.md?plain=1#L644-L654


crates/config/src/sections/rate_limiting.rs
```diff
@@ -31,6 +31,18 @@ pub struct RateLimitingConfig {
...
+
+    /// Controls how many unknown device codes may be presented to the token
+    /// endpoint, based on source IP address.
+    /// This bounds enumeration of device codes of the
+    /// Device Authorization Grant.
+    ///
+    /// Note: this limit only applies to device codes which don't match a
+    /// grant, so clients polling with a valid device code are unaffected.
+    /// Their polling rate is governed by the interval advertised in the
+    /// device authorization response instead.
```
sandhose, line 43: That is quite a flaw, isn't it? This means "if the code is valid you're fine", so even if you've hit the rate limit, you can still probe for valid codes?

crates/handlers/src/rate_limit.rs
```diff
@@ -148,6 +165,16 @@ impl LimiterInner {
             email_authentication_attempt_per_session: RateLimiter::keyed(
                 config.email_authentication.attempt_per_session.to_quota()?,
             ),
+            device_code_exchange_per_requester: RateLimiter::keyed(
+                config.device_code_exchange.to_quota()?,
+            ),
+            // This one isn't configurable: it has to match the polling interval we
+            // advertise in the device authorization response, else we would be telling
+            // clients to slow down whilst they poll at exactly the rate we asked for
```
sandhose, line 173: I disagree on both: in case you're scaling MAS horizontally, you will be allowing more code polls by having multiple instances, so you may want to tweak the numbers? I would also not be as tight and allow like twice per interval, because really it's not that expensive, and we just want to get really misbehaving clients out, not ones that are slightly too fast

crates/handlers/src/oauth2/token.rs
```diff
@@ -898,8 +928,19 @@ async fn device_code_grant(
...
+        .await?;
+
+    let Some(grant) = grant else {
+        // The device code doesn't match any grant. Rate-limit how many of those a
+        // single requester may present, to bound enumeration of device codes.
+        // RFC 8628 section 5.2
+        if let Err(e) = limiter.check_device_code_exchange(requester) {
+            warn!(error = &e as &dyn std::error::Error);
+            return Err(RouteError::DeviceCodeExchangeRateLimited);
+        }
```
sandhose, line 940: Yeah so, I would always evaluate that rate limit before any database lookup, but have it configured to allow let's say 10 concurrent polls?

sandhose (changes requested):

crates/handlers/src/rate_limit.rs
```diff
@@ -325,6 +356,46 @@ impl Limiter {
...
+    /// response: clients polling faster than that should be told to slow
+    /// down, as described in RFC 8628 section 3.5.
+    ///
+    /// # Errors
+    ///
+    /// Returns an error if the grant was polled too soon.
+    pub fn check_device_code_poll(&self, grant_id: Ulid) -> Result<(), DeviceCodePollLimitedError> {
+        self.inner
+            .device_code_poll_per_grant
+            .check_key(&grant_id)
```
hughns, line 394: Note to self: key on `requester` as well as `device_code` (not `grant_id`).

crates/handlers/src/rate_limit.rs
hughns: Two numbers to be configurable:

- the advertised poll interval (in the Device Authorization Response).
- the accepted poll interval (before a device gets limited). n.b. this is per-shard

crates/config/src/sections/rate_limiting.rs
hughns: Agreed.

PR #5974, opened 2026-09-14
Improve usability of user code during device code flow

hughns:
Use Crockford Base32 alphabet with `O`/`0` and `I`/`L`/`1` folding. Also help the browser to prompt for correct input.

The current alphabet gave 29.7 bits of entropy (it was random 0-9a-zA-Z which was then mapped to 0-9A-Z) whereas the new encoding is 30 bits.


crates/data-model/src/oauth2/device_code_grant.rs
```diff
@@ -8,11 +8,70 @@ use std::net::IpAddr;
...
+///
+/// Crockford also allows hyphens as readability separators and ignores them
+/// when decoding, but we never put any into the codes we hand out, so there is
+/// nothing for a user to optionally include and nothing to strip here.
+/// Whitespace and punctuation are likewise left alone, and so will simply fail
+/// to match.
+///
+/// This mapping is lossy against the alphanumeric alphabet user codes used to
+/// be drawn from, so callers must look a code up as plainly uppercased
+/// *before* falling back to this — see the lookup in the device link handler.
```
reivilibre, line 60: This explanation of the hazard is not very clear IMO

crates/data-model/src/oauth2/device_code_grant.rs
```diff
@@ -267,3 +326,81 @@ impl DeviceCodeGrant {
...
+        assert_eq!(normalize_user_code("u"), "U");
+    }
+
+    /// We never put separators into the codes we hand out, so there is nothing
+    /// for a user to optionally include and nothing to strip. Anything which
+    /// isn't part of the code is passed through, and will simply fail to match.
+    #[test]
+    fn test_normalize_user_code_does_not_strip_separators() {
+        assert_eq!(normalize_user_code("abc-def"), "ABC-DEF");
+        assert_eq!(normalize_user_code(" abcdef "), " ABCDEF ");
```
reivilibre, line 395: I guess we didn't do it before, but it'd be quite tempting to remove whitespace, as I have seen people get tripped up by forms that were choking on whitespace the user wasn't aware of.

Could `.trim()` but I'd be fairly tempted to go one step further and remove whitespace even from the middle.

reivilibre (approved):
SGTM otherwise

crates/data-model/src/oauth2/device_code_grant.rs
hughns: I considered it, but because MAS serves the form that the user is entering the code into the input is constrained at that point. (The user should not be entering a code from another form and IIRC we prevent cross-origin submissions to avoid CSRF attacks)

crates/data-model/src/oauth2/device_code_grant.rs
hughns: On re-reading it I completely agree. Attempted to improve in ada65be990a778f362119ffadd003307ec0edd28

crates/data-model/src/oauth2/device_code_grant.rs
```diff
@@ -8,11 +8,64 @@ use std::net::IpAddr;
```
MadLittleMods, line 8: How is usability improved by this change? (I don't know)

PR #5975, opened 2026-09-15
Add a shared boot helper for React islands

sandhose:
The helper parses the mount node's dataset, waits for the i18n setup before the first render, and swaps the page to a server-rendered error notice when booting fails. This fixes the account app staying on its loading screen forever when fetching the locale data fails, and gives upcoming server-rendered pages with React islands a common entry point. Waiting on the setup requires `setupI18n` to return the initialisation promise, so it now does.


barodeur (approved):
> This fixes the account app staying on its loading screen forever when fetching the locale data fails

Turns out that `i18n.init` returned Promise doesn't reject when locales don't load 😢

https://github.com/i18next/i18next/blob/e1c60d4dd28a16f91be7f55b3685ffcf9760619b/src/i18next.js#L211

sandhose:
> Turns out that i18n.init returned Promise doesn't reject when locales don't load 😢

Oh :( I'll fix that in a follow up

PR #5978, opened 2026-09-15
Extract the password complexity feedback out of the double-password input component

sandhose:
The strength meter, the improvement hints and the too-weak error move to a `PasswordComplexityFeedback` component backed by a `usePasswordComplexity` hook, so that forms which lay out their password fields differently can reuse them.

`PasswordCreationDoubleInput` takes the minimum complexity as a plain number instead of a `SiteConfig` fragment: this is important as we plan to render this input outside of the main React app, without loading data from the GraphQL API.

This mainly moves code around, no real behavior changed here.


frontend/src/utils/usePasswordComplexity.ts
```diff
@@ -0,0 +1,46 @@
...
+        score: 0,
+        scoreText: t("frontend.password_strength.placeholder"),
+        improvementsText: [],
+      });
+    } else {
+      loadPromise
+        .then((estimatePasswordComplexity) =>
+          estimatePasswordComplexity(deferredPassword, t),
+        )
+        .then((response) => setResult(response));
```
barodeur, line 41: In the future, we should eventually refactor this to use Suspense, or something less prone to race condition than `useEffect(() => promise.then())`

barodeur (approved):
🔥

PR #5981, opened 2026-09-17
Add MSC4198 `login_hint_types_supported` discovery metadata

tonkku107:
Implementing the new server metadata field that was added to the proposal last month


sandhose (approved):
Thanks!

PR #5983, opened 2026-09-18
Handle the whole password registration on `/register`

sandhose:
Registration was split over two pages: `/register` asked for a username and let you pick between a password and an upstream provider, then `/register/password` asked for the username again along with the email, the password and the terms of service. This merges the two into a single server-rendered page on `/register`.

This is the backend and template half of the React registration form. It carries no frontend change: `/register` is still a plain server-rendered form, it just has all the fields on it now.

**This is not meant to ship alone**, because it would make the registration form one step on servers with open registrations and local password + upstream IDPs options (so basically only matrix.org). The rest of the PRs can somewhat land independently, but this one should land *at least* with the next one, #5591


crates/handlers/src/lib.rs
```diff
@@ -395,9 +400,10 @@ where
             mas_router::Register::route(),
             get(self::views::register::get).post(self::views::register::post),
         )
+        // XXX: hard-coded redirect from /register/password to /register
```
reivilibre, line 403: weird comment.

Since we have the chance, why not put an explanation of the history here?
If it has an `XXX` that sounds like tech debt to be cleaned up. Is it? Doesn't seem right to me

crates/handlers/src/views/register/mod.rs
```diff
@@ -488,14 +595,22 @@ mod tests {
...
+            "password": "correcthorsebatterystaple",
+            "password_confirm": "correcthorsebatterystaple",
+            "accept_terms": "on",
         })));
         let response = state.request(request).await;
-        response.assert_status(StatusCode::SEE_OTHER);
-        response.assert_header_value(LOCATION, "/register/password");
+        response.assert_status(StatusCode::OK);
+        assert!(
+            response.body().contains("Username is too long"),
```
reivilibre, line 610: 200 OK with an error message seems suspicious

crates/handlers/src/views/register/mod.rs
```diff
@@ -488,14 +595,22 @@ mod tests {
             .expect("the upstream sessions cookie should be set");
         assert!(sessions[0].get("username").is_none());
 
-        // And it isn't put in the redirect to the password registration page either
+        // The password path sees it whole, and the policy rejects it
```
reivilibre, line 598: 'The password path'? 'sees it whole'? I have no idea what it's trying to say :D

crates/handlers/src/lib.rs
reivilibre, line 1: > This is not meant to ship alone, because it would make the registration form one step on servers with open registrations and local password + upstream IDPs options (so basically only matrix.org)

Not sure I'm following what this means.

reivilibre (approved):
I think this is fine? Though I don't follow the reason why this can't land

Issue #5991, opened 2026-09-21
Template rendering error when signing up with email


jake-penguins:
**Describe the bug**
When trying to create a new account and choosing to register with an email address, an error screen is displayed saying that a template failed to register, making new signups impossible.

The error message displayed:

> ### Unexpected error
> Could not render template "pages/register/password.html"
> `invalid operation: value of type undefined is not callable (in components/captcha.html:14)`

**To Reproduce**
1. Go to https://account.matrix.org/register
2. Enter any username (new or already-existing)
3. Click "Continue with email address"
4. Error screen

**Expected behavior**
The sign up form continues and allows the user to input their email and password.

**Screenshots**
<img width="1920" height="941" alt="Image" src="https://github.com/user-attachments/assets/789b6354-d92e-4db8-84e4-c467fda3e4b8" />

**Desktop (please complete the following information):**
 - OS: Linux
 - Browser: Brave
 - Version: 1.95.101

sandhose:
Thanks for the report! A fix is on the way in #5992, and we'll publish a new version 1.25.1 with it; we rolled back matrix.org to 1.24.0 in the meantime

MadLittleMods:
[`v1.25.1`](https://github.com/element-hq/matrix-authentication-service/releases/tag/v1.25.1) released with https://github.com/element-hq/matrix-authentication-service/pull/5992 🚀

PR #5992, opened 2026-09-21
Fix the rendering of the password registration page when a CAPTCHA provider is configured

sandhose:
Fixes #5991 

Regressed in https://github.com/element-hq/matrix-authentication-service/pull/5885

So in minijinja, when you import macros from a template, the macros have access to variables declared in the template importing them.

This means that `components/captcha.html` didn't have access to the `_` variable declared in `base.html` when imported from `pages/register/password.html` (even though that one "extended" from `base.html`).

This has two changes: one which makes sure that templates which require the "captcha" context render with and without a captcha provider set, the second which fixes the rendering


templates/base.html
```diff
@@ -16,6 +16,7 @@
 {% import "components/errors.html" as errors %}
 {% import "components/icon.html" as icon %}
 {% import "components/scope.html" as scope %}
+{% import "components/captcha.html" as captcha %}
```
MadLittleMods, line 19: Any sort of test we can add to exercise rendering of `/register/password` to catch this kind of thing in the future?

MadLittleMods: *squints* I guess that's what the `crates/templates/src/context/captcha.rs` changes are doing. Adding rendering tests with an without captcha

sandhose: One test we have is trying to render every template with sample template contexts. [a6de208](https://github.com/element-hq/matrix-authentication-service/pull/5992/commits/a6de208d83bfaaf06104fa62f3503e066b05e0ce) makes sure that the `WithCaptcha` context wrapper renders each template that uses.

If you want to be convinced by that, we also have a `mas-cli` command which runs this test and can output the result to a directory


```
# cargo run -- templates check --out-dir /tmp/rendered
…lots of templates rendered
# grep -C2 "noscript" /tmp/rendered/pages_register_password-index=0-captcha=some-locale=en.html


    <noscript>
      <div class="captcha-noscript">
        This form is protected by a CAPTCHA and requires JavaScript to be enabled to submit it. Please enable JavaScript in your browser and reload this page.
      </div>
    </noscript>

    <script type="module" src="/assets/captcha-BYpTkJ1m.js" crossorigin="anonymous" integrity="sha384-UX6GpmKNm/k7YyIZ9wzIdFaprndhumC1rAuDYwCxsI9QlhQWc4AcC865KK8eGVEO"></script>
```

and before the fix, it would fail to run:

```
# mas-cli templates check
2026-09-21T22:30:50.984600+02:00  INFO mas_templates::check:358 Rendering template name="pages/register/password.html" context={"lang":"cs","csrf_token":"fake_csrf_token","captcha_config":{"service":"recaptcha_v2","site_key":"site-key"},"form":{"fields":{},"errors":[]},"next":null} trace.id=b702e8661ff886f24c9626b5f82d4f7a
Error: Failed to render sample template "pages/register/password.html"-SampleIdentifier { components: [("index", "0"), ("captcha", "some"), ("locale", "cs")] } with context {"lang":"cs","csrf_token":"fake_csrf_token","captcha_config":{"service":"recaptcha_v2","site_key":"site-key"},"form":{"fields":{},"errors":[]},"next":null}

Caused by:
    0: could not render template "pages/register/password.html"
    1: invalid operation: value of type undefined is not callable (in components/captcha.html:14)
```

MadLittleMods: :+1: Perfect, thank you! `cargo run -- templates check` fails with a6de208 but works with the latest fixes here :white_check_mark:

MadLittleMods (approved):

sandhose:
The main test suite has passed, `cargo deny` is still failing and I don't want to wait on the ESS test job, so bypassing the protection rules and merging

